PR #51, opened 2024-09-16
Fix Wordle Example

pkiv:
# why
It doesn't work currently.

# what changed
I switched the ask endpoint to use o1, which is super slow. I also modified the prompts. I wonder if the prompt changes are good enough. 

# test plan

### DIFF
--- a/lib/inference.ts
+++ b/lib/inference.ts
@@ -128,22 +128,8 @@ export async function ask({
   question: string;
   client: OpenAI;
 }) {
-  console.log("asking question", question);
   const response = await client.chat.completions.create({
-    model: "o1-preview",
-    messages: [
-      {
-        role: "user", 
-        content: question
-      }
-    ]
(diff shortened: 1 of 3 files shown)


lib/inference.ts
```diff
@@ -128,8 +128,22 @@ export async function ask({
   question: string;
   client: OpenAI;
 }) {
+  console.log("asking question", question);
   const response = await client.chat.completions.create({
-    model: "gpt-4o",
+    model: "o1-preview",
```
filip-michalsky: I will push these up into env variable or an LLM router. Will add as an open issue. we shouldn't be hardcoding model names

filip-michalsky:
the wordle example works now!

pkiv:
@filip-michalsky I reset it to work with 4o so we don't have to fully switch to o1 yet.

PR #56, opened 2024-09-17
Decouple OpenAI from Stagehand

filip-michalsky:
# why
As a developer, I want to be able to use different LLM models with Stagehand. I also want to be able to use different LLMs  based on my task (extract, ask, observe and act)
As the library maintainer, I don't want to have a hard-coded dependency on one model provider.

# what changed
In examples/index.ts:

Added an optional model_name parameter to stagehand.act() calls.
Added model_name parameter to stagehand.ask() call.
Added a console.log for the correct variable.

Added a new file lib/LLMProvider.ts:

Introduced a LLMProvider class to manage different language models.
Implemented methods to get extraction and chat clients based on the model name.
Added support for OpenAI models: 'gpt-4o', 'gpt-4o-mini', 'o1-preview', 'o1-mini'.
Included a custom ValueError class for error handling.

In lib/index.ts:

Imported the new LLMProvider class.
Removed direct OpenAI and Instructor instantiation, replacing it with LLMProvider.
Added llmProvider as an optional parameter in the Stagehand constructor.
Modified extract, observe, ask, and act methods to use LLMProvider and accept a model_name parameter.

In lib/inference.ts:

Updated all functions (act, extract, observe, ask) to use LLMProvider instead of direct OpenAI or Instructor clients.
Added model_name parameter to all functions and used it when creating chat completions.
Modified function signatures to accept LLMProvider instead of OpenAI or Instructor clients.

# test plan
- evals pass (4/5, one has known bug)
- the example works
- more tests for the library in general are needed

### DIFF
--- a/examples/index.ts
+++ b/examples/index.ts
@@ -8,9 +8,9 @@ async function example() {
     verbose: true,
     debugDom: true,
   });
-  await stagehand.init();
+  await stagehand.init({ modelName: "gpt-4o" }); // optionally specify model_name, defaults to "gpt-4o"
   await stagehand.page.goto("https://www.nytimes.com/games/wordle/index.html");
-  await stagehand.act({ action: "start the game", model_name: "gpt-4o" }); // optionally specify model_name, defaults to "gpt-4o"
+  await stagehand.act({ action: "start the game", modelName: "gpt-4o" }); 
   await stagehand.act({ action: "close tutorial popup"});
 
   let guesses: { guess: string | null; description: string | null }[] = [];
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -70,6 +70,7 @@ export class Stagehand {
   public env: "LOCAL" | "BROWSERBASE";
   public verbose: boolean;
   public debugDom: boolean;
+  private defaultModelName: string;
 
   constructor(
     {
@@ -92,6 +93,7 @@ export class Stagehand {
     this.actions = {};
     this.verbose = verbose;
     this.debugDom = debugDom;
+    this.defaultModelName = "gpt-4o";
   }
 
   log({ category, message }: { category?: string; message: string }) {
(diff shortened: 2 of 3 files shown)


filip-michalsky:
Not addressed yet: 
- add other LLM providers such as Anthropic.
- rip out Instructor as a dependency since we can use https://platform.openai.com/docs/guides/structured-outputs/examples or similar.

I suggest these would be a separate PR.

lib/index.ts
```diff
@@ -265,10 +267,11 @@ export class Stagehand {
 
     return observationId;
   }
-  async ask(question: string): Promise<string | null> {
+  async ask(question: string, model_name: string = "gpt-4o"): Promise<string | null> {
```
pkiv: I think the model_name should be optional, with the stagehand.init() command setting the default, and they can override on individual calls if needed. This should be the same for all the other calls.

examples/index.ts
```diff
@@ -10,8 +10,8 @@ async function example() {
   });
   await stagehand.init();
   await stagehand.page.goto("https://www.nytimes.com/games/wordle/index.html");
-  await stagehand.act({ action: "start the game" });
-  await stagehand.act({ action: "close tutorial popup" });
+  await stagehand.act({ action: "start the game", model_name: "gpt-4o" }); // optionally specify model_name, defaults to "gpt-4o"
```
pkiv: setting the model on the action level is a bit repetitive. what if we could set the models for act, observe, and extract in the stagehand init function, and allow them to be overriden on a per-function basis?

lib/index.ts
filip-michalsky: Update. The init now has a default model name which can also be optionally set by the user. the default "defaults" to gpt-4o.

examples/index.ts
filip-michalsky: code DRYed.

filip-michalsky:
Should be good to go!

pkiv:
Approved

PR #57, opened 2024-09-18
Traverse child nodes if they're interactive

pkiv:
# why
Peeler complex was failing because it was selecting a div that wasn't actually inputable. 

# what changed
Always Traverse Children: By removing the continue statements, we ensure that all child nodes are examined, allowing nested interactive elements like <input> fields to be detected.

Collect All Relevant Elements: Even if a parent element is interactive, its child elements might also be interactive or relevant leaf nodes, so we add the parent to candidateElements without skipping its children.

This change should allow your code to correctly identify and return the <input> field instead of the wrapping <div>. Please try this modification and see if it resolves the issue.

# test plan
Evals - peeler complex works now!

### DIFF
--- a/lib/dom/process.ts
+++ b/lib/dom/process.ts
@@ -31,21 +31,21 @@ async function processElements(chunk: number) {
     if (element && isElementNode(element)) {
       const childrenCount = element.childNodes.length;
 
-      // if you have no children you are a leaf node
-      if (isLeafElement(element)) {
-        if ((await isActive(element)) && isVisible(element)) {
-          candidateElements.push(element);
-        }
-        continue;
-      } else if (isInteractiveElement(element)) {
-        if ((await isActive(element)) && isVisible(element)) {
-          candidateElements.push(element);
-        }
-        continue;
-      }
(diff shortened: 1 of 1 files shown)


lib/dom/process.ts
```diff
@@ -31,21 +31,21 @@ async function processElements(chunk: number) {
...
-      // if you have no children you are a leaf node
-      if (isLeafElement(element)) {
+      // Always traverse child nodes
+      for (let i = childrenCount - 1; i >= 0; i--) {
+        const child = element.childNodes[i];
+        DOMQueue.push(child as ChildNode);
+      }
+
+      // Check if element is interactive
+      if (isInteractiveElement(element)) {
```
filip-michalsky, line 41: LGTM!

One thing to note: this might significantly affect the speed/efficiency of the DOM crawler. Trust me, I have been deep in there scraping real estate websites. Maybe we add an Issue to look at the efficiency of crawling all child nodes vs a different approach?

filip-michalsky: also evals passed for me too

PR #59, opened 2024-09-18
[WIP] Fm/add anthropic llm

filip-michalsky:
# why
we want to add other LLM models

# what changed
- refactor llm dependencies and create a layer of abstraction not touching the Stagehand code.

# test plan
- evals need to pass

This will be updated.


evals/index.eval.ts
```diff
@@ -83,6 +83,7 @@ const peeler_complex = async () => {
   const { price } = await stagehand.extract({
     instruction: "get the price of the peeler",
     schema: z.object({ price: z.number().nullable() }),
+    modelName: "gpt-4o-2024-08-06",
```
filip-michalsky, line 86: since we are using openAI structured outputs, we need to specify the model name... on 10/2/2024 gpt-4o will repoint to the model supporting the structured outputs.

PR #62, opened 2024-09-19
three levels of logs

filip-michalsky:
# why
To address more granular logging control verbose
# what changed
changed logging from boolean to int - 0,1 and 2, with 0 no logging and 2 the maximum logging level.
# test plan
example and evals passed


filip-michalsky:
This can be merged once an approval is secured.

PR #70, opened 2024-09-22
add more evals (simple google search & github) + small bug fix

navidkpr:
# why
 Need more failed eval data to hill-climb on for improving Stagehand / auto-prompting (editing users prompts using LLMs)

# what changed
Eval cases added:
1. Simple google search
2. Two new github related actions (getting collaborators and commits). _Both failing right now_
3. Twitter sign up until verification with random email generation. _Failing right now_

# test plan

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -14,7 +14,7 @@ async function getBrowser(env: "LOCAL" | "BROWSERBASE" = "LOCAL") {
   if (process.env.BROWSERBASE_API_KEY && env !== "LOCAL") {
     console.log("Connecting you to broswerbase...");
     const browser = await chromium.connectOverCDP(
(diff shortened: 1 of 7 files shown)


lib/index.ts
```diff
@@ -217,6 +217,7 @@ export class Stagehand {
         progress: progress + newProgress + ", ",
         content: merge(content, output),
         chunksSeen,
+        modelName: modelName || this.defaultModelName,
```
filip-michalsky: I fixed this in my merged PR too I think, just fyi if it gets a conflict

filip-michalsky:
Confirming the 3 evals (twitter, extract github collaborators, extract last 20 github commits) also fail for me.

PR #72, opened 2024-09-23
add gh actions

filip-michalsky:
# why
need to run evals before merging PRs

# what changed
added GH actions script

# test plan
it tests itself.


filip-michalsky:
Also, removing instructor from dependencies.

filip-michalsky:
this one should be good to go. I gave up on adding the Braintrust Eval GH to the CI script, way too much friction

PR #75, opened 2024-09-24
add bananalyzer eval lib

navidkpr:
# why
Mainly because we needed a suite of reliable evals to improve Stagehand on.

We can now add any of the 319 test cases from bananalyzer with a single line. I will be adding the useful test cases in a separate PR. 

# what changed
1. Added a server to serve mhtml files
2. Moved schemas from Bananalyzer into Typescript
3. Added an integration between bananalyzer and Stagehand
4. Added 4 Bananalyzer test cases into Stagehand eval

# test plan
This is tested by default every time we run the eval system.

# For the future
Our current eval system uses active urls. This can cause a problem because the urls we have might change over time. For example, a single "." change will cause the eval to fail on an extract job. To get around that we can take snapshots of the websites we use in each eval using playwright and export that into a mhtml file.

This PR adds the functionality to test Stagehand on those snapshots. Making the site change issue null.

Also, other eval systems for web agents also use MHTML format (for example [Mind2Web](https://osu-nlp-group.github.io/Mind2Web/)). So next time it will be much easier to integrate with an external eval source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,7 +274,7 @@ You'll also need a Braintrust key to run evals
 BRAINTRUST_API_KEY=""%
 ```
 
-and then run the init script for bananalyzer static files:
+and then run the init script to fetch the eval examples from the bananalyzer repo:
 
 `./evals/bananalyzer-ts/init.sh`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -273,6 +278,10 @@ Then, run:
 
 `pnpm evals`
 
+For running the bananalyzer eval, you'll need to run the init script first to download the static assets:
+
+`./evals/bananalyzer-ts/init.sh`
+
```
filip-michalsky, line 294: @navidkpr can you pls add instructions how to run the bananalyzer evals?

PR #78, opened 2024-09-26
[WIP - do not merge] Fm/actions not found

filip-michalsky:
# why

 Paul mentioned a vision where stagehand will be a tool we can pass to an LLM to perform an action on the web. In that sense, we DO want to return success/ failure signal and a reason for the failure so that the agent/developer can heal their pipelines (either manually or agentically).

# what changed
We are returning a success/error from each `stagehand` method. This slightly changes the API but opens us a route to use stagehand as a tool to pass to LLMs more seamlessly.

# test plan
need more evals should be our motto

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -120,30 +120,41 @@ export class Stagehand {
     await download.delete();
   }
 
-  async init({ modelName = "gpt-4o" }: { modelName?: string } = {}) {
-    const { context } = await getBrowser(this.env, this.headless);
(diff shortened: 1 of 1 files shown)


filip-michalsky:
not loving the new API - might change back to just throw an action error if action is not performed

filip-michalsky:
we won't be merge this, reworked at https://github.com/browserbase/stagehand/pull/79

PR #79, opened 2024-09-27
add action message

filip-michalsky:
# why
This is second, reworked attempt.

Paul mentioned a vision where stagehand will be a tool we can pass to an LLM to perform an action on the web. In that sense, we DO want to return success/ failure signal and a reason for the failure so that the agent/developer can heal their pipelines (either manually or agentically).

# what changed
we are returning a message from a finished `act()` method. This is currently not utilized in evals so we need to add evals to test it.
# test plan
need to add evals for this.

### DIFF
--- a/evals/playground.ts
+++ b/evals/playground.ts
@@ -16,7 +16,7 @@ const homedepot = async () => {
     await stagehand.act({ action: "search for gas grills" });
     await stagehand.waitForSettledDom();
 
-    await stagehand.act({ action: "click on the best selling gas grill" });
+    await stagehand.act({ action: "click on the first gas grill" });
     await stagehand.waitForSettledDom();
 
     await stagehand.act({ action: "click on the Product Details" });
@@ -42,7 +42,6 @@ const homedepot = async () => {
 
     return true;
 
-    return isTitleValid;
   } catch (error) {
     console.error(`Error in homedepot function: ${error.message}`);
     return false;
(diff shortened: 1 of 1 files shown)


filip-michalsky:
My thinking behind this updated API design:

By return a message from `act()` we do not break anything in existing evals. It's a simply optional return value which developer can use to retry actions and perhaps change the prompt until the action gets done. It can also help with the prompt optimizer where an another LLM can change the action prompt if previous action failed, etc.

filip-michalsky:
This needs to get fixed - I resolved the merge conflicts but now missing some banalyzer dependencies. Also `homedepot` eval is still not passing.

lib/dom/process.ts
```diff
@@ -16,7 +16,7 @@ async function processElements(chunk: number) {
   const chunkHeight = viewportHeight * chunk;
   const offsetTop = chunkHeight;
 
-  window.scrollTo(0, offsetTop);
+  window.scrollTo({ top: offsetTop, left: 0, behavior: 'smooth' });
```
filip-michalsky, line 19: I added "smooth" to the scrolling behavior, the hypothesis is that it reduces hard reloads of the site for slow sites but its just based on the one eval (homedepot)

lib/index.ts
```diff
@@ -344,39 +366,58 @@ export class Stagehand {
     steps?: string;
     chunksSeen?: Array<number>;
     modelName?: string;
-  }): Promise<void> {
+  }): Promise<{ success: boolean; message: string; action: string }> {
```
filip-michalsky, line 369: this is for agents/humans to get more info. Saw @navidkpr did this in the extract method too.

PR #89, opened 2024-10-03
support multi-chunk extracts + prompt updates

navidkpr:
# why

# what changed

# test plan

### DIFF
--- a/evals/index.eval.ts
+++ b/evals/index.eval.ts
@@ -329,6 +329,38 @@ const wikipedia = async () => {
   };
 };
 
+
+// Validate that the action is not found on the page
+const nonsense_action = async () => {
+  const stagehand = new Stagehand({ env: "LOCAL", verbose: 1, debugDom: true, headless: true });
+  await stagehand.init();
+  
+  try {
+    await stagehand.page.goto("https://www.homedepot.com/");
+    await stagehand.waitForSettledDom();
+
+    const result = await stagehand.act({ action: "click on the first banana" });
+    console.log("result", result);
+
+    // Assert the output
+    const expectedResult = {
+      success: false,
+      message: 'Action not found on the current page after checking all chunks.',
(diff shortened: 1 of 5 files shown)


evals/index.eval.ts
```diff
@@ -148,6 +148,86 @@ const peeler_complex = async () => {
...
+const homedepot = async () => {
+  const stagehand = new Stagehand({
+    env,
+    verbose: 1,
+    headless: process.env.HEADLESS !== "false",
+  });
+  await stagehand.init();
+
+  try {
+    // await stagehand.page.goto("https://www.homedepot.com/");
```
filip-michalsky: @navidkpr we also need to test that the search bar works (uncomment this) and that stagehand finds the best selling grill by itself - this makes this eval even harder but I think we should solve it

lib/index.ts
```diff
@@ -52,14 +52,64 @@ async function getBrowser(env: "LOCAL" | "BROWSERBASE" = "LOCAL", headless: bool
...
+          "--enable-accelerated-2d-canvas",
+        ],
+        excludeSwitches: "enable-automation",
+        userDataDir: "./user_data",
+      },
     );
 
     console.log("Local browser started successfully.");
+
+    await applyStealthScripts(context);
```
filip-michalsky, line 70: this might help with the Costar eval which does not work in headless mode!

lib/prompt.ts
```diff
@@ -9,7 +9,7 @@ You are given:
 2. the steps that have been taken so far
 3. a list of active DOM elements in this chunk to consider to accomplish the goal. 
 
-You have 2 tools that you can call: doAction, and skipSection
+You have 2 tools that you can call: doAction, and skipSection. Do action only performs Playwright actions. Do not perform any other actions.
```
filip-michalsky, line 12: Hope this helps with me previously writing edge cases

PR #90, opened 2024-10-03
fix eval in ci

navidkpr:
# why

# what changed

# test plan

### DIFF
--- a/evals/index.eval.ts
+++ b/evals/index.eval.ts
@@ -450,5 +450,5 @@ Eval("stagehand", {
     }
   },
   scores: [exactMatch],
-  trialCount: 3,
+  // trialCount: 3,
 });


evals/index.eval.ts
```diff
@@ -123,120 +123,76 @@ const peeler_complex = async () => {
...
     await stagehand.context.close();
     return false;
   }
 };
 
-const twitter_signup = async () => {
-  const stagehand = new Stagehand({ env: "LOCAL", verbose: 1 });
+const extract_last_twenty_github_commits = async () => {
+  const stagehand = new Stagehand({
+    env: "LOCAL",
```
pkiv, line 168: Should we have this use BrowserBase instead of local? Will be faster and more repeatable that way

pkiv:
Lgtm

PR #100, opened 2024-10-07
[WIP] Fm/support iframes 3

filip-michalsky:
# why
We need to support i-frames

# what changed

# test plan

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -509,8 +509,12 @@ export class Stagehand {
   
     // Get the chunksSeen for the current frame
     const chunksSeen = chunksSeenPerFrame[frameId];
+<<<<<<< HEAD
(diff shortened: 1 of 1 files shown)


filip-michalsky:
@navidkpr can I get some feedback here pls, if I am on the right track, I will:

a) add multi-frame to extract() and observe()
b) add the eval to extract the filled out form

c) do we want to add wordle as its own 'playground' script? right now `pnpm example` runs the skin care inner i-frame fill in.

main philosophy here is to iterate through i-frames one by one if action/extraction cannot be completed, starting with the first one.

one big missing feature is extraction across multiple chunks vs completing after detecting something relevant in the first chunk.

filip-michalsky:
re-opened a new PR https://github.com/browserbase/stagehand/pull/100

PR #105, opened 2024-10-11
add vision to action

navidkpr:
# why
Vision as a fallback helps do complex tasks

# what changed

# test plan

### DIFF
--- a/lib/vision.ts
+++ b/lib/vision.ts
@@ -4,34 +4,52 @@ import path from "path";
 import sharp from "sharp";
 import { exec } from "child_process";
 
-interface AnnotationBox {
+type AnnotationBox = {
   x: number;
   y: number;
   width: number;
   height: number;
   id: string;
-}
+};
 
-interface NumberPosition {
+type NumberPosition = {
   x: number;
   y: number;
-}
+};
 
-export class ScreenshotAnnotator {
+export class ScreenshotService {
   private page: Page;
   private selectorMap: Record<number, string>;
   private annotationBoxes: AnnotationBox[] = [];
   private numberPositions: NumberPosition[] = [];
   private isDebugEnabled: boolean;
+  private verbose: 0 | 1 | 2;
 
   constructor(
     page: Page,
     selectorMap: Record<number, string>,
+    verbose: 0 | 1 | 2,
     isDebugEnabled: boolean = false,
   ) {
     this.page = page;
     this.selectorMap = selectorMap;
     this.isDebugEnabled = isDebugEnabled;
+    this.verbose = verbose;
+  }
+
+  log({
+    category,
+    message,
+    level = 1,
+  }: {
+    category?: string;
+    message: string;
+    level?: 0 | 1 | 2;
+  }) {
+    if (this.verbose >= level) {
+      const categoryString = category ? `:${category}` : "";
+      console.log(`[stagehand${categoryString}] ${message}`);
+    }
   }
 
   async getScreenshot(): Promise<Buffer> {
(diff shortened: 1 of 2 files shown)


lib/dom/process.ts
```diff
@@ -40,14 +69,22 @@ async function processElements(chunk: number) {
       // Check if element is interactive
       if (isInteractiveElement(element)) {
         if ((await isActive(element)) && isVisible(element)) {
-          candidateElements.push(element);
+          shouldAddElement = true;
```
kamath: can you rebase so changes are consistent to just this PR? looks like these changes are from #104. this helps us retrace our changes and isolate diffs better

lib/vision.ts
```diff
@@ -0,0 +1,187 @@
+import { type Page } from "@playwright/test";
+import fs from "fs";
+import path from "path";
+import sharp from "sharp";
+import { exec } from "child_process";
+
+interface AnnotationBox {
```
kamath: let's use types instead of interfaces unless necessary

lib/vision.ts
```diff
@@ -0,0 +1,187 @@
...
+  x: number;
+  y: number;
+  width: number;
+  height: number;
+  id: string;
+}
+
+interface NumberPosition {
```
kamath: same here (type over interface)

lib/vision.ts
```diff
@@ -0,0 +1,187 @@
...
+  private async createElementAnnotation(
+    id: string,
+    selector: string,
+  ): Promise<string> {
+    try {
+      const element = await this.page.locator(`xpath=${selector}`).first();
+      const box = await element.boundingBox();
+
+      if (!box) {
+        console.log(`[BROWSERBASE] [Debug] No bounding box for element ${id}`);
```
kamath: when are we using `console.log` vs. `this.log`? inconsistent logging here

lib/vision.ts
```diff
@@ -0,0 +1,187 @@
...
+        <rect x="${adjustedBox.x}" y="${adjustedBox.y}" width="${adjustedBox.width}" height="${adjustedBox.height}" 
+              fill="none" stroke="red" stroke-width="2" />
+        <circle cx="${numberPosition.x}" cy="${numberPosition.y}" r="${circleRadius}" fill="white" stroke="red" stroke-width="2" />
+        <text x="${numberPosition.x}" y="${numberPosition.y}" fill="red" font-size="16" font-weight="bold" 
+              text-anchor="middle" dominant-baseline="central">
+          ${id}
+        </text>
+      `;
+    } catch (error) {
+      console.error(
```
kamath: again logging

lib/vision.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    const timestamp = new Date().toISOString().replace(/[:.]/g, "-");
+    const filename = path.join(screenshotDir, `screenshot-${timestamp}.png`);
+
+    fs.writeFileSync(filename, screenshot);
+    console.log(`Screenshot saved to: ${filename}`);
+
+    // Open the screenshot with the default image viewer
+    if (process.platform === "win32") {
+      exec(`start ${filename}`);
```
kamath, line 204: this should be behind a flag, and i'd change `saveAndOpenScreenshot` to `saveScreenshot` with `open` as an optional param default false. if someone's running this in lambda or something they don't need to open the screenshot explicitly

kamath: even if in debug mode, it makes more sense to separate these if they're running it in the cloud

PR #107, opened 2024-10-12
first commit after repointing to main

filip-michalsky:
# why
We need to support i-frames

# what changed

# test plan

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -615,10 +615,12 @@ export class Stagehand {
         });
       } else {
         if (useVision === "fallback") {
-          return this.act({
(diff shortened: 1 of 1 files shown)


filip-michalsky:
this one passes evals but is slow as i-frames are checked sequentially.
I would suggest we merge this one as a sequential i-frame support.
I also have a branch to pack all of the i-frames into one LLM call, but it's not working yet and I have run out of time to work on this now. https://github.com/browserbase/stagehand/pull/108 This one is tricky since its touch the core DOM processing - it passes the example but not the other evals

PR #110, opened 2024-10-16
evaluate extraction progress in separate call

anishk23733:
# why

Changes for improvements to completion checking and progress updating.

The original issue was that in a single extraction step, we provide previously extracted content, progress so far, completed/not completed, DOM elements, etc. all at once and ask the model to give us newly extracted content from DOM elements, new progress, completed/not completed. The issue here is that the model needs to simultaneously decide what is important in the DOM and evaluate its progress based on the information it extracts. 

# what changed

This PR turns this into three steps: 1. extracting the content from the DOM (without previous progress, previous extraction content, etc.), 2. Refining content to ensure that we do not add redundant/duplicate data and updating the total response, and 3. deciding based on the total extracted content what the new progress is and whether the job is completed. The main idea here is that breaking this up rather than doing it all at once shows improvements and a reduction in hallucinations.

# test plan

Running evals. Shows improvement with awareness of the number of outputs in `arxiv` eval required since progress tracking is improved. Is able to do both of the Github evals.

This shows improvement on both the GitHub tests.

### DIFF
--- a/lib/prompt.ts
+++ b/lib/prompt.ts
@@ -108,7 +108,12 @@ export const actTools: Array<OpenAI.ChatCompletionTool> = [
 ];
 
 // extract
-const extractSystemPrompt = `you are extracting content on behalf of a user. You will be given an instruction, progress so far, and a list of DOM elements to extract from. Where applicable, return the exact text from the DOM elements with all symbols, characters and endlines as is. Only extract new information that has not already been extracted. Make sure you include the extraction in your response. Return null or an empty string if no new information is found for a string variable`;
+const extractSystemPrompt = `You are extracting content on behalf of a user. You will be given:
+1. An instruction
+2. A list of DOM elements to extract from
+
+Return the exact text from the DOM elements with all symbols, characters, and endlines as is.
+Only extract new information that has not already been extracted. Return null or an empty string if no new information is found.`;
 
 export function buildExtractSystemPrompt(): OpenAI.ChatCompletionMessageParam {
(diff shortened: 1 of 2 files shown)


evals/index.eval.ts
```diff
@@ -281,7 +281,9 @@ const extract_last_twenty_github_commits = async () => {
 
   try {
     await stagehand.page.goto("https://github.com/facebook/react");
+    await stagehand.waitForSettledDom();
```
kamath: could be wrong here, but i thought we're deprecating this?

lib/prompt.ts
```diff
@@ -120,23 +120,113 @@ export function buildExtractSystemPrompt(): OpenAI.ChatCompletionMessageParam {
...
-    Previously Extracted Content:\n${JSON.stringify(
-      previouslyExtractedContent,
-      null,
-      2,
-    )}
     DOM: ${domElements}`,
   };
 }
 
+// const filterSystemPrompt = `You are tasked with filtering out duplicate information from newly extracted content based on previously extracted content. Return only the new, non-duplicate information.`;
```
kamath: let's not commit commented code. you can always go back through commit history to find this if needed

anishk23733: Oh yes good call, removed that

evals/index.eval.ts
anishk23733: Got it, wasn't aware—removed. Just included since I saw it in some other tests—I figure we need to remove it there as well?

PR #113, opened 2024-10-19
Fm/support iframes fixed

filip-michalsky:
# why
We need to support i-frames

# what changed

# test plan


filip-michalsky:
This is naive, sequential i-frame (optional) processing. We should work on processing of the i-frames into one LLM call as well to reduce recursion.

PR #118, opened 2024-10-26
improve visbility of elements in dom processing

filip-michalsky:
# why

Certain DOMs were failing, namely extraction from here:
https://help.salesforce.com/s/articleView?id=000383278&type=1

# what changed

remove isTopElement check - but this makes other evals fail.

# test plan

### DIFF
--- a/examples/index.ts
+++ b/examples/index.ts
@@ -4,21 +4,30 @@ import { z } from "zod";
 async function example() {
   const stagehand = new Stagehand({
     env: "LOCAL",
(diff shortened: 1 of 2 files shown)


filip-michalsky:
the isVisible function - removing the isTopElement check. However, that makes the complex_peeler eval fail - with the new isVisible function, the peeler eval does not find the search bar :disappointed: I have not been able to find a middle ground which passes both test cases for this:

PR #148, opened 2024-10-29
Enforce TSC Throughout Repo

svapnil:
# why

TSC rules were not uniformly enforce throughout this package

# what changed

Fixed errors throw by TS compiler

# test plan

`tsc`

I'm not sure how to run tests, but will make sure scripts like `wordle work on this

### DIFF
--- a/evals/index.eval.ts
+++ b/evals/index.eval.ts
@@ -1163,13 +1163,13 @@ const exactMatch = (args: { input: any; output: any; expected?: any }) => {
   if (expected === true) {
     return {
       name: "Exact match",
-      score: args.output === true || args.output?._success == true,
+      score: Number(args.output === true || args.output?._success == true),
     };
   }
 
   return {
     name: "Exact match",
-    score: args.output === expected,
+    score: Number(args.output === expected),
   };
 };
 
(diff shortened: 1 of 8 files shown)


pkiv:
Will take a look at this, thanks for opening the PR!

svapnil:
Thanks @pkiv ! Not much changed, but just wrangling down typing mismatches to make sure `tsc` has no complaints, both by forcing types to be more specific and by updating braintrust's sdk

kamath:
Appreciate the effort you put into this! Closing for the following reasons:

1. We don't want to do `as` casting, and we're making a repo-wide push in future PRs to get away from this, since it doesn't actually enforce the type we want. Instead, when we do `const var: type = val` instead of `const var = val as type` we can get an error if `val` doesn't match `type`.
2. We also want to get away from `any`. However, strongly typing client-side code can also get a little hairy, so we want to make sure we get that right instead of writing code just to pass tsc. We don't really get any advantage from assigning `any` type.

There's a few key abstractions we need to get right to enable stuff like BYOM, and the less `any` and `as` we have, the easier we think that'll become. Hope this makes sense!

PR #151, opened 2024-10-31
make stricter completion condition for extract

anishk23733:
# why

The complete condition often fails when there are chunks left but the instruction is completed. This aims to improve it.

# what changed

# test plan

### DIFF
--- a/lib/prompt.ts
+++ b/lib/prompt.ts
@@ -252,8 +252,10 @@ const metadataSystemPrompt = `You are an AI assistant tasked with evaluating the
 Analyze the extraction response and determine if the task is completed or if more information is needed.
 
 Strictly abide by the following criteria:
-1. If you are certain that the instruction is completed, set the completion status to true, even if there are still chunks left.
-2. If there could still be more information to extract and there are still chunks left, set the completion status to false.`;
+1. If the instruction has been satisfied by the current extraction response, ALWAYS set completion status to true and stop processing, regardless of remaining chunks.
+2. Only set completion status to false if BOTH of these conditions are true:
+   - The instruction has not been fully satisfied yet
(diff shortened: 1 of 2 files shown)


lib/prompt.ts
```diff
@@ -252,8 +252,10 @@ const metadataSystemPrompt = `You are an AI assistant tasked with evaluating the
 Analyze the extraction response and determine if the task is completed or if more information is needed.
 
 Strictly abide by the following criteria:
-1. If you are certain that the instruction is completed, set the completion status to true, even if there are still chunks left.
-2. If there could still be more information to extract and there are still chunks left, set the completion status to false.`;
+1. If the instruction has been satisfied by the current extraction response, ALWAYS set completion status to true and stop processing, regardless of remaining chunks.
```
kamath: Nit: I think wording here could be "Once the instruction has been satisfied..." instead of "if". It might help in situations where you have to scroll a little and stop for example

lib/prompt.ts
```diff
@@ -252,8 +252,10 @@ const metadataSystemPrompt = `You are an AI assistant tasked with evaluating the
...
+2. Only set completion status to false if BOTH of these conditions are true:
+   - The instruction has not been satisfied yet
+   - There are still chunks left to process (chunksTotal > chunksSeen)`;
```
kamath, line 258: is there a way to parameterize `chunksTotal` and `chunksSeen`?

lib/prompt.ts
anishk23733: Valid—updated!

lib/prompt.ts
anishk23733: The actual parameters for these are passed in the user prompt since we get their values from the user. Let me know what you think

kamath (approved):

PR #152, opened 2024-11-01
Add: initFromPage method to Stagehand to support external Page objects

ChinmayShrivastava:
# why
I needed to use Stagehand with [Anon](https://www.anon.com/) for authentication purposes, and therefore needed to initialize Stagehand by passing an external [Page](https://playwright.dev/docs/pom) object.

# what changed
I added `initFromPage` method to initialize Stagehand with an external Page.

# test plan
Just want to ensure that this is the best way to achieve the desired results. Not sure if any other tests are required.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -281,6 +281,26 @@ export class Stagehand {
       await this._waitForSettledDom();
       return result;
     };
+
+    // Set the browser to headless mode if specified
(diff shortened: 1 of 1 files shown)


pkiv (approved):

pkiv:
Great change! Thanks for the PR!

PR #155, opened 2024-11-01
fix wait for settled dom always timing out all the time

navidkpr:
# why

# what changed

# test plan

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -407,7 +407,7 @@ export class Stagehand {
           );
           this.log({
             category: "dom",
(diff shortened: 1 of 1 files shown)


lib/index.ts
```diff
@@ -407,7 +407,7 @@ export class Stagehand {
           );
           this.log({
             category: "dom",
-            message: `DOM settle timeout of ${timeout}ms exceeded, continuing anyway`,
+            message: `DOM settle timeoutcx of ${timeout}ms exceeded, continuing anyway`,
```
kamath: timeoutcx?

navidkpr: Fixed!

PR #157, opened 2024-11-02
add caching on the llm provider level

navidkpr:
# why
To save model call costs on tasks that are repeated over and over again.

This is just a first of PRs to enable caching on different layers of Stagehand.

For example, when running evals twice in a row with caching enabled, we get a ~50% cache hit on our LLM calls.

# what changed
Added an LLM Cache class to keep track of previous model calls and their corresponding responses.

# test plan
Run evals with caching on

### DIFF
--- a/lib/llm/LLMCache.ts
+++ b/lib/llm/LLMCache.ts
@@ -93,6 +93,7 @@ export class LLMCache {
         }
 
         fs.writeFileSync(this.lockFile, process.pid.toString(), { flag: "wx" });
+        this.count_lock_acquire_failures = 0;
         this.lock_acquired = true;
         return true;
       } catch (error) {
--- a/lib/llm/LLMCache.ts
+++ b/lib/llm/LLMCache.ts
@@ -1,0 +1,129 @@
+import * as fs from "fs";
+import * as path from "path";
+import * as crypto from "crypto";
+
+interface CacheEntry {
+  timestamp: number;
+  data: any;
+}
+
+interface CacheStore {
+  [key: string]: CacheEntry;
+}
+
+export class LLMCache {
+  private cacheDir: string;
+  private cacheFile: string;
+  private logger: (message: {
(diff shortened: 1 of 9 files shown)


lib/llm/LLMCache.ts
```diff
@@ -0,0 +1,139 @@
...
+    }
+  }
+
+  private createHash(data: any): string {
+    const hash = crypto.createHash("sha256");
+    return hash.update(JSON.stringify(data)).digest("hex");
+  }
+
+  private readCache(): CacheStore {
+    if (fs.existsSync(this.cacheFile)) {
```
filip-michalsky, line 137: I see that we read/write the JSON every time we do an LLM call. what if we have multiple parallel processes? Should we not have the object in memory instead and instantiate higher in the abstraction?

filip-michalsky: we might want to consider Redis down the line i guess- could open as an issue

lib/llm/OpenAIClient.ts
```diff
@@ -1,27 +1,51 @@
...
 
   async createChatCompletion(options: ChatCompletionOptions) {
+    const cacheOptions = {
+      model: options.model,
+      messages: options.messages,
+      temperature: options.temperature,
+      top_p: options.top_p,
+      frequency_penalty: options.frequency_penalty,
+      presence_penalty: options.presence_penalty,
+      image: options.image,
```
filip-michalsky, line 42: hashing the image base64 string might be quite memory intensive - any thoughts on that?

filip-michalsky (approved):
It would be good to add some tests for caching on/off maybe?

lib/llm/LLMCache.ts
edlng: > we might want to consider Redis down the line i guess- could open as an issue

#2239

PR #159, opened 2024-11-02
configure browserbase session via sdk to support full parametrization

blaize-at-psci-labs:
# why

# what changed

# test plan


filip-michalsky:
Github actions currently dont run from forked contributions, as a temporary hack, i ran evals here:
https://github.com/browserbase/stagehand/pull/162

PR #160, opened 2024-11-02
Npour/add llm caching

navidkpr:
# why

# what changed

# test plan


examples/example.ts
filip-michalsky, line 1: pls add 
await stagehand.context.close(); 
to close the browser to end the example

PR #169, opened 2024-11-07
Fix Fuji-Web link in README

hargup:
# why
- The older link was wrong, pointing to a user, instead of the right repo

# what changed
- Replaced wrong link with correct link 

# test plan
- N/A


kamath (approved):
Great catch, thanks!

PR #171, opened 2024-11-08
add variables to act

navidkpr:
# why
Multiple users have asked for a more cachable way to use workflows like "log in to this account".

# what changed
We now allow users to pass in variables to act. Allowing the user to use the same prompt using different variables to increase caching likelihood

# test plan
Test it manually

### DIFF
--- a/evals/index.eval.ts
+++ b/evals/index.eval.ts
@@ -269,7 +269,10 @@ const peeler_complex = async () => {
     await stagehand.page.goto(`https://chefstoys.com/`, { timeout: 60000 });
 
     await stagehand.act({
-      action: "search for peelers",
+      action: "search",
+      variables: {
+        search_query: "peeler",
+      },
     });
 
     await stagehand.act({
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -705,6 +705,7 @@ export class Stagehand {
     verifierUseVision,
     retries = 0,
     requestId,
+    variables,
   }: {
     action: string;
     steps?: string;
@@ -714,6 +715,7 @@ export class Stagehand {
     verifierUseVision: boolean;
     retries?: number;
     requestId?: string;
+    variables: Record<string, string>;
   }): Promise<{ success: boolean; message: string; action: string }> {
     const model = modelName ?? this.defaultModelName;
 
@@ -793,6 +795,7 @@ export class Stagehand {
       screenshot: annotatedScreenshot,
       logger: this.logger,
       requestId,
+      variables,
     });
 
     this.log({
@@ -824,6 +827,7 @@ export class Stagehand {
           useVision,
           verifierUseVision,
           requestId,
+          variables,
         });
       } else if (useVision === "fallback") {
         this.log({
@@ -840,6 +844,7 @@ export class Stagehand {
           useVision: true,
           verifierUseVision,
           requestId,
+          variables,
         });
       } else {
         if (this.enableCaching) {
@@ -915,6 +920,7 @@ export class Stagehand {
               retries: retries + 1,
               chunksSeen,
               requestId,
+              variables,
             });
           }
         }
@@ -945,6 +951,7 @@ export class Stagehand {
               retries: retries + 1,
               chunksSeen,
               requestId,
+              variables,
             });
           }
         }
@@ -969,6 +976,7 @@ export class Stagehand {
               retries: retries + 1,
               chunksSeen,
               requestId,
+              variables,
             });
           }
         }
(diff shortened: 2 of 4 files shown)


evals/index.eval.ts
```diff
@@ -112,7 +112,7 @@ const vanta = async () => {
   for (const observation of observations) {
     try {
       const observationResult = await stagehand.page
-        .locator(observation.selector)
+        .locator(observation.selector[0])
```
pkiv: Will this break existing observe integrations?

evals/index.eval.ts
```diff
@@ -269,7 +269,10 @@ const peeler_complex = async () => {
     await stagehand.page.goto(`https://chefstoys.com/`, { timeout: 60000 });
 
     await stagehand.act({
-      action: "search for peelers",
+      action: "search",
```
pkiv: I really think that this needs to have an API like `await stagehand.act({action: "search for %search_query%", variables: {search_query: "peeler"});`

evals/playground.ts
```diff
@@ -115,12 +121,337 @@ const vanta = async () => {
   };
 };
 
-async function main() {
-  // const homedepotResult = await homedepot();
-  const vantaResult = await vanta();
+const peeler_complex = async () => {
```
pkiv: Can we remove the playground file from the PR, or at least leave the playground blank?

lib/dom/global.d.ts
```diff
@@ -4,17 +4,17 @@ declare global {
     chunkNumber: number;
     processDom: (chunksSeen: Array<number>) => Promise<{
       outputString: string;
-      selectorMap: Record<number, string>;
+      selectorMap: Record<number, string[]>;
```
pkiv, line 7: This is a breaking contract change, but probably NBD because it's so early.

lib/index.ts
```diff
@@ -1,13 +1,15 @@
...
 import fs from "fs";
 import { Browserbase } from "@browserbasehq/sdk";
-import { act, extract, observe, verifyActCompletion } from "./inference";
+import { extract, observe } from "./inference";
 import { AvailableModel, LLMProvider } from "./llm/LLMProvider";
 import path from "path";
 import { ScreenshotService } from "./vision";
 import { modelsWithVision } from "./llm/LLMClient";
+import { ActionCache } from "./cache/ActionCache";
+import { StagehandActHandler } from "./act";
```
pkiv: should we break up extract and observe similarly to keep things consistent?

lib/index.ts
```diff
@@ -558,30 +575,17 @@ export class Stagehand {
...
-  private _generateId(operation: string) {
-    return crypto.createHash("sha256").update(operation).digest("hex");
-  }
-
   private async _recordObservation(
     instruction: string,
     result: { selector: string; description: string }[],
   ): Promise<string> {
-    const id = this._generateId(instruction);
+    const id = generateId(instruction);
```
pkiv, line 587: should the URL be a part of the cache key?

pkiv: or is it? unclear what instruction will be

package.json
```diff
@@ -13,7 +13,7 @@
     "cache:clear": "rm -rf .cache",
     "evals": "npm run build-dom-scripts && npx braintrust eval evals/index.eval.ts",
     "build-dom-scripts": "mkdir -p ./lib/dom/build && mkdir -p ./dist/dom/build && esbuild ./lib/dom/*.ts --bundle --outdir=./dist/dom/build",
-    "bundle-dom-scripts": "esbuild ./lib/dom/index.ts --bundle --outfile=./lib/dom/bundle.js --platform=browser --target=es2015 --minify",
+    "bundle-dom-scripts": "esbuild ./lib/dom/index.ts --bundle --outfile=./lib/dom/bundle.js --platform=browser --target=es2015 --minify && cat ./lib/dom/bundle.js | pbcopy",
```
pkiv: Should remove `&& cat ./lib/dom/bundle.js | pbcopy`

pkiv (changes requested):

pkiv:
After IRL discussion - I'm supportive of the current approach, as long as we add support for string templating variables for better prompt control in the future.

evals/index.eval.ts
pkiv: Navid is going to take the first one for each

lib/index.ts
pkiv: Navid will do this in a separate PR

evals/index.eval.ts
pkiv: Will be addressed in a future PR

pkiv (approved):

PR #172, opened 2024-11-08
browserbaseResumeSessionID to connect to an running bb session

markokraemer:
# why
For my use case I am using keepAlive for browser base sessions. I have an Web-Browser control API I am creating for softgen.ai´s web interaction agent that gets bb session id as param. Stagehand only created new sessions, but I have to reconnect to the existing one. 

# what changed
added browserbaseResumeSessionID?: string, to getBrowser() as an optional param to be able to resume an existing running session. Validation with browserbase sdk to check whether the session is running, throws error if not.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -20,6 +20,7 @@ async function getBrowser(
     level?: 0 | 1 | 2;
   }) => void,
   browserbaseSessionCreateParams?: Browserbase.Sessions.SessionCreateParams,
+  browserbaseResumeSessionID?: string,
 ) {
   if (env === "BROWSERBASE" && !process.env.BROWSERBASE_API_KEY) {
     logger({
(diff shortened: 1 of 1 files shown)


lib/index.ts
```diff
@@ -31,44 +32,67 @@ async function getBrowser(
     env = "LOCAL";
   }
 
-  if (env === "BROWSERBASE" && !process.env.BROWSERBASE_PROJECT_ID) {
```
kamath, line 34: I'm assuming this was removed bc you don't need project ID to connect to a running session, but I don't see this checking for project ID before creating session (see my other comment)

lib/index.ts
```diff
@@ -31,44 +32,67 @@ async function getBrowser(
...
+    } else {
+      // Create new session (existing code)
+      logger({
+        category: "Init",
+        message: "Creating new Browserbase session...",
+        level: 0,
+      });
+      
+      const session = await browserbase.sessions.create({
+        projectId: process.env.BROWSERBASE_PROJECT_ID,
```
kamath, line 79: We're not enforcing project ID to exist anywhere anymore

kamath (changes requested):
Nice work! I think this is almost there. I left a few comments on project ID validation, but I'd also request that while you're here, let's assign `browserbase_project_id` and `browserbase_api_key` to constants at the beginning of `getBrowser` and refer to those. That way, we don't have to repeatedly refer to `process.env` variables that may or may not exist.

kamath:
Merging this into `anirudh/bb-resume-session`; this PR unblocks some other tasks we can complete if we get it over the hump. Thanks for your contributions here!

markokraemer:
Should´ve caught that I deleted the validation. 

Happily : ) Great work on Stagehand <3

markokraemer:
Now you are obligated to add 
- apiKey
- projectId 
to the Config of Stagehand()

It wont get it automatically from .env at all I believe. So I had to update my code.

markokraemer:
If intentional behaviour, make sure to update readme. Else I think its more convenient that it tries to get from .env if none are provided in the constructor

kamath:
Great catch, I submitted #177 to fix this. Should be released soon :)

PR #177, opened 2024-11-10
refresh page

kamath:
# why

# what changed
- add init script on context instead of page
- fix issue in #174 that made passing in API Key and Project ID necessary

# test plan


pkiv (approved):

kamath:
addressed in #178

PR #179, opened 2024-11-13
fix gray outline missing + action completion attr creation

navidkpr:
Fixes:
1. The last big change we pushed out, introduced a small regression. As a result, the gray outline showing the elements Stagehand is looking out is missing. This commit fixes that. We now process selectorMap properly now (using the updated type `Record<number, string[]`

2. Improved the action prompt:
- Improved the structure
- Made it more straightforward
- Improved working for `completed` arg and prioritized precision over recall

### DIFF
--- a/.changeset/little-weeks-worry.md
+++ b/.changeset/little-weeks-worry.md
@@ -1,0 +1,13 @@
+---
+"@browserbasehq/stagehand": minor
+---
+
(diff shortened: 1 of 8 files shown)


kamath (approved):

kamath (approved):
add changeset before merging though

PR #189, opened 2024-11-20
Add structured logging

kamath:
# why
Logs are a mess right now and really hard to debug. This adds structured logging that can be more easily digested and analyzed downstream

# what changed
Added structured JSON logging, most notably in an `auxiliary` field that contains any variable fields for a log

# test plan
Examples work


kamath:
Also need to update `EvalLogger`

pkiv (approved):

PR #191, opened 2024-11-20
fix: no cache dir created when enableCaching is set to false

its-jz:
# why
A `tmp/.cache` directory is created even when enableCaching is set to false. 

# what changed
1. Passing enableCaching all the way down to BaseCache
2. If enableCaching is false, ensureCacheDirectory will not create a directory
3. Added a devdependency because it eems like "@playwright/test" was missing
# test plan

### DIFF
--- a/lib/cache/BaseCache.ts
+++ b/lib/cache/BaseCache.ts
@@ -16,6 +16,7 @@ export class BaseCache<T extends CacheEntry> {
   private readonly CACHE_MAX_AGE_MS = 7 * 24 * 60 * 60 * 1000; // 1 week in milliseconds
   private readonly CLEANUP_PROBABILITY = 0.01; // 1% chance
 
+  protected enableCaching: boolean;
   protected cacheDir: string;
   protected cacheFile: string;
   protected lockFile: string;
@@ -38,10 +39,12 @@ export class BaseCache<T extends CacheEntry> {
       message: string;
       level?: number;
     }) => void,
+    enableCaching: boolean,
     cacheDir: string = path.join(process.cwd(), "tmp", ".cache"),
     cacheFile: string = "cache.json",
   ) {
     this.logger = logger;
+    this.enableCaching = enableCaching;
     this.cacheDir = cacheDir;
     this.cacheFile = path.join(cacheDir, cacheFile);
     this.lockFile = path.join(cacheDir, "cache.lock");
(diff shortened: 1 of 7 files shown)


lib/cache/BaseCache.ts
```diff
@@ -16,6 +16,7 @@ export class BaseCache<T extends CacheEntry> {
   private readonly CACHE_MAX_AGE_MS = 7 * 24 * 60 * 60 * 1000; // 1 week in milliseconds
   private readonly CLEANUP_PROBABILITY = 0.01; // 1% chance
 
+  protected enableCaching: boolean;
```
kamath, line 19: It doesn't make sense to add this field here. Rather, we should just not init a cache if `enableCaching` is set to false in `LLMProvider` and `actHandler`. It makes more sense to make the cache default `undefined` in `LLMProvider` and `actHandler`, and before a cache operation there, check `enableCaching`.

lib/cache/BaseCache.ts
```diff
@@ -71,7 +74,7 @@ export class BaseCache<T extends CacheEntry> {
   }
 
   protected ensureCacheDirectory(): void {
-    if (!fs.existsSync(this.cacheDir)) {
+    if (this.enableCaching && !fs.existsSync(this.cacheDir)) {
```
kamath, line 77: This is an antipattern. Since this method is `protected`, it can only be called from other cache objects. If we just don't init a cache, the directory won't be created

lib/handlers/actHandler.ts
```diff
@@ -273,9 +273,8 @@ export class StagehandActHandler {
 
         this.logger({
           category: "action",
-          message: `Clicked element, ${
```
kamath, line 276: Make sure you format with `prettier` to keep the diff relevant to code changes

package.json
```diff
@@ -27,6 +27,7 @@
   "devDependencies": {
     "@changesets/changelog-github": "^0.5.0",
     "@changesets/cli": "^2.27.9",
+    "@playwright/test": "^1.49.0",
```
kamath, line 30: Why was this added?

kamath (changes requested):

kamath:
Addressing in #197

kamath:
Sincerely appreciate your help and time here in trying to resolve this issue! There might be a simpler short-term fix in #197 in the meantime

Issue #180, opened 2024-11-18
Build DOM scripts so that importing doesn't fail in Next.js


jeremypress:
Reported here: https://stagehand-dev.slack.com/archives/C07UCP76U8G/p1731825603169509

Our DOM scripts are not found when using Next.js due to static imports, these can be built correctly so they are used like other source files in a package.

```
 ⨯ unhandledRejection: Error: ENOENT: no such file or directory, open '<some>/<path>/app/.next/server/dist/dom/build/xpathUtils.js'
    at Generator.next (<anonymous>) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '<some>/<path>/app/.next/server/dist/dom/build/xpathUtils.js'
}
```

PR #192, opened 2024-11-20
Pass in script content to playwright to make running browser scripts easier

jeremypress:
# why

We run a variety of code in the remote browser to do things like DOM pruning and drawing debug boxes. This means the code needs to be built to run in a browser, but it has to be passed into playwright as a separate file/string at runtime. 

Previously, we built the files as vanilla js and then tried to find their path at runtime

https://github.com/browserbase/stagehand/issues/180

1. we didn't like this approach, hand copying files into dist locations is fragile
2. it broke in next.js

# what changed

Simplified a few things
1. We build the code to js, and then export it as a string var that playwright can consume
2. This is just a code file, so we can import it normally
3. we were bundling more than we needed to, so that is now slimmed down. and done as one file instead of 3

# test plan
1. Example runs great
2. I ran the next.js repo example (can upload that)
3. still to do, run evals and non next.js example project


lib/dom/genDomScripts.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * We have a collection of typescript functions that we need to run in the browser.
+ * First, we build them into a single js file
+ * Second, due to framework differences we need to get our script content as a string to avoid pathing issues due to file routing in frameworks like Next.js
+ * Playwright allows us to pass in script content directly as a string instead of reading a file from a path
+ * https://github.com/browserbase/stagehand/issues/180
+ *
+ * We can't rely on the normal build process for stagehand, because we need our script content as a string so that the import *just works*
+ */
+const fs = require("fs");
```
kamath: Do we want to be using `require` over `import`?

lib/dom/genDomScripts.js
```diff
@@ -0,0 +1,30 @@
...
+  bundle: true,
+  outdir: path.join(__dirname, "build"),
+});
+
+const scriptContent = fs.readFileSync(
+  path.join(__dirname, "./build/index.js"),
+  "utf8",
+);
+
+const output = `export const scriptContent = ${JSON.stringify(scriptContent)};`;
```
kamath: is this kosher? seems scary to have stringified js

lib/dom/genDomScripts.js
jeremypress: yeah that was claude, will clean up

lib/dom/genDomScripts.js
jeremypress: it's not super kosher! but if you think about it the only way to get this script into a remote browser is to send it across the "wire" as data anyways right?

kamath: fair haha, just wanted to see if there was a way to have stuff like syntax highlighting. if this is the only way then ig we have no other option

jeremypress: yeah luckily the logic is self is typescript, it's just this bridge to a string that we have less control over

PR #195, opened 2024-11-20
Re-add structured logging

kamath:
Reverts browserbase/stagehand#194
- Adds changeset
- Fixes example


kamath:
#189 was already approved, just added changeset and reverted example

PR #196, opened 2024-11-21
Add evals for each model

kamath:
# why
We want evals for OpenAI and Anthropic. This does that and adds tags for each model if we want to break down performance. 

Also, OpenAIClient wasn't working as expected, for some reason LogLine was being imported from the wrong place. This must've gotten lost in a previous PR, and due to the typing hell we're in, it got lost along the way. This fixes that.

# what changed
- Added evals for both Claude 3.5 Sonnet and GPT 4o for all evals with model name tags
- Fixes `example.ts` to work with OpenAI and Anthropic

# test plan
[Braintrust run](https://www.braintrust.dev/app/Browserbase/p/stagehand/experiments/anirudh%2Ffix%2Fopenai-example-1732153595?c=anirudh/fix/openai-example-1732152948)

Also ran `example.ts` with `modelName` of `gpt-4o`, `claude-3-5-sonnet-20241022`, and empty and they all work

### DIFF
--- a/evals/index.eval.ts
+++ b/evals/index.eval.ts
@@ -1,5 +1,6 @@
 import { Eval } from "braintrust";
 import { Stagehand } from "../lib";
+import type { AvailableModel } from "../lib/llm/LLMProvider";
 import { z } from "zod";
 import process from "process";
 import { EvalLogger } from "./utils";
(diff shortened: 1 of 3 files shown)


pkiv (approved):
l g t m

jeremypress (commented):
looks good, might make sense in the future to have "stable" models that we can pin in the LLM Providers file, then any thing we run could pull stable Anthropic, stable Open AI etc, and not have manually set model names

PR #200, opened 2024-11-21
add tsc

kamath:
# why
We don't have actual typing, and we can't easily compile the project with TSC. There's still a way to go with getting rid of `any` and `as`, but this at least allows us to build the project with `tsc`. 

# what changed
Add stronger typing 

# test plan
Ran tsc and ran example with GPT and Claude


kamath:
unnecessary

PR #202, opened 2024-11-22
Client Parameterization

ChinmayShrivastava:
Add: 
Independent events (act, extract, observe) configured to take in LLMClient objects as parameters as an alternative to the model name and the LLMProvider class.

### DIFF
--- a/lib/inference.ts
+++ b/lib/inference.ts
@@ -24,7 +24,7 @@ import { resolveLLMClient } from "./llm/LLMProvider";
 export async function verifyActCompletion({
   goal,
   steps,
-  llmClient,
+  initllmClient,
   llmProvider,
   modelName,
   screenshot,
@@ -34,7 +34,7 @@ export async function verifyActCompletion({
 }: {
   goal: string;
   steps: string;
-  llmClient?: LLMClient;
+  initllmClient?: LLMClient;
   llmProvider?: LLMProvider;
   modelName?: AvailableModel;
   screenshot?: Buffer;
@@ -42,7 +42,8 @@ export async function verifyActCompletion({
   logger: (message: { category?: string; message: string }) => void;
   requestId: string;
 }): Promise<boolean> {
-  llmClient = resolveLLMClient(llmClient, llmProvider, modelName, requestId);
+  const llmClient =
+    initllmClient || resolveLLMClient({ llmProvider, modelName, requestId });
 
   const messages: ChatMessage[] = [
     buildVerifyActCompletionSystemPrompt(),
@@ -105,7 +106,7 @@ export async function act({
   action,
   domElements,
   steps,
-  llmClient,
+  initllmClient,
   llmProvider,
   modelName,
   screenshot,
@@ -117,7 +118,7 @@ export async function act({
   action: string;
   steps?: string;
   domElements: string;
-  llmClient?: LLMClient;
+  initllmClient?: LLMClient;
   llmProvider?: LLMProvider;
   modelName?: AvailableModel;
   screenshot?: Buffer;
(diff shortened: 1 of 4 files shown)


lib/llm/LLMProvider.ts
```diff
@@ -77,3 +77,20 @@ export class LLMProvider {
     }
   }
 }
+
+export const resolveLLMClient = (
```
kamath: This doesn't need to take in `llmClient`. Also, instead of ordered params, it would be better to have an object passed in with a function signature like `({ llmProvider, modelName, requestId }: { llmClient?: llmClient, modelName?: AvailableModel, requestId?: string })`

lib/inference.ts
```diff
@@ -31,14 +34,16 @@ export async function verifyActCompletion({
...
+  llmClient?: LLMClient;
+  llmProvider?: LLMProvider;
+  modelName?: AvailableModel;
   screenshot?: Buffer;
   domElements?: string;
   logger: (message: { category?: string; message: string }) => void;
   requestId: string;
 }): Promise<boolean> {
-  const llmClient = llmProvider.getClient(modelName, requestId);
+  llmClient = resolveLLMClient(llmClient, llmProvider, modelName, requestId);
```
kamath: It's generally bad practice to re-assign a function param. You can do

`llmClient = initLlmClient || resolveLLMClient(llmProvider, modelName, requestId)`

kamath:
You can also make this a draft PR if not ready to merge

lib/inference.ts
```diff
@@ -17,11 +17,14 @@ import {
...
 import { z } from "zod";
 import { AvailableModel, LLMProvider } from "./llm/LLMProvider";
+import { LLMClient } from "./llm/LLMClient";
 import { AnnotatedScreenshotText, ChatMessage } from "./llm/LLMClient";
+import { resolveLLMClient } from "./llm/LLMProvider";
 
 export async function verifyActCompletion({
   goal,
   steps,
+  llmClient,
```
kamath: Can't recall the exact syntax but you can do something like `llmClient: initLlmClient` here, so that the user passes in an `llmClient` with that name, and we reassign downstream if necessary

ChinmayShrivastava: If we use llmClient for destructuring, that would still create a duplicate variable and throw errors.

ChinmayShrivastava: I will just make the param initllmClient

kamath:
generally LGTM, just wanna make sure we're not pushing commented code

navidkpr:
I just looked through your code.

Would love to know the use case you're trying to cover with this PR. Mainly, does this cover your use case?

Screenshot 2024-11-24 at 12 55 44 AM
Screenshot 2024-11-24 at 12 54 35 AM
Screenshot 2024-11-24 at 12 54 37 AM

kamath:
@navidkpr we're paving the way to detach dependency of VLM on LLM. this is one step towards that, and it was my suggestion to go down this route; what if we get rid of `modelName` here as an arg (beyond initial end-user calls) and instead just pass down `llmClient` objects instead of both `modelName` and `llmClient`?

Plus, we're going to need to pass in additional parameters to the client besides just `modelName` -- consider the use case of parameterizing API key or other factors; if we continue passing `modelName` down, we're restricting LLM client modularity

kamath:
Sorry for the back and forth here @ChinmayShrivastava. We were working on a few things internally that were pressing, and it became easier for us to do it ourselves internally in #221. Sincerely grateful for the hard work you put in here though! Hope this update is helpful; we're going to be cutting a release shortly with some major updates

PR #217, opened 2024-11-24
add o1 support manually by parsing tool usage / schema validation +  add try catch wrappers around logger errors

navidkpr:
add o1 support manually by parsing tool usage / schema validation + add try catch wrappers around logger errors

# why
o1 models do not not support tool usage, temperature and many other things. This PR manually changes the prompt to make up for that.

# what changed
Override parameters for o1 models. Implemented the tool usage and schema validation in the prompt for o1

There were also some standing bugs in evals / logging that I removed.

# test plan
Run evals with the new / old models


evals/utils.ts
```diff
@@ -18,7 +19,16 @@ function parseLogLine(logLine: LogLine): LogLineEval {
           ]),
         )
       : undefined,
-  } as LogLineEval;
+    } as LogLineEval;
+  } catch (error) {
+    console.error(`Error parsing log line:`, logLine);
+
+    return {
```
kamath, line 11: can just return logline here, no?

evals/index.eval.ts
```diff
@@ -1189,7 +1180,6 @@ const arxiv: EvalFunction = async ({ modelName, logger }) => {
               )
               .nullable(),
           }),
-          modelName: "gpt-4o-2024-08-06",
```
kamath, line 1197: ty 🙏

evals/index.eval.ts
```diff
@@ -776,12 +774,8 @@ const google_jobs: EvalFunction = async ({ modelName, logger }) => {
 
     const jobDetails = await stagehand.extract({
       instruction:
-        "Extract the following details from the job posting: application deadline, minimum qualifications (degree and years of experience), and preferred qualifications (degree and years of experience)",
```
kamath, line 784: do we need to remove this? might be better to have a score that shows completion of each field?

lib/handlers/actHandler.ts
```diff
@@ -158,7 +158,8 @@ export class StagehandActHandler {
         goal: action,
         steps,
         llmProvider: this.llmProvider,
-        modelName: model,
+        // o1 is overkill for this task + this task uses a lot of tokens. So we switch it 4o
+        modelName: (model === "o1-mini" || model === "o1-preview") ? "gpt-4o" : model,
```
kamath: not the biggest fan of this approach, but until we find a better one, can we move this into `verifyActCompletion` body instead of having it be a param?

lib/llm/LLMProvider.ts
```diff
@@ -17,6 +19,8 @@ export class LLMProvider {
     "gpt-4o": "openai",
```
kamath, line 14: might make sense to have a type called `ModelVendor: "openai" | "anthropic"` and edit `modelToProviderMap` to have type `{ [key in AvailableModel]: ModelVendor }`

lib/llm/OpenAIClient.ts
```diff
@@ -3,8 +3,11 @@ import { zodResponseFormat } from "openai/helpers/zod";
 import { LLMClient, ChatCompletionOptions } from "./LLMClient";
 import { LLMCache } from "../cache/LLMCache";
 import { LogLine } from "../types";
+import { zodToJsonSchema } from "zod-to-json-schema";
+import { validateZodSchema } from "../utils";
 
 export class OpenAIClient implements LLMClient {
+  public type: "openai" = "openai";
```
kamath, line 12:
```suggestion
  public type: ModelVendor = "openai";
```

lib/prompt.ts
```diff
@@ -206,30 +206,38 @@ const extractSystemPrompt = `You are extracting content on behalf of a user. You
...
-
-ONLY print the content using the print_extracted_data tool provided.
-ONLY print the content using the print_extracted_data tool provided.
 `;
 
-export function buildExtractSystemPrompt(): ChatMessage {
-  const content = extractSystemPrompt.replace(/\s+/g, " ");
+export function buildExtractSystemPrompt(llmClient: LLMClient): ChatMessage {
+  let content = extractSystemPrompt.replace(/\s+/g, " ");
+  if (llmClient.type === "anthropic") {
```
kamath: not a fan of this approach -- would prefer that `buildExtractSystemPrompt` becomes a method of `llmClient`, and `AnthropicClient` implements this there.

lib/prompt.ts
```diff
@@ -206,30 +206,38 @@ const extractSystemPrompt = `You are extracting content on behalf of a user. You
...
   domElements: string,
 ): ChatMessage {
-  return {
-    role: "user",
-    content: `Instruction: ${instruction}
-DOM: ${domElements}
+  let content = `Instruction: ${instruction}
+DOM: ${domElements}`;
 
+  if (llmClient.type === "anthropic") {
```
kamath: see other comment, these methods make more sense to go into `LLMClient`

evals/index.eval.ts
kamath: from huddle: not on page anymore

evals/utils.ts
navidkpr: Done

evals/index.eval.ts
navidkpr: ofc

lib/handlers/actHandler.ts
navidkpr: Done

lib/prompt.ts
navidkpr: I took a bit of a different approach and handled this from outside. Let me know if u like this more.

kamath:
Rebased and continuing in #279

PR #221, opened 2024-11-25
parameterize llmclient

kamath:
# why

Builds off #202 (thanks @ChinmayShrivastava!!) to parameterize LLM Client. Right now, we're passing in `modelName` into very nested functions, which means any additional config to the client, like API Key, also has to be passed down to each function. Instead, it makes more sense to init an LLM Client at the beginning, and pass down that client object.

# what changed
- nothing for the end user
- instead of modelName, we're passing llmClient between functions
- some new types in `types.ts`

# test plan
- evals

### DIFF
--- a/evals/index.eval.ts
+++ b/evals/index.eval.ts
@@ -1,10 +1,9 @@
 import { Eval } from "braintrust";
 import { Stagehand } from "../lib";
-import type { AvailableModel } from "../lib/llm/LLMProvider";
 import { z } from "zod";
 import process from "process";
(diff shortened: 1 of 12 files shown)


navidkpr (approved):
Looks good!

Only one small clean up left. Can you remove `llmProvider` from inference.ts functions?

kamath:
Great catch! addressed [here](https://github.com/browserbase/stagehand/pull/221/commits/4a188507633617177940e931f16397e95b2ef018)

PR #224, opened 2024-11-25
add a short doc on releasing

jtoar:
# why

releasing is scary, but everyone working on stagehand should feel empowered to do it

# what changed

we will add more github workflows to automate the process, but this PR starts small by adding a short doc on how to do a release manually from the CLI

# test plan

i've been testing on my fork


kamath (approved):
thanks for this @jtoar! 🐐

PR #225, opened 2024-11-25
fix tmp user contexts

sameelarif:
# why
Current handling causes errors on Windows, this PR makes handling consistent on both OS.

# what changed
Ensured the `tmp` folder is created in within the project directory, and used the `path` module to ensure it works cross-platform.

# test plan
Run evals on Windows and Mac. Can confirm it's running well on Windows. I don't have access to a MacOS system atm, so that needs to be tested.


kamath:
Have you tried running this with caching? My concern is primarily around that; we use the `tmp` directory for maintaining the local LLM cache

sameelarif:
Can test shortly, though in theory it shouldn't have any effect on LLM caching as they create separate sub directories

sameelarif:
image

Works smoothly as tested with the example script after enabling caching @kamath

PR #232, opened 2024-11-26
Configure release management

kamath:
# why
We now have a more standardized approach to releasing with less manual involvement

# what changed
- Created a release workflow based on [changeset docs](https://github.com/changesets/action)
- Added docs for how releases now work
- Moved from tsup to `esbuild`
- Exposed more types to users
- Removed unnecessary deps like `anthropic` and `anthropic-ai`

# test plan
- Ran in a personal [fork](https://github.com/kamath/stagehand)
- This was able to successfully [publish](https://www.npmjs.com/package/@akamath426/stagehand?activeTab=versions) both alpha version on merge to main and [open a PR](https://github.com/kamath/stagehand/pulls?q=is:pr+%22Version+Packages%22) for the official release
- When the PR was merged, it successfully updated the version and tagged a release to npm

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -9,11 +9,12 @@
   "changelog": [
     "@changesets/changelog-github",
     {
-      "repo": "browserbase/stagehand"
+      "repo": "kamath/stagehand"
     }
   ],
   "snapshot": {
     "useCalculatedVersion": true,
-    "prereleaseTemplate": "{tag}-{datetime}-{commit}"
+    "prereleaseTemplate": "alpha-{commit}",
+    "tag": "alpha"
   }
 }
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -9,7 +9,7 @@
   "changelog": [
     "@changesets/changelog-github",
     {
-      "repo": "kamath/stagehand"
+      "repo": "browserbase/stagehand"
     }
   ],
   "snapshot": {
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,6 +24,9 @@ jobs:
       - name: Install Dependencies
         run: npm install
 
+      - name: Build
+        run: npm run build
+
       - name: Create Release Pull Request or Publish to npm
         id: changesets
         uses: changesets/action@v1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,42 +1,43 @@
-name: "release"
+name: Release
+
 on:
   push:
-    tags:
-      - "v*"
+    branches:
+      - main
+
+concurrency: ${{ github.workflow }}-${{ github.ref }}
 
 jobs:
-  publish:
-    runs-on: "ubuntu-22.04"
-    permissions:
-      id-token: write
-      contents: write
-      packages: write
-      pull-requests: write
-      issues: read
+  release:
+    name: Release
+    runs-on: ubuntu-latest
     steps:
-      - uses: "actions/checkout@v3"
+      - name: Checkout Repo
+        uses: actions/checkout@v3
+
+      - name: Setup Node.js 18.x
+        uses: actions/setup-node@v3
         with:
-          fetch-depth: 0
-      - uses: "actions/setup-node@v3"
+          node-version: 18.x
+          registry-url: "https://registry.npmjs.org"
+
+      - name: Install Dependencies
+        run: npm install
+
+      - name: Create Release Pull Request or Publish to npm
+        id: changesets
+        uses: changesets/action@v1
         with:
-          node-version: 18
-          cache: "npm"
-      - name: "Cancel previous runs"
-        uses: "styfle/cancel-workflow-action@0.11.0"
-        continue-on-error: true
-        with:
-          access_token: ${{ secrets.GITHUB_TOKEN }}
-      - run: "npm i"
-      - run: "npm run build"
-      - name: "release / pull_request"
-        id: "changesets"
-        uses: "dotansimha/changesets-action@v1.5.2"
-        with:
-          publish: "npm run release"
-          version: "npm exec changeset version"
-          commit: "Update stagehand version"
-          title: "Upcoming Release Changes"
-          createGithubReleases: "aggregate"
+          publish: npm run release
         env:
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Publish Canary
+        if: github.ref == 'refs/heads/main'
+        run: |
+          git checkout main
+          npm version prerelease --preid=canary.${GITHUB_SHA::8} --no-git-tag-version
+          npm publish --tag canary --access public
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,10 +15,10 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v3
 
-      - name: Setup Node.js 18.x
+      - name: Setup Node.js 20.x
         uses: actions/setup-node@v3
         with:
-          node-version: 18.x
+          node-version: 20.x
           registry-url: "https://registry.npmjs.org"
 
       - name: Install Dependencies
@@ -40,7 +40,7 @@ jobs:
         if: github.ref == 'refs/heads/main'
         run: |
           git checkout main
-          npm version prerelease --preid=canary.${GITHUB_SHA::8} --no-git-tag-version
-          npm publish --tag canary --access public
+          npm run release-canary
         env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,6 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: 20.x
-          registry-url: "https://registry.npmjs.org"
 
       - name: Install Dependencies
         run: npm install
--- a/README.md
+++ b/README.md
@@ -33,8 +33,7 @@
 - [Acknowledgements](#acknowledgements)
 - [License](#license)
 
-> [!NOTE]
-> `Stagehand` is currently available as an early release, and we're actively seeking feedback from the community. Please join our [Slack community](https://join.slack.com/t/stagehand-dev/shared_invite/zt-2tdncfgkk-fF8y5U0uJzR2y2_M9c9OJA) to stay updated on the latest developments and provide feedback.
+> [!NOTE] > `Stagehand` is currently available as an early release, and we're actively seeking feedback from the community. Please join our [Slack community](https://join.slack.com/t/stagehand-dev/shared_invite/zt-2tdncfgkk-fF8y5U0uJzR2y2_M9c9OJA) to stay updated on the latest developments and provide feedback.
 
 ## Intro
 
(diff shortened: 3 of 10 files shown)


.github/workflows/release.yml
```diff
@@ -1,42 +1,46 @@
...
+        uses: actions/setup-node@v3
         with:
-          access_token: ${{ secrets.GITHUB_TOKEN }}
-      - run: "npm i"
-      - run: "npm run build"
-      - name: "release / pull_request"
-        id: "changesets"
-        uses: "dotansimha/changesets-action@v1.5.2"
+          node-version: 18.x
+          registry-url: "https://registry.npmjs.org"
```
jtoar: this is a fairly old version of node; is the changeset action asking for it?

docs/release.md
jtoar, line 1: it may be valuable to keep the original contents around and just frame them under "manually releasing a version from the cli should you have to"

.github/workflows/release.yml
```diff
@@ -1,42 +1,46 @@
...
+      - name: Create Release Pull Request or Publish to npm
+        id: changesets
+        uses: changesets/action@v1
         with:
-          publish: "npm run release"
-          version: "npm exec changeset version"
-          commit: "Update stagehand version"
-          title: "Upcoming Release Changes"
-          createGithubReleases: "aggregate"
+          publish: npm run release
```
jtoar, line 33: i think you need to build the package before releasing it right? i'll reread the action, but pretty sure this will result in publishing an empty tarball

.github/workflows/release.yml
```diff
@@ -1,42 +1,46 @@
...
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Publish Canary
+        if: github.ref == 'refs/heads/main'
+        run: |
+          git checkout main
+          npm version prerelease --preid=canary.${GITHUB_SHA::8} --no-git-tag-version
+          npm publish --tag canary --access public
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
jtoar: i'd leverage changesets to do this too (pretty sure it has config for ["snapshot" releases](https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md) or something similar), and we also need to build the tarball here before releasing

.github/workflows/release.yml
jtoar: looks like it uses node20: https://github.com/changesets/action/blob/31ff97e8acd3cfd08714eab3721f55da10003255/action.yml#L4

.github/workflows/release.yml
jtoar: looks like publish is supposed to build and publish: https://github.com/changesets/action/blob/31ff97e8acd3cfd08714eab3721f55da10003255/action.yml#L7-L8

jtoar: an example from changesets as well: https://github.com/changesets/changesets/blob/f03f70bffc9a033747f45b0d071c6c2ed8818ac7/.github/workflows/changeset-version.yml#L25-L26

jtoar (changes requested):
we're headed in the right direction but don't think this one will work just yet

.github/workflows/release.yml
kamath: Our current releases were using node v18; I overrode the version for consistency. Can def make it node 20

docs/release.md
kamath: ack

.github/workflows/release.yml
kamath: I might've committed this after you commented, but we do run `npm run build` right before this

.changeset/config.json
```diff
@@ -14,6 +14,7 @@
   ],
   "snapshot": {
     "useCalculatedVersion": true,
-    "prereleaseTemplate": "{tag}-{datetime}-{commit}"
+    "prereleaseTemplate": "alpha-{commit}",
```
kamath, line 17: This does ths whole commit -- I saw suggestions for `shortCommit` or `{commit:6}` but neither of these worked. I think this should be fine, just not elegant ig

package.json
```diff
@@ -13,8 +13,11 @@
     "cache:clear": "rm -rf .cache",
     "evals": "npm run build-dom-scripts && tsx evals/index.eval.ts",
     "build-dom-scripts": "tsx lib/dom/genDomScripts.ts",
-    "build": "npm run build-dom-scripts && tsup lib/index.ts --dts",
-    "release": "changeset publish"
+    "build-types": "tsc --emitDeclarationOnly --outDir dist",
+    "build-js": "esbuild lib/index.ts --bundle --platform=node --outdir=dist --format=esm",
```
kamath: Not an expert here on tsup v. esbuild, I think this is right? Would love some confirmation

tsconfig.json
```diff
@@ -12,7 +12,8 @@
     "paths": {
       "*": ["node_modules/*", "lib/types/*"]
     },
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "declaration": true
```
kamath, line 16: I think this is a net positive, again would love confirmation

examples/debugUrl.ts
```diff
@@ -8,8 +8,6 @@ async function debug(url: string) {
   });
   await stagehand.init();
   await stagehand.page.goto(url);
-
-  await stagehand.startDomDebug();
```
kamath, line 12: private method

package.json
```diff
@@ -13,12 +13,14 @@
     "cache:clear": "rm -rf .cache",
     "evals": "npm run build-dom-scripts && tsx evals/index.eval.ts",
     "build-dom-scripts": "tsx lib/dom/genDomScripts.ts",
-    "build": "npm run build-dom-scripts && tsup lib/index.ts --dts",
-    "release": "changeset publish"
+    "build-types": "tsc --emitDeclarationOnly --outDir dist",
+    "build-js": "esbuild lib/index.ts --bundle --platform=node --outdir=dist --format=esm --minify --tree-shaking=true --packages=external",
```
kamath: Would love a second pair of eyes here to make sure I did this right

.github/workflows/release.yml
kamath: I _think_ this should be addressed

package.json
```diff
@@ -13,12 +13,14 @@
...
+    "build": "npm run build-dom-scripts && npm run build-js && npm run build-types",
+    "release": "npm run build && changeset publish",
+    "release-canary": "npm run build && changeset version --snapshot && changeset publish --tag alpha"
   },
   "files": [
-    "dist/**",
-    "patches/**"
```
kamath, line 21: unnecessary

lib/index.ts
```diff
@@ -872,3 +872,9 @@ export class Stagehand {
       });
   }
 }
+
```
kamath, line 881: these are all the types we also want public-facing; there's a few more that are just for internal uses, like `inference` that aren't here

docs/release.md
```diff
@@ -2,6 +2,22 @@
 
 We use [Changesets](https://github.com/changesets/changesets) to version and release our packages.
 
+When we merge to main, the release workflow will:
+
+1. Create a release pull request with:
+   - A version bump for the package calculated by the changesets.
+   - A changelog entry summarizing the changes in the release.
+1. Create an `alpha` version of the package on whatever is merged to main, and you can install it with `npm install @browserbasehq/stagehand@alpha`. This is useful for testing the release before it's published to the `latest` tag.
```
jtoar:
```suggestion
1. Create an `alpha` version of the package with whatever is merged to main, and you can install it with `npm install @browserbasehq/stagehand@alpha`. This is useful for testing the release before it's published to the `latest` tag.
```

docs/release.md
```diff
@@ -2,6 +2,22 @@
...
+
+When the pull request is merged, the release workflow will publish the package to npm with the version calculated by the changesets.
+
+For more information on how changesets work, see the [changesets docs](https://github.com/changesets/changesets) and our [release.yml file](/.github/workflows/release.yml).
+
+# Manually Releasing
+
+> **Warning**
```
jtoar: this will render nicer on github

```suggestion
> [!WARNING]
```

README.md
```diff
@@ -33,8 +33,7 @@
 - [Acknowledgements](#acknowledgements)
 - [License](#license)
 
-> [!NOTE]
-> `Stagehand` is currently available as an early release, and we're actively seeking feedback from the community. Please join our [Slack community](https://join.slack.com/t/stagehand-dev/shared_invite/zt-2tdncfgkk-fF8y5U0uJzR2y2_M9c9OJA) to stay updated on the latest developments and provide feedback.
+> [!NOTE] > `Stagehand` is currently available as an early release, and we're actively seeking feedback from the community. Please join our [Slack community](https://join.slack.com/t/stagehand-dev/shared_invite/zt-2tdncfgkk-fF8y5U0uJzR2y2_M9c9OJA) to stay updated on the latest developments and provide feedback.
```
jtoar: probably want to revert this one, doesn't look like it's rendering correctly

.github/workflows/release.yml
```diff
@@ -1,42 +1,46 @@
...
+        uses: actions/setup-node@v3
         with:
-          access_token: ${{ secrets.GITHUB_TOKEN }}
-      - run: "npm i"
-      - run: "npm run build"
-      - name: "release / pull_request"
-        id: "changesets"
-        uses: "dotansimha/changesets-action@v1.5.2"
+          node-version: 20.x
+          registry-url: "https://registry.npmjs.org"
```
jtoar: will double check but don't think this is necessary

```suggestion
```

.github/workflows/release.yml
```diff
@@ -1,42 +1,46 @@
...
         with:
-          publish: "npm run release"
-          version: "npm exec changeset version"
-          commit: "Update stagehand version"
-          title: "Upcoming Release Changes"
-          createGithubReleases: "aggregate"
+          publish: npm run release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
jtoar, line 36: minor but i'd rename this to `NPM_` since that's what it's used for

```suggestion
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```

package.json
jtoar: one thing that sticks out is i think this package is commonjs but the output format is esm; i'm not sure if we actually use anything commonjs specific, or if this package is commonjs for a reason, but we should be thorough

package.json
```diff
@@ -13,12 +13,15 @@
     "cache:clear": "rm -rf .cache",
     "evals": "npm run build-dom-scripts && tsx evals/index.eval.ts",
     "build-dom-scripts": "tsx lib/dom/genDomScripts.ts",
-    "build": "npm run build-dom-scripts && tsup lib/index.ts --dts",
-    "release": "changeset publish"
+    "build-types": "tsc --emitDeclarationOnly --outDir dist",
+    "build-js": "esbuild lib/index.ts --bundle --platform=node --outdir=dist --format=esm --minify --tree-shaking=true --packages=external",
```
jtoar: i'd recommend against minifying; i think it's important to keep the dist readable, debuggable. users can always minify the package if they're bundling before deploying to their production environment if size is a concern

```suggestion
    "build-js": "esbuild lib/index.ts --bundle --platform=node --outdir=dist --format=esm --tree-shaking=true --packages=external",
```

jtoar: tree shaking is already enabled according to the docs:

```suggestion
    "build-js": "esbuild lib/index.ts --bundle --platform=node --outdir=dist --format=esm --minify --packages=external",
```

https://esbuild.github.io/api/#tree-shaking

jtoar: the `--packages=external` piece doesn't seem to be working as intended. i'm seeing the whole `@browserbasehq/sdk` included in the dist. may have to do with relative imports or something similar. maybe we want `--external:*`. the intention is to exclude dependencies but bundle the files here right?

jtoar (approved):
let's remove the build step changes for now and address those separately

docs/release.md
kamath: 786b6974afb339d4dbbd63ba3acd27f87c00d5ab

docs/release.md
kamath: 786b6974afb339d4dbbd63ba3acd27f87c00d5ab

README.md
kamath: 67903a30a5a76136a4a58ac848efdac313dfd716

.github/workflows/release.yml
kamath: i think the env var needs to be `NODE_AUTH_TOKEN` iirc it wasn't working as `NPM_TOKEN`

.github/workflows/release.yml
kamath: da995a23e13f3ccf69cde8df7fab290d7e5e40fa

PR #233, opened 2024-11-27
Llm typing improvements

sameelarif:
# why
Type casting leaves room for error, we should chip away at that to enforce stricter typing

# what changed
Format & validate objects passed into `LLMClient` instances to make sure data sent is always correct

# test plan
Run evals


kamath:
Generally lgtm, but before we merge I want to add some stuff like `eslint` to the project. This is a big change, and there's some potentially functional changes like removing function args, and I want to make sure there's no breaking changes or other logic errors that slip through the cracks

PR #234, opened 2024-11-27
reduce chunk size

seanmcguire12:
# why
when processing candidate elements, we process the webpage in chunks, and check if an element is visible. if we use the exact viewport height to determine the chunk size, and there are banners/nav menus that are fixed to the top (or bottom) of the webpage, then some candidate elements (that should be included) will be hidden under these elements as we chunk/scroll. 
# what changed
we set the `viewportHeight` (used to determine chunk height) to be 75% of `window.innerHeight` instead of using the full `window.innerHeight`.
# test plan
can add some evals to show where candidate elements were missing before 
# notes
this may be more of a temporary/quick fix. We may want to investigate dynamically detecting the presence of fixed elements (like banners, nav menus, etc) and determining the chunk height based on their height

### DIFF
--- a/lib/dom/process.ts
+++ b/lib/dom/process.ts
@@ -27,7 +27,7 @@ export async function processDom(chunksSeen: Array<number>) {
 export async function processAllOfDom() {
   console.log("Stagehand (Browser Process): Processing all of DOM");
 
-  const viewportHeight = window.innerHeight;
+  const viewportHeight = window.innerHeight * 0.75;
   const documentHeight = document.documentElement.scrollHeight;
   const totalChunks = Math.ceil(documentHeight / viewportHeight);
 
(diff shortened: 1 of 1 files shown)


filip-michalsky:
@seanmcguire12 this looks good, but a small question - the way we calculate how many chunks are at a page:

https://github.com/browserbase/stagehand/blob/main/lib/dom/process.ts#L413

still uses the full `window.innerHeight`, so, if there is some important stuff in the footer, we might never reach it if we truncate scrolling by 25%?

I agree with this approach for not missing selectors, but what about the footer stuff edge case?

filip-michalsky:
> @seanmcguire12 this looks good, but a small question - the way we calculate how many chunks are at a page:
> 
> https://github.com/browserbase/stagehand/blob/main/lib/dom/process.ts#L413
> 
> still uses the full `window.innerHeight`, so, if there is some important stuff in the footer, we might never reach it if we truncate scrolling by 25%?
> 
> I agree with this approach for not missing selectors, but what about the footer stuff edge case?

alas, I did not scroll through the whole PR to see you did adjust this line as well. Sorry. LGTM

PR #236, opened 2024-11-27
Reduce chunk size 

seanmcguire12:
# why
when processing candidate elements, we process the webpage in chunks, and check if an element is visible. if we use the exact viewport height to determine the chunk size, and there are banners/nav menus that are fixed to the top (or bottom) of the webpage, then some candidate elements (that should be included) will be hidden under these elements as we chunk/scroll. 
# what changed
we set the `viewportHeight` (used to determine chunk height) to be 75% of `window.innerHeight` instead of using the full `window.innerHeight`.
# test plan
can add some evals to show where candidate elements were missing before 
# notes
this may be more of a temporary/quick fix. We may want to investigate dynamically detecting the presence of fixed elements (like banners, nav menus, etc) and determining the chunk height based on their height


lib/dom/debug.ts
```diff
@@ -101,7 +101,7 @@ function setupChunkNav() {
       cleanupMarkers();
       cleanupNav();
       window.chunkNumber -= 1;
-      window.scrollTo(0, window.chunkNumber * window.innerHeight);
+      window.scrollTo(0, window.chunkNumber * (window.innerHeight * 0.75));
```
kamath: Could use `viewportHeight` here?

lib/dom/debug.ts
```diff
@@ -128,7 +128,7 @@ function setupChunkNav() {
       cleanupMarkers();
       cleanupNav();
       window.chunkNumber += 1;
-      window.scrollTo(0, window.chunkNumber * window.innerHeight);
+      window.scrollTo(0, window.chunkNumber * (window.innerHeight * 0.75));
```
kamath: `viewportHeight` unless I'm missing something -- otherwise we can make `.75` a constant just in case we want to change that later

kamath:
Can we make a function in `dom/utils.ts` called `calculateViewportHeight()` or otherwise just set `window.viewportHeight` in `utils.ts` to not repeat the same constants?

kamath (changes requested):

kamath:
ESLint is going to flag for no magic numbers when we add it, just trying to avoid future headaches

lib/dom/utils.ts
```diff
@@ -15,3 +15,7 @@ export async function waitForDomSettle() {
 }
 
 window.waitForDomSettle = waitForDomSettle;
+
+export function calculateViewportHeight() {
+  return Math.ceil(window.innerHeight * 0.75);
```
kamath, line 20: ESLint might still throw an error here, we can cross that bridge when we get to it

kamath (approved):

PR #237, opened 2024-11-28
Improve AI Grant Example

jsj:
# why
https://github.com/browserbase/stagehand/pull/235 is not easy to work because it's an an image. Update to the code inline with md and the example in `examples/`

# what changed
Update markdown, New example in `examples/`, remove image

# test plan
Documentation


kamath:
Thanks so much for taking the time to do this! However, I don't think this needs to be merged. 

The image is moreso to show how simple Stagehand is to write than Playwright, and imo it does a fair job encapsulating the difference with a side-by-side comparison that gets lost when you make it plain text. 

We could have it as an example, but our existing examples do a fine job showing the capabilities of Stagehand. There's not much to be gained from an extra example. I believe @seanmcguire12 is adding something like this as an eval as well, which is where I would recommend this be moved instead.

PR #238, opened 2024-11-28
run an individual eval by its name

seanmcguire12:
# why
easier to test individual evals
# how to run
run this command in the cli: `npm run evals -- eval_name` and replace `eval_name` with the name of the eval you want to run. eg., `npm run evals -- extract_github_stars`


kamath (approved):
Thanks for this!!

PR #249, opened 2024-11-30
Fix broken evals

seanmcguire12:
# why
- `extract_last_twenty_github_commits` fails with Anthropic because the token limit is too small
- `extract_github_stars` fails with Anthropic because we are expecting the rounded version of the number of stars ~230000, but claude sometimes outputs the precise number, eg., 229,757
# what changed
- increased Anthropic token limit from 1500 to 3000
- added a tolerance of +/- 1000 in `extract_github_stars` so it still passes if the precise number of stars is given

### DIFF
--- a/.changeset/odd-seahorses-beam.md
+++ b/.changeset/odd-seahorses-beam.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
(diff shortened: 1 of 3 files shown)


kamath (approved):
great catches! make sure to add a changeset before merging

PR #252, opened 2024-12-02
Gemeni Models integration

subashrijal5:
## Why
- We need to expand our LLM client support to include Google's Generative AI alongside our existing Anthropic client

## What Changed
- Added a new `GoogleClient` class in the LLM clients directory
- Implemented New model client and provider  for the Gemeni models.  "gemini-1.5-pro", "gemini-1.5-flash" and "gemini-1.5-flash-8b"
##. Note
- gemini-1.5-pro might throw 429 resource exhausted error on free plan because they have rate limit of only 32k tokens. Check more about it here. https://ai.google.dev/gemini-api/docs/models/gemini#gemini-1.5-pro 
- it adds Google generative ai package. https://github.com/google-gemini/generative-ai-js


kamath:
Thanks so much for taking the time to do this! Curious if you ran any evals already on the different models and how they performed? Transparently, we're trying to limit our scope of LLMs for now to figure out how we want to better incorporate extensibility, so there's a good chance we might refactor a lot of this code. That being said, don't want your contributions and time spent to go to waste, so just want to make sure this has meaningful value-add over OpenAI/Anthropic

examples/example.ts
```diff
@@ -9,7 +9,12 @@ async function example() {
     enableCaching: false,
   });
 
-  await stagehand.init();
+  await stagehand.init({
+    modelName: "gemini-1.5-flash",
```
vladionescu, line 13: Recommend updating this default to the newly released "gemini-2.0-flash-exp"

vladionescu:
Gemini has an [OpenAI-compatible API endpoint](https://ai.google.dev/gemini-api/docs/openai), as well as their own client SDK which is implemented here. I prefer using their own SDK, just flagging in case it becomes unstable (Google has the [genai](https://pypi.org/project/google-genai/) Python SDK in beta alongside the [generativeai](https://pypi.org/project/google-generativeai/) SDK...).

> Thanks so much for taking the time to do this! Curious if you ran any evals already on the different models and how they performed? Transparently, we're trying to limit our scope of LLMs for now to figure out how we want to better incorporate extensibility, so there's a good chance we might refactor a lot of this code. That being said, don't want your contributions and time spent to go to waste, so just want to make sure this has meaningful value-add over OpenAI/Anthropic

@kamath in my unscientific experiments, Gemini 2.0 Flash is capable, excels at following instructions (less fiddling with system prompts needed), and very fast. Definitely a contender for this usecase.

kamath:
Thanks @vladionescu! We'll keep this in mind. We're investing heavily into eval + debugging tools rn, so once we have a good baseline, we're likely to use a more standardized model router. I'm hesitant to add more models just yet until we have a better sense of failure modes -- ie did code fail due to Stagehand or a model

kamath:
Hey! Closing this since we have #352 that allows you to [add your own custom LLMClient](https://github.com/browserbase/stagehand/blob/main/examples/external_client.ts). 

We also have #382 in flight to add vercel AI SDK, which would then support gemini as well. 

Thank you so much for the time and effort that went into this PR though!!! Sincerely appreciate it :)

PR #256, opened 2024-12-03
Try to fix releases

kamath:
# why
Continuing #232, need to mess around with permissions config until this works as expected. This worked in a personal repo, but organizations are a little peskier.

# what changed
Publish canary before PR to see if that step works

# test plan
Merge to main


kamath:
No changeset necessary; this doesn't affect main package

PR #269, opened 2024-12-03
add eslint

sameelarif:
# why
Currently lots of type ambiguity, this should make development more strict.

# what changed
Added and fixed many types, removed `any` casting.

# test plan
Run evals to ensure no logic has been damaged, otherwise `npx eslint .` should return 0 errors.

### DIFF
--- a/evals/index.eval.ts
+++ b/evals/index.eval.ts
@@ -1529,10 +1529,10 @@ const tasks: Record<string, EvalFunction> = {
   extract_press_releases,
 };
 
-interface EvalArgs<TInput, TOutput> {
+interface EvalArgs<TInput, TOutput, TExpected> {
   input: TInput;
   output: TOutput;
-  expected: unknown;
+  expected: TExpected;
   metadata?: { model: AvailableModel; test: string };
 }
 
--- a/evals/index.eval.ts
+++ b/evals/index.eval.ts
@@ -1514,6 +1514,7 @@ const tasks: Record<string, EvalFunction> = {
   peeler_simple,
   peeler_complex,
   wikipedia,
+  nonsense_action,
   simple_google_search,
   extract_github_stars,
   extract_collaborators_from_github_repository,
--- a/evals/index.eval.ts
+++ b/evals/index.eval.ts
@@ -51,7 +51,7 @@ type EvalFunction = (args: {
   logs: LogLine[];
   debugUrl: string;
   sessionUrl: string;
-  error?: any;
+  error?: unknown;
 }>;
 
 const expedia: EvalFunction = async ({ modelName, logger }) => {
@@ -1526,24 +1526,43 @@ const tasks: Record<string, EvalFunction> = {
   arxiv,
   expedia,
   amazon_add_to_cart,
-  extract_press_releases
+  extract_press_releases,
 };
 
-const exactMatch = (args: {
-  input: any;
-  output: any;
-  expected?: any;
-}): {
+interface EvalArgs<TInput, TOutput> {
+  input: TInput;
+  output: TOutput;
+  expected: unknown;
+  metadata?: { model: AvailableModel; test: string };
+}
+
+interface EvalResult {
   name: string;
   score: number;
-} => {
+}
+
+interface EvalInput {
+  name: string;
+  modelName: AvailableModel;
+}
+
+const exactMatch = (
+  args: EvalArgs<EvalInput, boolean | { _success: boolean }, unknown>,
+): EvalResult => {
   console.log(`Task "${args.input.name}" returned: ${args.output}`);
 
   const expected = args.expected ?? true;
   if (expected === true) {
     return {
       name: "Exact match",
-      score: args.output === true || args.output?._success == true ? 1 : 0,
+      score:
+        typeof args.output === "boolean"
+          ? args.output
+            ? 1
+            : 0
+          : args.output._success
+            ? 1
+            : 0,
     };
   }
 
@@ -1553,19 +1572,21 @@ const exactMatch = (args: {
   };
 };
 
-const errorMatch = (args: {
-  input: any;
-  output: any;
-  expected?: any;
-}): {
-  name: string;
-  score: number;
-} => {
+const errorMatch = (
+  args: EvalArgs<
+    EvalInput,
+    boolean | { _success: boolean; error?: unknown },
+    unknown
+  >,
+): EvalResult => {
   console.log(`Task "${args.input.name}" returned: ${args.output}`);
 
   return {
     name: "Error rate",
-    score: args.output?.error !== undefined ? 1 : 0,
+    score:
+      typeof args.output === "object" && args.output.error !== undefined
+        ? 1
+        : 0,
   };
 };
 
@@ -1585,17 +1606,26 @@ const testcases = [
   "laroche_form",
   "arxiv",
   "amazon_add_to_cart",
-  "extract_press_releases"
+  "extract_press_releases",
   // "expedia"
 ];
 
-const generateSummary = async (summary: any, results: any[]) => {
+const generateSummary = async (
+  summary: { scores?: Record<string, { score: number | null }> },
+  results: Array<{
+    input: EvalInput;
+    output?: boolean | { _success: boolean };
+  }>,
+) => {
   const exactMatch = summary.scores?.["Exact match"] || { score: null };
 
   const taskStatuses = results.map((result) => ({
     name: result.input.name,
     modelName: result.input.modelName,
-    success: result.output?._success || false,
+    success:
+      typeof result.output === "boolean"
+        ? result.output
+        : result.output?._success || false,
   }));
 
   const totalTasks = taskStatuses.length;
@@ -1614,7 +1644,10 @@ const generateSummary = async (summary: any, results: any[]) => {
     failedTasks,
   };
 
-  fs.writeFileSync("eval-summary.json", JSON.stringify(formattedSummary, null, 2));
+  fs.writeFileSync(
+    "eval-summary.json",
+    JSON.stringify(formattedSummary, null, 2),
+  );
   console.log("Evaluation summary written to eval-summary.json");
 };
 
@@ -1636,6 +1669,7 @@ const filter = args[0];
               model,
               test,
             },
+            expected: true,
           })),
         );
 
@@ -1705,4 +1739,3 @@ const filter = args[0];
     process.exit(1);
   }
 })();
-
--- a/lib/cache/BaseCache.ts
+++ b/lib/cache/BaseCache.ts
@@ -5,7 +5,7 @@ import { LogLine } from "../../types/log";
 
 export interface CacheEntry {
   timestamp: number;
-  data: any;
+  data: unknown;
   requestId: string;
 }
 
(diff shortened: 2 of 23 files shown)


evals/index.eval.ts
```diff
@@ -1514,6 +1514,7 @@ const tasks: Record<string, EvalFunction> = {
   peeler_simple,
   peeler_complex,
   wikipedia,
+  nonsense_action,
```
kamath, line 1517: did we validate this works as expected?

evals/index.eval.ts
```diff
@@ -1585,17 +1607,26 @@ const testcases = [
...
 ];
 
-const generateSummary = async (summary: any, results: any[]) => {
+const generateSummary = async (
+  summary: { scores?: Record<string, { score: number | null }> },
+  results: Array<{
+    input: EvalInput;
+    output?: boolean | { _success: boolean };
+  }>,
+) => {
```
kamath, line 1620:
```suggestion
  results: {
    input: EvalInput;
    output?: boolean | { _success: boolean };
  }[],
) => {
```

evals/index.eval.ts
```diff
@@ -1553,19 +1573,21 @@ const exactMatch = (args: {
...
-  output: any;
-  expected?: any;
-}): {
-  name: string;
-  score: number;
-} => {
+const errorMatch = (
+  args: EvalArgs<
+    EvalInput,
+    boolean | { _success: boolean; error?: unknown },
```
kamath, line 1579: Not for this PR, but every eval should return `{ _success: boolean; error?: unknown }`. cc @seanmcguire12

evals/index.eval.ts
```diff
@@ -1636,6 +1670,7 @@ const filter = args[0];
               model,
               test,
             },
+            expected: true,
```
kamath, line 1673: nice!

lib/dom/debug.ts
```diff
@@ -3,8 +3,9 @@ import { calculateViewportHeight } from "./utils";
 export async function debugDom() {
   window.chunkNumber = 0;
 
-  const { selectorMap: multiSelectorMap, outputString } =
```
kamath, line 6: Maybe not for this PR, but what happens if we remove `outputString` entirely?

lib/handlers/actHandler.ts
```diff
@@ -421,7 +424,7 @@ export class StagehandActHandler {
         await Promise.race([
           this.stagehand.page.waitForLoadState("networkidle"),
           new Promise((resolve) => setTimeout(resolve, 5_000)),
-        ]).catch((e: Error) => {
+        ]).catch(() => {
           this.logger({
```
kamath, line 428: can we add `trace` and `message` from the error to `auxiliary` in the logger like we have for other error logs? just to add color to these logs

lib/cache/BaseCache.ts
```diff
@@ -124,7 +124,7 @@ export class BaseCache<T extends CacheEntry> {
           level: 1,
         });
         return true;
-      } catch (error) {
+      } catch {
```
kamath, line 127: can you add a log here with `trace` and `message` from `error` in `auxiliary`? again to add color

lib/handlers/observeHandler.ts
```diff
@@ -97,12 +97,15 @@ export class StagehandObserveHandler {
 
     await this.waitForSettledDom(domSettleTimeoutMs);
     await this.startDomDebug();
-    let { outputString, selectorMap } = await this.stagehand.page.evaluate(
```
kamath, line 100: we can eslint-ignore this line, it's cleaner this way

lib/index.ts
```diff
@@ -391,7 +393,7 @@ export class Stagehand {
 
     // Overload the page.goto method
     const originalGoto = this.page.goto.bind(this.page);
-    this.page.goto = async (url: string, options?: any) => {
+    this.page.goto = async (url: string, options) => {
```
kamath, line 396: why not add `options?: GotoOptions` as below?

lib/index.ts
```diff
@@ -544,7 +546,7 @@ export class Stagehand {
               (log) => log.id !== logObj.id,
             );
         })
-        .catch((e) => {
+        .catch(() => {
```
kamath, line 549: error logs pls 🙏

lib/inference.ts
```diff
@@ -33,13 +33,17 @@ export async function verifyActCompletion({
...
+  const verificationSchema = z.object({
+    completed: z.boolean().describe("true if the goal is accomplished"),
+  });
 
-  const response = await llmClient.createChatCompletion({
-    messages,
+  type VerificationResponse = z.infer<typeof verificationSchema>;
+
+  const response = await llmClient.createChatCompletion<VerificationResponse>({
+    messages: [
```
kamath, line 43: 🔥

lib/llm/AnthropicClient.ts
```diff
@@ -79,7 +87,7 @@ export class AnthropicClient extends LLMClient {
             },
           },
         });
-        return cachedResponse;
+        return cachedResponse as T;
```
kamath, line 90: can we do something like `const cacheOptions: AnthropicTransformedResponse | undefined = this.cache.get...` to avoid `as` forced typecast

lib/llm/AnthropicClient.ts
```diff
@@ -99,13 +107,39 @@ export class AnthropicClient extends LLMClient {
...
     }
 
-    const systemMessage = options.messages.find((msg) => msg.role === "system");
+    const systemMessage = options.messages.find(
+      (msg) =>
+        msg.role === "system" &&
+        (typeof msg.content === "string" ||
+          (Array.isArray(msg.content) &&
+            msg.content.every((content) => content.type !== "image_url"))),
+    );
```
kamath, line 116: let's clean this up, this is not easy to read. make this an if/else chain

lib/llm/AnthropicClient.ts
```diff
@@ -99,13 +107,39 @@ export class AnthropicClient extends LLMClient {
...
+        };
+      } else {
+        return {
+          role: msg.role as "user" | "assistant",
+          content: msg.content.map((content) => {
+            if (content.type === "text") {
+              return { type: "text", text: content.text };
+            }
+
+            // don't handle image content for now
```
kamath, line 135: elaborate on why not

lib/llm/AnthropicClient.ts
```diff
@@ -138,22 +178,20 @@ export class AnthropicClient extends LLMClient {
...
       // TODO (kamath): fix this forced typecast
       const schemaProperties =
         (
           jsonSchema.definitions?.MySchema as {
-            properties?: Record<string, any>;
+            properties?: Record<string, unknown>;
           }
         )?.properties ||
-        (jsonSchema as { properties?: Record<string, any> }).properties;
+        (jsonSchema as { properties?: Record<string, unknown> }).properties;
```
kamath, line 194: clean up

lib/llm/AnthropicClient.ts
```diff
@@ -261,7 +299,7 @@ export class AnthropicClient extends LLMClient {
           this.cache.set(cacheOptions, result, options.requestId);
         }
 
-        return result;
+        return result as T;
```
kamath, line 302: I like how earlier you said "we can cast because..." - either add that or do something like `const result: AnthropicTransformedResponse`

lib/llm/AnthropicClient.ts
```diff
@@ -309,6 +347,6 @@ export class AnthropicClient extends LLMClient {
       });
     }
 
-    return transformedResponse;
+    return transformedResponse as T;
```
kamath, line 350: see earlier comments about `as`

lib/llm/OpenAIClient.ts
```diff
@@ -138,10 +147,47 @@ export class OpenAIClient extends LLMClient {
       },
     });
 
-    const response = await this.client.chat.completions.create({
+    delete openAiOptions.requestId;
+    delete openAiOptions.image;
```
kamath, line 151: would strongly prefer not to have `delete`; instead even if tedious, let's make a new variable with all the same fields sans `requestId` and `image`

lib/llm/OpenAIClient.ts
```diff
@@ -138,10 +147,47 @@ export class OpenAIClient extends LLMClient {
...
+      options.messages.map((message) => {
+        if (Array.isArray(message.content)) {
+          const contentParts = message.content.map((content) => {
+            if ("image_url" in content) {
+              return {
+                image_url: {
+                  url: content.image_url.url,
+                },
+                type: "image_url",
+              } as ChatCompletionContentPartImage;
```
kamath, line 163: assign to variable of type `ChatCompletionContentPartImage` and return/reference that variable instead of `as`

lib/llm/OpenAIClient.ts
```diff
@@ -138,10 +147,47 @@ export class OpenAIClient extends LLMClient {
...
+            } else {
+              return {
+                text: content.text,
+                type: "text",
+              } as ChatCompletionContentPartText;
```
kamath, line 168: same here, no `as`

lib/llm/OpenAIClient.ts
```diff
@@ -138,10 +147,47 @@ export class OpenAIClient extends LLMClient {
...
+            }
+          });
+
+          return {
+            ...message,
+            content: contentParts,
+          } as ChatCompletionMessageParam;
```
kamath, line 175: no `as`

lib/llm/OpenAIClient.ts
```diff
@@ -138,10 +147,47 @@ export class OpenAIClient extends LLMClient {
...
+        }
+
+        return message as ChatCompletionMessageParam;
```
kamath, line 178: no `as`

lib/llm/OpenAIClient.ts
```diff
@@ -201,6 +247,6 @@ export class OpenAIClient extends LLMClient {
       this.cache.set(cacheOptions, response, options.requestId);
     }
 
-    return response;
+    return response as T;
```
kamath, line 250: no `as`

kamath (changes requested):
Amazing work. I think this is almost there, just want to make sure we're not introducing more problems. This is a pretty sizable PR, so I'd highly recommend making another PR to go into `sameel/add-eslint` so it's easier to see the specific diffs I highlighted in my comments.

evals/index.eval.ts
sameelarif: It's intended to be there according to @filip-michalsky. Lot's of changes coming to evals soon though so I think it's fine to keep as-is for now

lib/dom/debug.ts
sameelarif: It's already removed in this PR

sameelarif: Definitely a possibility. It's not used anywhere.

evals/index.eval.ts
filip-michalsky: it's *supposed* to validate that stagehand cannot find an action to take and then exits peacefully when I originally wrote it... the eval itself will need to be updated bc it probably does not work as expected and @sameelarif will crush the new eval suite either way

sameelarif:
Merging into [eslint-final](https://github.com/browserbase/stagehand/tree/eslint-final) to unify changes with other branch

PR #272, opened 2024-12-04
Organize evals

sameelarif:
# why
Evals are messy right now and hard to maintain.

# what changed
Tests are now split into folders to categorize what is is that they're testing. We also automatically handle what used to be manual entries, so all you have to do to add a new eval is export a new function from within any of the category folders.

# test plan
Ensure eval commands work:
```bash
npm run evals -- <eval file name>
npm run evals category - <category name>
```


.github/workflows/ci.yml
```diff
@@ -4,9 +4,10 @@ on:
   push:
     branches:
       - main
+      - organize-evals
```
kamath, line 8: this will only run when you push to `main` or `organize-evals`. We want this to run on a push to any branch ([see Slack thread](https://browserbase.slack.com/archives/C074FSMUDMF/p1733425862438849))

.github/workflows/ci.yml
```diff
@@ -33,63 +34,151 @@ jobs:
...
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+
+      - name: Install dependencies
+        run: npm install --no-frozen-lockfile
+
+      - name: Install Playwright browsers
```
kamath, line 177: do we need to do this if running on bb?

.github/workflows/ci.yml
```diff
@@ -5,8 +5,12 @@ on:
     branches:
       - main
```
kamath, line 6: This is only going to run when we merge to `main` - check out [workflow docs](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-pull_request-workflow-based-on-files-changed-in-a-pull-request) -- would be awesome to have this run on push to a branch like `dev/*`

.github/workflows/ci.yml
```diff
@@ -5,8 +5,12 @@ on:
...
 
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
```
kamath, line 17: are we going to have another `yml` file for running experimental evals? don't see them being triggered anywhere here

kamath (changes requested):
Looking good, let's get a sample run on this PR so we can test outputs.

.github/workflows/ci.yml
seanmcguire12: yeah wasnt sure if we wanted to be running these in CI. We might want to make this a separate yml and only have it run on merge into main

kamath (approved):
lgtm, let's change this PR to merge into dev to ensure CI works there as well, then let's get a fast follow cleanup PR in to remove the organize-evals branch in CI, and merge dev into main :)

PR #274, opened 2024-12-04
Feat: Add ESLint

sameelarif:
# why
#269 got to a pretty gnarly diff, so in order to resolve comments there, we made #273 and merged it into a branch called `eslint-final`. This PR should theoretically add eslint and much more rigorous typing to Stagehand.

# what changed
Rigorous internal type system. No more `any`, very limited `as` casting

# test plan


kamath (commented):
LGTM, let's get some sanity check evals to run on this before merging :)

kamath (approved):

PR #278, opened 2024-12-06
improve evals

sameelarif:
# what changed
Streamlined process of adding new evals. Also added a schedule to the CI, so that we get performance metrics every 6 hours. Additionally, we introduced a variety of new evals that help us test specific functions. Here's the breakdown of the new folder structure:
`evals`
- `act` - `act()`-specific tests
- `extract` - `extract()`-specific tests
- `observe` - `observe()`-specific tests
- `combination` - tests that evaluate how multiple methods work together
- `experimental` - tests that don't currently work. this is essentially a to-do list for us to move into the other folders eventually.


kamath (approved):
let it rip

PR #281, opened 2024-12-08
eslint - final version

sameelarif:
# why
#269 got to a pretty gnarly diff, so in order to resolve comments there, we made #273 and merged it into a branch called `eslint-final`. This PR should theoretically add eslint and much more rigorous typing to Stagehand.

# what changed
Rigorous internal type system. No more `any`, very limited `as` casting


kamath:
Changed base to evals/eslint at risk of redundancy so we run CI

kamath (approved):

PR #283, opened 2024-12-09
customize eval config from env

sameelarif:
# why
We need to be able to easily select which models/categories to evaluate on local/CI without modifying the source code.

# what changed
We now read models and categories from .env


evals/index.eval.ts
```diff
@@ -2,30 +2,47 @@ import { Eval } from "braintrust";
 import fs from "fs";
 import path from "path";
 import process from "process";
-import { EvalFunction } from "../types/evals";
+import { EvalCategory, EvalFunction } from "../types/evals";
 import { AvailableModel } from "../types/model";
 import { EvalLogger, env } from "./utils";
 
-const models: AvailableModel[] = ["gpt-4o", "claude-3-5-sonnet-latest"];
+const MODELS: AvailableModel[] = process.env.EVAL_MODELS.split(",").map(
```
kamath: let's do `const DEFAULT_EVAL_MODELS = process.env.EVAL_MODELS ? 
process.env.EVAL_MODELS.split(",") : ["gpt-4o", "claude-3-5-sonnet-latest"]`

and then reference `DEFAULT_EVAL_MODELS` directly. SAme with below. Good practice to not directly reference env vars

kamath (changes requested):

types/model.ts
```diff
@@ -2,14 +2,18 @@ import type { ClientOptions as AnthropicClientOptions } from "@anthropic-ai/sdk"
...
+import { z } from "zod";
 
-export type AvailableModel =
-  | "gpt-4o"
-  | "gpt-4o-mini"
-  | "gpt-4o-2024-08-06"
-  | "claude-3-5-sonnet-latest"
-  | "claude-3-5-sonnet-20241022"
-  | "claude-3-5-sonnet-20240620";
+export const AvailableModel = z.enum([
```
kamath: Rename to `AvailableModelSchema` for clarity

types/evals.ts
```diff
@@ -12,3 +13,13 @@ export type EvalFunction = (args: {
   sessionUrl: string;
   error?: any;
 }>;
+
+export const EvalCategory = z.enum([
```
kamath: rename to `EvalCategorySchema` for clarity

kamath (changes requested):
Almost there, let's just avoid having two things share the same name. Also lint + build isn't passing

evals/index.eval.ts
```diff
@@ -85,10 +99,22 @@ if (args.length > 0) {
   }
 }
 
-const models: AvailableModel[] =
+const DEFAULT_EVAL_MODELS = process.env.EVAL_MODELS
```
kamath, line 102: can we instead have like `DEFAULT_EVAL_MODELS` and `EXPERIMENTAL_EVAL_MODELS`? the filter can then be parameterized too and it's easier to read

kamath (approved):

PR #285, opened 2024-12-09
add stockx eval + fix it

navidkpr:
# why
Stagehand was failing on this task before

# what changed

# test plan


kamath:
merging into `evals/npour/stockx-eval` to run evals

kamath (approved):

PR #286, opened 2024-12-09
add stockx eval + fix it (#285)

kamath:
Running evals from #285. Code changes already approved there, will merge without approval if evals pass


kamath:
merging since combo evals itself is broken, fixed in #284

PR #290, opened 2024-12-09
Evals/string similarity

seanmcguire12:
# why
- extract evals sometimes fail when there is a small difference between expected output and actual output (like an extra space, punctuation, capital letters, etc
- we can try to mitigate this by normalizing strings with regex, but it is difficult to cover all possible cases. regex is brittle and hard to understand.
# what changed
- for specific evals where this was a problem, instead of using plain string matching, I've added a string similarity function and set a default threshold of 0.85
# test plan


kamath (approved):

kamath:
Down to merge despite combo evals not passing since everything else is passing and we need to make combo better

PR #201, opened 2024-11-22
pass in text-rendered-webpage instead of outputString

seanmcguire12:
# why
ideally, this improves upon the following issues with `extract`:
- ensure the whole webpage is processed instead of stopping processing too early
- improve overall accuracy of `extract`, ie, for larger extraction tasks, collect more of the correct data and less of the irrelevant/incorrect data 
- provide the model/LLM with a more accurate depiction of what the webpage actually looks like (in text form)

# what changed
Instead of passing HTML to the LLM, we now pass a text-rendered representation of the webpage. Ideally this gives the LLM more context about how the webpage is actually organized, and reduces unnecessary tokens (HTML tags, indices, etc)

# test plan
- create more evals for larger extractions with multiple fields in the schema
- create more evals on complex websites, (ecommerce sites, javascript heavy sites, etc)

PR #293, opened 2024-12-11
Extract improvement

seanmcguire12:
from #201 

# why
ideally, this improves upon the following issues with `extract`:
- ensure the whole webpage is processed instead of stopping processing too early
- improve overall accuracy of `extract`, ie, for larger extraction tasks, collect more of the correct data and less of the irrelevant/incorrect data 
- provide the model/LLM with a more accurate depiction of what the webpage actually looks like (in text form)

# what changed
Instead of passing HTML to the LLM, we now pass a text-rendered representation of the webpage. Ideally this gives the LLM more context about how the webpage is actually organized, and reduces unnecessary tokens (HTML tags, indices, etc)

# test plan
- create more evals for larger extractions with multiple fields in the schema
- create more evals on complex websites, (ecommerce sites, javascript heavy sites, etc)

### DIFF
--- a/types/stagehand.ts
+++ b/types/stagehand.ts
@@ -22,11 +22,6 @@ export interface ConstructorParams {
   modelClientOptions?: ClientOptions;
 }
 
-export interface InitResult {
-  debugUrl: string;
-  sessionUrl: string;
-}
-
 export interface InitOptions {
   /** @deprecated Pass this into the Stagehand constructor instead. This will be removed in the next major version. */
   modelName?: AvailableModel;
@@ -39,6 +34,7 @@ export interface InitOptions {
 export interface InitResult {
   debugUrl: string;
   sessionUrl: string;
+  sessionId: string;
 }
 
 export interface InitFromPageOptions {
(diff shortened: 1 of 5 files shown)


lib/inference.ts
```diff
@@ -160,6 +160,7 @@ export async function extract({
   chunksSeen,
   chunksTotal,
   requestId,
+  isUsingTextExtract,
```
kamath: where is this being passed in?

types/stagehand.ts
```diff
@@ -69,6 +69,7 @@ export interface ExtractOptions<T extends z.AnyZodObject> {
   modelName?: AvailableModel;
   modelClientOptions?: ClientOptions;
   domSettleTimeoutMs?: number;
+  useTextExtract?: boolean;
```
kamath: need to add to lib/index.ts to parameterize what we're sending to inference.ts

kamath:
Need to add changeset and actually allow for parameterizing `textExtract`. Let's default `textExtract` to false for now but announce/publicize `textExtract`. Also add to README

kamath:
Also in this PR or a fast-follow, can you add `textExtract` as a param to experimental evals? Would love to experiment with those results

lib/inference.ts
seanmcguire12: being passed in on line 459 in `extractHandler.ts` inside `domExtract`

PR #296, opened 2024-12-12
Tidy: deprecate + rename init/constructor

kamath:
# why
We have params being passed in to `init` that are also in the constructor. This is redundant and doesn't make sense. Cleaning this up. Also deprecating `initFromPage` in favor of `browserbaseResumeSessionId`. Lastly, change `browserBase` to `browserbase`.

Now you can just do 

```
const stagehand = new Stagehand({ yourParams });
await stagehand.init()
```

# what changed
param names + docs

# test plan
evals CI

### DIFF
--- a/.changeset/gorgeous-readers-rush.md
+++ b/.changeset/gorgeous-readers-rush.md
@@ -1,0 +1,7 @@
+---
+"@browserbasehq/stagehand": minor
+---
+
+- Deprecate fields in `init` in favor of constructor options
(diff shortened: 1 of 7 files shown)


seanmcguire12 (approved):

ilias-t:
is there a migration guide somewhere for this change?

> Deprecate `initFromPage` in favor of `browserbaseResumeSessionID` in constructor

PR #303, opened 2024-12-13
Evals for TextExtract

seanmcguire12:
# why
- separate PR for evals made for #304 
# what changed
- added some new evals in a new dir called `text_extract`
- this means `text_extract` can be run as an eval category
- some evals in `text_extract` are duplicated (copies exist in `experimental`)
- once #304 is merged, you'll be able to run any eval or category of evals and set the extraction method
# example:
- running a specific eval using `text_extract` would look like: `npm run evals -- extract_aigrant_companies --extract-method=textExtract`
- running a category of evals using `text_extract` would look like: `npm run evals category extract --extract-method=textExtract`
- if an extract method is not specified, the default extract method is `domExtract`


.github/workflows/ci.yml
```diff
@@ -243,7 +243,7 @@ jobs:
             exit 1
           fi
 
-  run-experimental-evals:
```
kamath, line 246: can we have `run-text-extract-evals` run right after extract evals? we could also have two separate extract CI, one `run-text-extract` and one `run-dom-extract`. the way it's currently setup, CI to skip running text extract evals if combo evals fail. we can also add text extract as a param for combo/evals, but priority is future PRs shouldn't break text extract or dom extract.

kamath (approved):

PR #308, opened 2024-12-15
Fix press release eval

seanmcguire12:
# why
src website change
# what changed
make expected values tolerant to changes


evals/text_extract/extract_press_releases.ts
```diff
@@ -39,10 +40,10 @@ export const extract_press_releases: EvalFunction = async ({
       }),
       modelName,
       useTextExtract,
-    });
+    })) as { items: { title: string; publish_date: string }[] };
```
kamath: can we do `satsifies` instead of `as`

kamath (approved):

PR #316, opened 2024-12-17
chore: rename browserbaseResumeSessionID -> browserbaseSessionID

kamath:
# why
user convenience

# what changed
`browserbaseResumeSessionId` -> `browserbaseResumeSessionID`

# test plan
lint/build

### DIFF
--- a/.changeset/flat-avocados-sit.md
+++ b/.changeset/flat-avocados-sit.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": minor
+---
+
+rename browserbaseResumeSessionID -> browserbaseSessionID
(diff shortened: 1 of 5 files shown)


pkiv (approved):

jtoar (commented):
to avoid this being a breaking change you probably want to still accept `browserbaseResumeSessionID`, just mark it as deprecated (can do in the release notes and via typescript js/tsdoc [`@deprecated`](https://jsdoc.app/tags-deprecated) i believe) and map it internally to `browserbaseSessionID`. you may have to also handle the case where both are passed, and `browserbaseSessionID` would take precedence

seanmcguire12 (approved):

kamath:
@jtoar i'm ok with that. the publicity that it's `browserbase` and not `browserBase` might actually be good for the brand

PR #317, opened 2024-12-17
Setting model defaults in LLMProvider

dannypaz:
# why

I ran through the examples in the README and set the ANTHROPIC_API_KEY but the code yelled at me for OpenAI. I've added defaults to LLMProvider to check for the existence of API keys and added helpful error messages if api keys are not specified

# what changed

- if clientOption.apiKey is not set, then we check environment variables (OPENAI_API_KEY or ANTHROPIC_API_KEY) and if they do not exist we send a friendly error
- we default model in LLMProvider instead of Stagehand Constructor
- we default model to OpenAI unless there is an ANTHROPIC api key set. If there is, then we default to anthropic

# test plan

- Set ANTHROPIC_API_KEY (unset OPENAI_API_KEY) and run example. Code should init correctly
- Do not set any api keys and run example. Code should give error about openai api key
- Set `modelName` to 'anthropic` and do not set api keys. Code should give error about anthropic key
- Test that clientOptions for both openai and anthropic work correctly

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@
 - [Acknowledgements](#acknowledgements)
 - [License](#license)
 
-> [!NOTE]  
+> [!NOTE]
 > `Stagehand` is currently available as an early release, and we're actively seeking feedback from the community. Please join our [Slack community](https://join.slack.com/t/stagehand-dev/shared_invite/zt-2tdncfgkk-fF8y5U0uJzR2y2_M9c9OJA) to stay updated on the latest developments and provide feedback.
 
 ## Intro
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -63,13 +63,15 @@ npm install @browserbasehq/stagehand zod
...
 
-Ensure that an OpenAI API Key or Anthropic API key is accessible in your local environment.
+Ensure that an OpenAI API Key or Anthropic API key is accessible in your local environment (only one is required).
 
 ```
 export OPENAI_API_KEY=sk-...
 export ANTHROPIC_API_KEY=sk-...
 ```
 
+NOTE: Stagehand client will default to openai if these are not specified
```
kamath, line 73:
```suggestion
NOTE: Stagehand client will default to OpenAI if these are not specified
```

kamath (approved):
wow, thanks so much! can you add a changeset for this? it should just be `npx changeset`. i can also take this over the finish line if you don't wanna do that by EOD. great work!

PR #319, opened 2024-12-18
[2.0] Wrap StagehandPage and StagehandContext

kamath:
# why
We want Stagehand to provide a better Playwright experience, so when the user runs `stagehand.page` we can return a playwright `Page` wrapper as `StagehandPage`

# what changed
- Rewrite `examples/example.ts` to just be a scratchpad. Good Stagehand code can be found with `npx create-browser-app`
- Wrap `stagehand.page` in a `StagehandPage` object
- Wrap `stagehand.context` in a `StagehandContext` object
- Add end-to-end (e2e) Playwright tests in `evals/deterministic` that test Playwright + Browserbase SDK. Added to CI as well before AI evals

# test plan
- Made a deterministic e2e test that can be run with `npm run e2e`. This just runs Playwright tests using `Stagehand`


evals/deterministic/auxiliary/logo.png
seanmcguire12, line 1: fire

seanmcguire12 (approved):

PR #322, opened 2024-12-19
Refactor index.evals.ts

seanmcguire12:
# why
- `evals/index.evals.ts` and `evals/utils.ts` became unnecessarily big with poorly defined scopes and hard to read code.
# what changed
- this PR aims to break down these files into smaller chunks, grouping functions into files that are share a common purpose
# test plan
- evals


evals/initStagehand.ts
```diff
@@ -0,0 +1,75 @@
+/**
```
kamath, line 1: we have a file called `stagehand.config.ts` in `examples/`. Let's follow that pattern here as well.

kamath (approved):

PR #324, opened 2024-12-21
V2

kamath:
# why
We want act/extract/observe to be at the page level to give us the ability to modify the Playwright page as needed. act/extract/observe at the stagehand level will be deleted in v2, but it's deprecated as of now

# what changed
- Move act/extract/observe into `stagehand.page`, forward `stagehand.act/extract/observe` into `stagehand.page`, and deprecate act/extract/observe at the stagehand level. 
- Also add a bunch of new deterministic e2e tests using base playwright to make sure we're not breaking anything fundamental in playwright itself

# test plan
- evals!

### DIFF
--- a/types/page.ts
+++ b/types/page.ts
@@ -3,5 +3,5 @@ import { ActResult } from "./act";
 import { ActOptions } from "./stagehand";
 
 export interface Page extends PlaywrightPage {
-  act?: (options: ActOptions) => Promise<ActResult>;
+  act: (options: ActOptions) => Promise<ActResult>;
 }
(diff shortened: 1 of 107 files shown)


types/page.ts
```diff
@@ -0,0 +1,17 @@
...
+import type { ActResult } from "./act";
+import type {
+  ActOptions,
+  ExtractOptions,
+  ExtractResult,
+  ObserveOptions,
+  ObserveResult,
+} from "./stagehand";
+import type { z } from "zod";
+export interface Page extends PlaywrightPage {
```
sameelarif, line 11: should be named `StagehandPage` imo to reduce potential confusion

sameelarif (approved):

PR #325, opened 2024-12-21
Only startDomDebug if it's enabled

pkiv:
# why
* DomDebug was being enabled even if it was disabled

# what changed
* Add a matching check just like cleanupDomDebug has

# test plan
* Tested locally


kamath:
honestly was thinking we just 🪓 it altogether. don't think anybody uses this, and it gets in the way

pkiv:
@kamath your call - I don't have time to make that bigger change ATM so would recommend getting this in to fix it for now.

PR #327, opened 2024-12-23
add create-browser-app cli to stagehand

kamath:
# why
We can use `create-browser-app` to create a stagehand app. Want to add this here for consistency. Will likely need another PR where we move Stagehand-specific code into a `stagehand/` or `src/` directory; this is getting pretty messy.

# what changed
nothing to users. just added source for `create-browser-app` cli

# test plan
works on npm


kamath:
closing, doesn't need to be in this repo

PR #331, opened 2024-12-23
add extract evals for stagehand.page

kamath:
evals that build off of #329 on `stagehand.page` instead of `stagehand` for extract


seanmcguire12 (approved):
approved pending CI

Issue #268, opened 2024-12-03
Return token usage per function call


kamath:
We need better telemetry on how many input/output tokens went into each Stagehand function so we can help people better measure cost with Stagehand and help facilitate cost-benefit analysis of using smaller/cheaper models over flagship models

ConProgramming:
Definite plus one here. It'd be fantastic to be able to do a `.totalCost()` at the end of a run, to see how economical using Stagehand for scale is.

andrewgcodes:
I've submitted a PR https://github.com/browserbase/stagehand/pull/339

cmaycumber:
This would be awesome. Following this.

PR #339, opened 2024-12-27
token usage tracking (Issue #268)

andrewgcodes:
# why
being able to track input/output/total tokens when using stagehand can be helpful for controlling costs, picking which model to use, etc.

See Issue #268 (https://github.com/browserbase/stagehand/issues/268)

# what changed
- Added LLMUsageEntry interface to track tokens per function call
- Implemented usage recording in OpenAI and Anthropic clients
- Added recordUsage() and getUsage() methods to Stagehand class
- Added getTotalTokensUsed() for aggregating total token usage

Works for act(), observe(), and extract()
12 files edited, added tests and and example (hackernews.ts) demonstrating how to log tokens used. Please see diff

# test plan
i wrote a few test cases and it seems to work but someone on the browserbase team should definitely test as well before merging!

### DIFF
--- a/examples/hackernews.ts
+++ b/examples/hackernews.ts
@@ -1,0 +1,95 @@
+import { Stagehand } from "../lib";
+import { z } from "zod";
+
+function logTokenUsage(
+  functionName: string,
+  entry: {
+    promptTokens: number;
+    completionTokens: number;
+    totalTokens: number;
+  },
+) {
+  console.log(
+    `\n\x1b[1m${functionName} Token Usage:\x1b[0m
+    \x1b[36mPrompt Tokens:     ${entry.promptTokens.toString().padStart(6)}\x1b[0m
+    \x1b[32mCompletion Tokens: ${entry.completionTokens.toString().padStart(6)}\x1b[0m
+    \x1b[33mTotal Tokens:      ${entry.totalTokens.toString().padStart(6)}\x1b[0m`,
+  );
+}
+
+async function scrapeHackerNews() {
+  console.log("🚀 Starting Hacker News scraper...");
+  const stagehand = new Stagehand({
+    env: "LOCAL",
+    verbose: 2, // Maximum verbosity
+    debugDom: true,
+    enableCaching: false,
+    modelName: "gpt-4o",
+    headless: false,
+    modelClientOptions: {
+      apiKey: process.env.OPENAI_API_KEY,
+    },
+  });
+
+  try {
+    console.log("🌟 Initializing Stagehand...");
+    await stagehand.init();
+
+    console.log("🌐 Navigating to Hacker News...");
+    await stagehand.page.goto("https://news.ycombinator.com");
+
+    // Extract top article info
+    const schema = z.object({
+      title: z.string(),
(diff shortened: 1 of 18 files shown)


kamath:
whoa!! thanks @andrewgcodes! i'll take a more detailed look at this in a bit, but thank you so much for taking the time to do this!

examples/hackernews.ts
```diff
@@ -0,0 +1,95 @@
...
+    });
+
+    // Log token usage for extraction
+    const extractUsage = stagehand.getUsage();
+    const extractEntry = extractUsage.find(
+      (entry) => entry.functionName === "extract",
+    );
+    if (extractEntry) {
+      logTokenUsage("HN-ARTICLE-EXTRACT", extractEntry);
+    }
```
kamath: this isn't really elegant; ideally the user should be able to have all of this in one line

kamath (changes requested):
Looking more, this is a phenomenal PR, and I sincerely appreciate you adding test cases.  

However, we're trying super hard to keep the Stagehand interface as minimal as possible to avoid adding unnecessary complexity to the end-user. Having a user write out the code I highlighted feels unnecessarily cumbersome. 

I acknowledge the frustration around token usage being a black box, and it's something we're actively fixing before the holiday break ends. We as the Browserbase team need to do better about this. 

I'm imagining an interface to act/extract/observe with something like the following:

```
const result = page.extract(extractOptions) # returns { ...extractResult, _stagehandTokenUsage: TokenUsageObject }
```

Let's talk more in #stagehand-discussion in the [Stagehand public Slack](https://join.slack.com/t/stagehand-dev/shared_invite/zt-2wknhxh47-VAed2iE8M8QSQTPpq2VK2Q) (I see you're part of this already but linking here for posterity)

andrewgcodes:
Hey Anirudh, thanks for the quick review and kind words! I'll ping you on
the Slack

lib/llm/LLMProvider.ts
```diff
@@ -53,6 +53,7 @@ export class LLMProvider {
   getClient(
     modelName: AvailableModel,
     clientOptions?: ClientOptions,
+    stagehand?: any,
```
sameelarif, line 56: Would definitely try to avoid using `any` here

filip-michalsky:
this is amazing! looking for the discussion in Slack. Also I see some Devin in here @andrewgcodes  :D

cmaycumber:
Is there any reason this stalled out? Can I help at all?

seanmcguire12:
token usage can be accessed via `stagehand.metrics`

cyberpwnn:
> token usage can be accessed via `stagehand.metrics`

No it cant, it doesnt work, its not documented, and it always shows 0. Please provide a way to access metrics.

seanmcguire12:
@cyberpwnn either set `useAPI: false` or `env: "LOCAL"` in your stagehand constructor and you should see them if you run something like: 
```typescript
const metrics = stagehand.metrics;
console.log(JSON.stringify(metrics, null, 2));
```

PR #340, opened 2024-12-28
support iframes + better debugging dev ex in dom processing

navidkpr:
# why
We want to be able to extract elements from iframes and use them for extraction / taking action.

This PR unlocks iframe support overall, but iframe access is still blocked on browser-level (by iframe same origin policy) for most websites. We will need to override this in our own browser to unlock the full power of this PR.

# what changed
When processing doms, we also process elements from iframes. For visibility checks, we check if the parent iframe is visible and if the element is visible inside the iframe. For element in view, we now calculate the element height based on the iframe window location.

# test plan
Tested locally by running on websites with extensions that added iframes. You can test this by installing https://system32.sh on your browser and checking if the dom processor extract elements from the extension on top of the page.


package.json
```diff
@@ -17,6 +17,7 @@
     "evals": "npm run build-dom-scripts && tsx evals/index.eval.ts",
     "e2e": "npm run build-dom-scripts && cd evals/deterministic && npx playwright test",
     "build-dom-scripts": "tsx lib/dom/genDomScripts.ts",
+    "bundle-dom-scripts": "esbuild ./lib/dom/index.ts --bundle --outfile=./lib/dom/bundle.js --platform=browser --target=es2015 --minify",
```
kamath: when is this supposed to be run? should it be part of `npm run build`? can we use `tsup` here for consistency?

lib/dom/xpathUtils.ts
```diff
@@ -130,7 +236,6 @@ async function generateComplexXPath(element: ChildNode): Promise<string> {
       const el = currentElement as Element;
       let selector = el.tagName.toLowerCase();
 
-      // List of attributes to consider for uniqueness
```
kamath: why is this removed?

seanmcguire12 (approved):

kamath (approved):

lib/index.ts
```diff
@@ -147,6 +148,11 @@ async function getBrowser(
       const session = await browserbase.sessions.create({
         projectId,
         ...browserbaseSessionCreateParams,
+        ...(shouldUseUnsafeMode
+          ? {
+              unsafeMode: true,
```
efecan-fragile: Needs to be nested within the `browserSettings` object as per slack convo! Will likely need an SDK update to include the types.

neoromantic:
Just curious, would it support nested iframes? I have such case in my work with stagehand/browserbase

gorlitzer:
Hello there, any update on this?

Szetty:
@navidkpr any updated ?

Szetty:
@navidkpr @pkiv @kamath @seanmcguire12 any idea when this will be finished ?

Szetty:
@navidkpr @pkiv @kamath @seanmcguire12 any updates ?

Szetty:
@seanmcguire12 any reason why this was closed ? maybe it was resolved differently ?

seanmcguire12:
@Szetty this was closed because it wasn't the approach we wanted to take for solving the iframe problem. As noted in the PR desc, it'll still be blocked by the same origin policy, so it didn't make much sense to merge before having this part solved. 

We're currently working on a different approach that uses CDP instead of injected/browser side JavaScript so that we can get around this. 

The current workaround for this is to use `stagehand.agent` with a computer-use model. Hoping we'll be able to deliver a solution that works with the classic stagehand primitives (`act`, `extract`, `observe`) very soon though, so stay tuned!

PR #347, opened 2024-12-29
prompt: update instructions for browser automation

feder-cr:
- enhance the system prompt with more detailed guidelines for Playwright usage
- add sections on handling popups and missing fields, and tool usage specifics
- improve clarity on variable handling and focus on executing targeted actions

# why

many times models like the gpt-4o-mini fail to generate skipSection action, with this prrompt this almost never happens.

### DIFF
--- a/lib/prompt.ts
+++ b/lib/prompt.ts
@@ -4,25 +4,45 @@ import { ChatMessage } from "./llm/LLMClient";
 // act
 const actSystemPrompt = `
 # Instructions
-You are a browser automation assistant. Your job is to accomplish the user's goal across multiple model calls by running playwright commands.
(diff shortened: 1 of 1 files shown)


kamath:
Hey @feder-cr, thanks for doing this! Do you have any examples of where this prompt performed better than the existing one? Would love to add to our eval set :)

feder-cr:
@kamath Tested with DeepSeek and GPT-Mini for application across various job search middleware, but this happens because, also for other kind of sites, less powerful models with the current prompt struggle to assess which sections to skip without having to copy anything.

kamath:
Thanks for taking the time to do this!! I think it should be addressed in #373 though; lmk if this doesn't address your needs :) happy to re-open if so

Issue #183, opened 2024-11-19
Expand LLMs to Ollama and beyond


kamath:
Start with Ollama and allow users to bring their own LLMs. Requires custom tool use integration

PR #349, opened 2024-12-29
ollama client example

arihanv:
# why
Support open-source LLMs and VLMs - #183 

# what changed
- created client for Ollama
- currently supports llama-3.2 but other LLMs can be added to the list of available models
- created example navigating hackernews and extracting posts with llama 3.2

# TODO
- Support vision models (llama-3.2-vision, moondream, etc.)
- Add other models to the available models list


kamath:
covered in #352

Issue #350, opened 2024-12-30
Parameterize `LLMClient` for people to use whatever LLM they want


kamath:
We only run evals on 4o, o1, and Sonnet rn, and we don't want to unnecessarily complicate evals. However, we don't want to stop users from experimenting with different LLMs. Thus, we should expose `LLMClient` and allow people to pass in an `LLMClient` of their choosing if they want to use providers like Ollama, Groq, Gemini, etc. Relates to #350, #183, #184, #252, #307

kamath:
done with #352

PR #352, opened 2024-12-30
allow llmClient to be optionally passed in

arihanv:
# why
Allow users to easily define their own llm clients rather than adding support for each major provider in the main codebase #350 

# what changed
Added `llmClient` to `Stagehand` constructor so that custom clients can be passed in

### DIFF
--- a/examples/external_client.ts
+++ b/examples/external_client.ts
@@ -1,0 +1,46 @@
+import { type ConstructorParams, type LogLine, Stagehand } from "../lib";
+import { z } from "zod";
+import { OllamaClient } from "./external_clients/ollama";
+
+const StagehandConfig: ConstructorParams = {
+  env: "BROWSERBASE",
+  apiKey: process.env.BROWSERBASE_API_KEY,
+  projectId: process.env.BROWSERBASE_PROJECT_ID,
+  verbose: 1,
+  llmClient: new OllamaClient(
+    (message: LogLine) =>
+      console.log(`[stagehand::${message.category}] ${message.message}`),
+    false,
+    undefined,
+    "llama3.2",
+  ),
+  debugDom: true,
+};
+
+async function example() {
+  const stagehand = new Stagehand(StagehandConfig);
+
+  await stagehand.init();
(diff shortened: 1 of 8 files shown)


kamath:
this looks great! can we add an example of how to use a custom client (like in #349 with ollama/llama 3.2) in `examples/`? would love `custom_client.ts` that implements an `LLMClient` for Llama 3.2 using Ollama that can be passed in on `init`. 

we also need to work towards a fast-follow PR that can allow a user to pass in a custom `LLMClient` to act/extract/observe as well

arihanv:
#349 now proposes an example ollama client which extends the `LLMClient` and uses Llama 3.2.

kamath:
Perfect, I think both look good -- can we add #349 here as follows:
- Move the Ollama Client to `./examples/external_clients/ollama.ts`
- Create an example at `./examples/external_client.ts` that references the LLMClient above

This way, we can more easily clone the branch and make sure nothing else is breaking. tysm 🙇 🙏

kamath:
Also, please add a changeset (`npx changeset`) with a `minor` tag (since this changes the interface to the end user), and add any supplementary docs to the README

examples/external_client.ts
```diff
@@ -0,0 +1,46 @@
...
+import { z } from "zod";
+import { OllamaClient } from "./external_clients/ollama";
+
+const StagehandConfig: ConstructorParams = {
+  env: "BROWSERBASE",
+  apiKey: process.env.BROWSERBASE_API_KEY,
+  projectId: process.env.BROWSERBASE_PROJECT_ID,
+  verbose: 1,
+  llmClient: new OllamaClient(
+    (message: LogLine) =>
```
kamath, line 11: for future PR: we should make `LLMClient` default inherit logger from Stagehand so we don't have to do this

kamath (approved):
hell yeah! merging into an evals branch to make sure we're not missing anything

examples/external_client.ts
kamath: https://github.com/browserbase/stagehand/pull/352

PR #354, opened 2024-12-31
use string similarity on extract_regulations eval

seanmcguire12:
# why
- target website has a typo, LLM corrects the typo when printing extracted data
# what changed
- use string similarity to allow for small differences. we pass with/without the typo


kamath:
don't think this is necessary; would rather make `extract` promptable to avoid spelling errors

PR #355, opened 2025-01-01
fix(logger): enable disabling logger

charlypoly:
# why

Setting `verbose: 0` does not put the logger in quiet mode

# what changed

# test plan

```bash
let a = 0
let c = null

a || 1 // -> 1
c || 1 // -> 1

a ?? 1 // -> 0
c ?? 1 // -> 0

```


kamath (approved):
Happy new year! Thanks Charly :)

PR #362, opened 2025-01-03
Fix text extract collisions

seanmcguire12:
# note
- draft PR until i fix/add a couple evals
# why
- collisions when placing characters on the canvas result in jumbled/missing data
# what changed
- calculate the maximum width of the canvas upfront (before placing characters on the canvas)
- use heuristic value instead of attempting to determine the avg width of characters on the webpage. having excess space is better than not enough space (results in collisions)
# test plan
- evals 

# example
## here is an example of `formattedText` before fix:

Screenshot 2025-01-10 at 12 16 10 PM

## `formattedText` after fix:

Screenshot 2025-01-10 at 12 20 13 PM


kamath (approved):

kamath:
ah wait make sure to add a changeset before merging

PR #365, opened 2025-01-03
filter eval triggers

sameelarif:
# why
Evals take a long time. We should only run evals on the categories that are affected.

# what changed
The workflow for evals now reads the labels of the PR and only runs the eval categories that are tagged.

# test plan
Test workflows

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - main
+    paths:
+      - "lib/**"
+      - "types/**"
   pull_request:
     types:
       - opened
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -4,10 +4,14 @@ on:
   push:
     branches:
       - main
+    paths:
+      - "lib/**"
+      - "types/**"
```
kamath: actually i think this should be `lib` or `evals`. changes to types should be caught in a build with either tsc or eslint

sameelarif: agreed, changed to that

kamath (approved):
finally. lfg 🚀

PR #366, opened 2025-01-03
update: LLMClient default inherits logger from Stagehand

arihanv:
# why
Simplifies the usage of `LLMClient` by eliminating the need for the logger to be defined. Instead, logger is inherited by default from `Stagehand`.

Before:
```
llmClient: new OllamaClient(
  (message: LogLine) =>
    console.log(`[stagehand::${message.category}] ${message.message}`),
  false,
  undefined,
  "llama3.2",
),
```

After:
```
llmClient: new OllamaClient({
  modelName: "llama3.2",
})
```

Additionally, other data members, namely `enableCaching` and `cache,` take on default values so they also do not have to be passed in by the user upon the creation of a client.

# what changed
The provided and example `LLMClient`s now use named properties rather than positional arguments. This follows a better constructor pattern and enables the user to omit certain properties so that their default values can be used.

### DIFF
--- a/.changeset/mean-swans-fix.md
+++ b/.changeset/mean-swans-fix.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": minor
+---
+
+Logger in LLMClient is inherited by default from Stagehand. Named rather than positional arguments are used in implemented LLMClients.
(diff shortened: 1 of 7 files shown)


kamath:
merging into internal `evals/llmclient-logger` to run evals!

kamath (approved):

PR #371, opened 2025-01-03
add branding to readme and link to docs

kamath:
just a branding update :) 

need to update contributing and roadmap as well separately. also removed a lot of api reference in favor of docs


filip-michalsky (approved):
LGTM!

PR #372, opened 2025-01-03
add o1 to model

navidkpr:
# why
-

# what changed
add o1 to list of supported models

# test plan
tested using example

### DIFF
--- a/evals/deterministic/tests/page/o1.test.ts
+++ b/evals/deterministic/tests/page/o1.test.ts
@@ -1,0 +1,33 @@
+import { test, expect } from "@playwright/test";
+import { Stagehand } from "../../../../lib";
+import StagehandConfig from "../../stagehand.config";
+import { z } from "zod";
+
+test.describe("StagehandPage - content", () => {
+  test("should retrieve the full HTML content of the page", async () => {
+    test.setTimeout(180_000);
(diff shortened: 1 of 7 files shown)


lib/llm/OpenAIClient.ts
```diff
@@ -100,9 +100,19 @@ export class OpenAIClient extends LLMClient {
         });
       }
     }
+
+    if (this.modelName === "o1") {
+      delete options.temperature;
```
kamath, line 105: can we instead do something like:
```
{ temperature, top_p, frequency_penalty, presence_penalty, ...options } = options
```

then we can log each as being removed since model is o1

kamath (commented):
generally looks good, just one comment on code style to avoid `delete`. e2e test is failing for everyone rn, most likely unrelated to this PR. looking into what's going on; will approve/merge once complete

evals/deterministic/tests/page/o1.test.ts
```diff
@@ -0,0 +1,33 @@
+import { test, expect } from "@playwright/test";
```
kamath, line 1: this isn't a `deterministic` test, this directory is for e2e tests that don't leverage AI at all. this would be a better fit going in `evals/tasks`, and adding the `combination` label in `evals.config.json`

PR #373, opened 2025-01-03
allow `systemPrompt` input

sameelarif:
# why
We want to allow the input of guidelines or "system" instructions for Stagehand to follow while carrying out actions.

# what changed
Added an `systemPrompt` parameter to the Stagehand constructor, allowing users to guide the LLM in making decisions.

# test plan
Run existing evals and possibly develop new ones for testing custom instructions.


kamath:
Do we want these to be chat messages? Why not add to system prompts in `buildActSystemPrompt` etc. in `inference.ts` and `prompt.ts`? That way we can provide unique system prompt overrides for act/extract/observe as well

lib/prompt.ts
```diff
@@ -96,10 +96,26 @@ ${domElements}
   };
 }
 
-export function buildActSystemPrompt(): ChatMessage {
+export function buildUserInstructionsString(
+  userProvidedInstructions?: string,
+): string {
+  return `# Custom Instructions Provided by the User
```
kamath:
```suggestion
  if (!userProvidedInstructions) return ''
  return `# Custom Instructions Provided by the User
```

lib/prompt.ts
```diff
@@ -96,10 +96,26 @@ ${domElements}
...
+
+export function buildActSystemPrompt(
+  userProvidedInstructions?: string,
+): ChatMessage {
   return {
     role: "system",
-    content: actSystemPrompt,
+    content:
+      actSystemPrompt +
+      "\n\n" +
```
kamath: add to `buildUseInstructionsString` in the case the user instructions are undefined; otherwise you just have two empty lines for no reason

lib/llm/OpenAIClient.ts
```diff
@@ -194,7 +194,7 @@ export class OpenAIClient extends LLMClient {
 
       options.messages.push(screenshotMessage);
     }
-
+    console.log(options.messages);
```
kamath: avoid `console.log`

lib/prompt.ts
sameelarif: Edited it to use join()
```ts
    content: [
      actSystemPrompt,
      buildUserInstructionsString(userProvidedInstructions),
    ].join("\n\n"),
```

examples/instructions.ts
```diff
@@ -0,0 +1,32 @@
...
+ * This example shows how to use custom instructions with Stagehand.
+ */
+import { Stagehand } from "../lib";
+import StagehandConfig from "./stagehand.config";
+
+async function example() {
+  const stagehand = new Stagehand({
+    ...StagehandConfig,
+    instructions:
+      "if the users says `secret12345`, click on the 'quickstart' tab. additionally, if the user says to type something, write the translate their input into french and type it.",
```
kamath:
```suggestion
      "if the users says `secret12345`, click on the 'quickstart' tab. additionally, if the user says to type something, translate their input into french and type it.",
```

lib/prompt.ts
```diff
@@ -237,10 +259,14 @@ ONLY print the content using the print_extracted_data tool provided.
...
 
+  const userInstructions = buildUserInstructionsString(
+    userProvidedInstructions,
+  );
+
   const content =
     `${baseContent}${contentDetail}\n\n${instructions}\n${toolInstructions}${
       additionalInstructions ? `\n\n${additionalInstructions}` : ""
-    }`.replace(/\s+/g, " ");
+    }\n\n${userInstructions}`.replace(/\s+/g, " ");
```
kamath: can you do `[prompt, userInstructions].join("\n\n")` instead to avoid unnecessary blank lines at the end in case no user instructions

lib/prompt.ts
```diff
@@ -337,12 +363,15 @@ You will be given:
...
+export function buildObserveSystemPrompt(
+  userProvidedInstructions?: string,
+): ChatMessage {
   const content = observeSystemPrompt.replace(/\s+/g, " ");
 
   return {
     role: "system",
-    content,
+    content:
+      content + "\n\n" + buildUserInstructionsString(userProvidedInstructions),
```
kamath: do `.join` here as well

types/stagehand.ts
```diff
@@ -22,6 +22,10 @@ export interface ConstructorParams {
   modelName?: AvailableModel;
   llmClient?: LLMClient;
   modelClientOptions?: ClientOptions;
+  /**
+   * Instructions for stagehand.
+   */
+  instructions?: string;
```
kamath: maybe rename to `systemPrompt` to make it more clear?

kamath (changes requested):

kamath (approved):

Issue #361, opened 2025-01-02
Support popups


kamath:
We need to support AI actions on popups for the following use case:

> Because I need to go through the Continue with Google flow, which creates a popup when you click on it.
> 
> https://nextdoor.com/login/
> 
> If you are signed out, try clicking the Continue with Google button here. There a popup that appears asking you to type in your credentials

This requires the following Playwright code (from [docs](https://playwright.dev/docs/pages#handling-popups)):

```
// Start waiting for popup before clicking. Note no await.
const popupPromise = page.waitForEvent('popup');
await page.getByText('open the popup').click();
const popup = await popupPromise;
// Interact with the new popup normally.
await popup.getByRole('button').click();
console.log(await popup.title());
```

We want `popup` to have the same act/extract/observe capabilities as a Stagehand `Page`

PR #374, opened 2025-01-04
support popups

sameelarif:
# why
[Addresses #361]
Stagehand does not handle new pages when they are triggered from `page.click` or other Playwright methods.
# what changed
Added a global listener for new pages, and we now automatically switch to new pages when created.
```ts
// `newPage` is overridden to be a StagehandPage object
page.on("popup", async (newPage) => {
  // This will close the popup and navigate to it on the existing page
  await Promise.all([page.goto(newPage.url()), newPage.close()]);
});

await page.goto("https://nextdoor.com/login/");
// This action will create a new tab, and now Stagehand will automatically switch to it upon creation
await page.click("#google-button > div");
await page.waitForEvent("domcontentloaded");
await page.act({
  action: "type 'test@gmail.com' into the email field",
});
```
Example usage:
# test plan
Existing evals

### DIFF
--- a/evals/deterministic/tests/page/on.test.ts
+++ b/evals/deterministic/tests/page/on.test.ts
@@ -1,0 +1,37 @@
+import { test } from "@playwright/test";
+import { Stagehand } from "../../../../lib";
+import StagehandConfig from "../../stagehand.config";
+
+test.describe("StagehandPage - page.on()", () => {
+  test("should navigate to the popup page and close it", async () => {
+    const stagehand = new Stagehand(StagehandConfig);
+    await stagehand.init();
+
+    const stagehandPage = stagehand.page;
+    await stagehandPage.goto(
+      "https://docs.browserbase.com/integrations/crew-ai/introduction",
+    );
+
+    let clickPromise: Promise<void>;
+
+    stagehandPage.on("popup", async (newPage) => {
+      clickPromise = newPage.click(
+        "body > div.page-wrapper > div.navbar-2.w-nav > div.padding-global.top-bot > div > div.navigation-left > nav > a:nth-child(7)",
+      );
+    });
+
+    await stagehandPage.goto(
+      "https://docs.browserbase.com/integrations/crew-ai/introduction",
+    );
+
+    await stagehandPage.click(
+      "#content-area > div.relative.mt-8.prose.prose-gray.dark\\:prose-invert > p:nth-child(2) > a",
+    );
+
+    await clickPromise;
+
+    await stagehandPage.waitForTimeout(5000);
+
+    await stagehand.close();
+  });
+});
--- a/evals/deterministic/tests/page/on.test.ts
+++ b/evals/deterministic/tests/page/on.test.ts
@@ -1,36 +1,133 @@
-import { test } from "@playwright/test";
+import { expect, test } from "@playwright/test";
 import { Stagehand } from "../../../../lib";
 import StagehandConfig from "../../stagehand.config";
 
 test.describe("StagehandPage - page.on()", () => {
-  test("should navigate to the popup page and close it", async () => {
+  test("should click on the crewAI blog tab", async () => {
     const stagehand = new Stagehand(StagehandConfig);
     await stagehand.init();
 
-    const stagehandPage = stagehand.page;
-    await stagehandPage.goto(
+    const page = stagehand.page;
+    await page.goto(
       "https://docs.browserbase.com/integrations/crew-ai/introduction",
     );
 
     let clickPromise: Promise<void>;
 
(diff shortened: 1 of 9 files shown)


lib/StagehandPage.ts
```diff
@@ -108,6 +114,26 @@ export class StagehandPage {
         return target[prop as keyof PlaywrightPage];
       },
     });
+
+    this.intContext.context.on("page", async (newPage) => {
```
kamath: unsure if we should do this automatically. instead, we should allow the user to poll for new pages, i.e. modify `StagehandContext` such that you return a `StagehandPage` when you poll for pages in the context

evals/deterministic/tests/page/on.test.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    const page = stagehand.page;
+    await page.goto(
+      "https://docs.browserbase.com/integrations/crew-ai/introduction",
+    );
+
+    let navigatePromise: Promise<unknown>;
+
+    page.on("popup", async (newPage) => {
+      navigatePromise = Promise.allSettled([
```
kamath, line 48: nice! i didn't know this was a thing

lib/StagehandPage.ts
```diff
@@ -34,13 +34,24 @@ export class StagehandPage {
   ) {
     this.intPage = Object.assign(page, {
       act: () => {
-        throw new Error("act() is not implemented on the base page object");
+        throw new Error(
+          "act() is not implemented on the base page object. Ensure you are calling init() on the Stagehand object before calling methods on the page object.",
```
kamath:
```suggestion
          "You seem to be calling `act` on a page in an uninitialized `Stagehand` object. Ensure you are running `await stagehand.init()` on the Stagehand object before referencing the `page` object.",
```

lib/StagehandPage.ts
```diff
@@ -34,13 +34,24 @@ export class StagehandPage {
...
+        );
       },
       extract: () => {
-        throw new Error("extract() is not implemented on the base page object");
+        throw new Error(
+          "extract() is not implemented on the base page object. Ensure you are calling init() on the Stagehand object before calling methods on the page object.",
```
kamath:
```suggestion
          "You seem to be calling `extract` on a page in an uninitialized `Stagehand` object. Ensure you are running `await stagehand.init()` on the Stagehand object before referencing the `page` object.",
```

lib/StagehandPage.ts
```diff
@@ -34,13 +34,24 @@ export class StagehandPage {
...
+        );
       },
       observe: () => {
-        throw new Error("observe() is not implemented on the base page object");
+        throw new Error(
+          "observe() is not implemented on the base page object. Ensure you are calling init() on the Stagehand object before calling methods on the page object.",
```
kamath:
```suggestion
          "You seem to be calling `observe` on a page in an uninitialized `Stagehand` object. Ensure you are running `await stagehand.init()` on the Stagehand object before referencing the `page` object.",
```

lib/StagehandPage.ts
```diff
@@ -34,13 +34,24 @@ export class StagehandPage {
...
+        );
+      },
+      on: () => {
+        throw new Error(
+          "on() is not implemented on the base page object. Stagehand overrides Playwright's on() method to update it's behavior, so ensure you call init() on the Stagehand object before you call on() on the page object.",
```
kamath:
```suggestion
          "You seem to be referencing a page in an uninitialized `Stagehand` object. Ensure you are running `await stagehand.init()` on the Stagehand object before referencing the `page` object.",
```

don't wanna highlight the stagehand overrides too much to confuse the end user. let's say something simpler

types/page.ts
```diff
@@ -9,13 +12,17 @@ import type {
...
+export interface Page extends Omit<PlaywrightPage, "on"> {
   act: (options: ActOptions) => Promise<ActResult>;
   extract: <T extends z.AnyZodObject>(
     options: ExtractOptions<T>,
   ) => Promise<ExtractResult<T>>;
   observe: (options?: ObserveOptions) => Promise<ObserveResult[]>;
+
+  on: {
+    (event: "popup", listener: (page: Page) => unknown): Page;
+  } & PlaywrightPage["on"];
```
kamath, line 25: Don't we want to put `PlaywrightPage` first to ensure type hierarchy? could be wrong

sameelarif: Putting `PlaywrightPage` first breaks the other types; not sure why. Should be fine as-is though.

kamath: as long as it's type-safe; ig it otherwise would've thrown an error in the evals or example when you tried to run `observe`

kamath (approved):

Issue #368, opened 2025-01-03
Remove dependency on LLM API Key


kamath:
Stagehand theoretically _can_ run without AI (see e2e tests), so we shouldn't throw an error if an `OPENAI_API_KEY` or `ANTHROPIC_API_KEY` is not present, especially as we add new `LLMClient` implementations.

We should make model API keys optional, and throw an error on `createChatCompletion` (and wherever else an external model is actually used) if there's an invalid model API key.

PR #379, opened 2025-01-04
dont require LLMClient to use stagehand

seanmcguire12:
related: #368 
# why
- we want users to be able to use non-ai features of stagehand without requiring them to set up an LLM client
# what changed
- only use default LLM client if they have an API key, otherwise dont pass in an LLM client & only error if they try to use AI features
# test plan
- e2e tests, evals


kamath:
don't think existing evals are going to be enough here; can we add a simple eval in `combination` that tries to use stagehand without any LLM key?

kamath (approved):
good stuff. i like that we now remove LLM API keys from e2e altogether to test this :)

PR #383, opened 2025-01-07
Refactor LLM Types

sameelarif:
# why
We want to reduce dependency on OpenAI types as we continue to add more models

# what changed
Added unifying types for LLM inputs and outputs

# test plan
Existing evals

### DIFF
--- a/lib/llm/AnthropicClient.ts
+++ b/lib/llm/AnthropicClient.ts
@@ -7,13 +7,9 @@ import {
 } from "@anthropic-ai/sdk/resources";
 import { zodToJsonSchema } from "zod-to-json-schema";
 import { LogLine } from "../../types/log";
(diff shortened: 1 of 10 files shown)


lib/llm/AnthropicClient.ts
```diff
@@ -39,7 +35,7 @@ export class AnthropicClient extends LLMClient {
     this.clientOptions = clientOptions;
   }
 
-  async createChatCompletion<T = AnthropicTransformedResponse>(
+  async createChatCompletion<T = LLMResponse>(
```
kamath, line 38: finally 🙏🙏🙏🙏🙏

kamath (approved):
lfg lfg lfg lfg 🧹 great cleanup

PR #384, opened 2025-01-07
leave comments for eval performance

sameelarif:
# why
We want an easy way to view eval category performance

# what changed
Added comments from the CI, which log the performance of each eval category

# test plan
Run existing evals and observe if comments are left by CI


kamath (changes requested):
p sure this leaves a bunch of comments; this will get annoying super quick. we want one comment that can edit itself

kamath:
closing, don't need rn

PR #396, opened 2025-01-09
fix: avoid unnecessary 'miss cache' on OpenAiClient

arlenvasconcelos:
# why
Fix: cache issue
# what changed
Remove cached value of ZodObject before cache key is generated
# test plan


arlenvasconcelos:
Sorry sorry sorry. 
Sorry again.
It was a mistake. I am testing in my forked repository. 

And I would to create a PR for in my cloned repository

PR #397, opened 2025-01-09
rm/replace flakey evals

seanmcguire12:
# why
- target website in`extract_regulations.ts` is unstable
- target website in `ibm.ts` changes too frequently & therefore the eval consistently fails
# what changed
- replaced `extract_regulations.ts` with `extract_csa.ts` which uses a more stable target website and accomplishes a similar goal
- deleted IBM eval


evals/tasks/extract_csa.ts
```diff
@@ -0,0 +1,150 @@
...
+  useTextExtract,
+}) => {
+  const { stagehand, initResponse } = await initStagehand({
+    modelName,
+    logger,
+  });
+
+  const { debugUrl, sessionUrl } = initResponse;
+
+  await stagehand.page.goto(
```
kamath: nit: let's establish the pattern of just doing `const { page, context } = stagehand` and using that in evals as well.

kamath (approved):

PR #405, opened 2025-01-14
Scroll on 'main scrollable element' when possible

seanmcguire12:
# why
- currently, we only scroll on the root DOM element. thus, we ignore elements that are inside a scrollable container
- this is problematic, because there may be other elements that we need to scroll into visibility so that they are included in the list of candidate elements
# examples
## Zillow:
Screenshot 2025-01-14 at 3 02 33 PM

## apartments.com
Screenshot 2025-01-14 at 3 03 46 PM

# what changed
- updated `processAllOfDom` to attempt scrolling on a large scrollable container, and fall back to the original approach (scrolling on the root DOM) if there are no large scrollable containers

# important note
- this PR should only impact `processAllOfDom`
- this means that we will still need a future PR to extend this logic to `processDom` to enable `act` to take actions on elements that are positioned deep within a scrollable container
- the thought process behind this decision was to control the scope of this PR

# test plan
- added evals for Zillow & apartments.com

### DIFF
--- a/evals/tasks/extract_zillow.ts
+++ b/evals/tasks/extract_zillow.ts
@@ -13,14 +13,12 @@ export const extract_zillow: EvalFunction = async ({
     domSettleTimeoutMs: 3000,
     configOverrides: {
       debugDom: false,
-    }
+    },
   });
 
   const { debugUrl, sessionUrl } = initResponse;
 
-  await stagehand.page.goto(
-    "https://zillow-eval.surge.sh/"
-  );
+  await stagehand.page.goto("https://zillow-eval.surge.sh/");
   // timeout for 5 seconds
   await stagehand.page.waitForTimeout(5000);
   const real_estate_listings = await stagehand.page.extract({
--- a/evals/tasks/extract_zillow.ts
+++ b/evals/tasks/extract_zillow.ts
@@ -1,0 +1,70 @@
+import { z } from "zod";
+import { initStagehand } from "../initStagehand";
+import { EvalFunction } from "../../types/evals";
+
+export const extract_zillow: EvalFunction = async ({
+  modelName,
+  logger,
+  useTextExtract,
+}) => {
+  const { stagehand, initResponse } = await initStagehand({
+    modelName,
+    logger,
+    domSettleTimeoutMs: 3000,
+  });
+
+  const { debugUrl, sessionUrl } = initResponse;
+
+  await stagehand.page.goto(
+    "https://www.zillow.com/homes/San-Francisco,-CA_rb/",
+  );
+  const real_estate_listings = await stagehand.page.extract({
+    instruction:
+      "Extract all the real estate listings with their prices and their addresses.",
+    schema: z.object({
+      listings: z.array(
+        z.object({
+          price: z.string().describe("The price of the listing"),
+          trails: z.string().describe("The address of the listing"),
+        }),
+      ),
+    }),
+    modelName,
+    useTextExtract,
+  });
+
+  await stagehand.close();
+  const listings = real_estate_listings.listings;
+  const expectedLength = 38;
+
+  if (listings.length < expectedLength) {
+    logger.error({
+      message: "Incorrect number of listings extracted",
+      level: 0,
+      auxiliary: {
+        expected: {
+          value: expectedLength.toString(),
+          type: "integer",
+        },
+        actual: {
+          value: listings.length.toString(),
+          type: "integer",
+        },
+      },
+    });
+    return {
+      _success: false,
+      error: "Incorrect number of listings extracted",
+      logs: logger.getLogs(),
+      debugUrl,
+      sessionUrl,
+    };
+  }
+
+  return {
+    _success: true,
+    logs: logger.getLogs(),
+    debugUrl,
+    sessionUrl,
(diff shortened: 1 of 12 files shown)


evals/tasks/extract_zillow.ts
```diff
@@ -0,0 +1,70 @@
...
+        }),
+      ),
+    }),
+    modelName,
+    useTextExtract,
+  });
+
+  await stagehand.close();
+  const listings = real_estate_listings.listings;
+  const expectedLength = 38;
```
kamath, line 41: are we sure this is fixed? this doesn't seem sustainable

lib/dom/global.d.ts
```diff
@@ -19,7 +19,10 @@ declare global {
     }>;
     debugDom: () => Promise<void>;
     cleanupDebug: () => void;
-    scrollToHeight: (height: number) => Promise<void>;
+    scrollToHeight: (
```
kamath: nit: can we make this serialized json

lib/dom/process.ts
```diff
@@ -9,6 +9,85 @@ export function isTextNode(node: Node): node is Text {
   return node.nodeType === Node.TEXT_NODE && Boolean(node.textContent?.trim());
 }
 
+function getMainScrollableElement(): HTMLElement {
+  // Default to <html> (document.documentElement)
+  let mainScrollable: HTMLElement = document.documentElement;
+  let maxScrollableArea = 0;
+
+  // Find the largest candidate
```
kamath: so this is basically just getting literally the biggest element and marking that main? using this approach should totally be opt-in; i feel like this can get funky super fast

lib/dom/process.ts
```diff
@@ -9,6 +9,85 @@ export function isTextNode(node: Node): node is Text {
...
+        maxScrollableArea = area;
+        mainScrollable = elem;
+      }
+    }
+  }
+
+  // Verify we can actually scroll that candidate
+  if (mainScrollable !== document.documentElement) {
+    if (!canElementScroll(mainScrollable)) {
+      console.log(
```
kamath, line 47: future note: we should consolidate DOM logs to a `window.log` function

lib/dom/process.ts
```diff
@@ -58,22 +158,46 @@ export async function processAllOfDom() {
   };
 }
 
-export async function scrollToHeight(height: number) {
-  window.scrollTo({ top: height, left: 0, behavior: "smooth" });
+export async function scrollToHeight(
+  containerOrWindow: HTMLElement | Window,
```
kamath: let's name this differently and make it a custom type as well -- something like `mainContainer: StagehandContainer`, where `StagehandContainer` is `HTMLElement | Window`.

that way you can just do something like `mainContainer.scrollTo({ top: height, left: 0, behavior: "smooth" })` without all these if else statements

lib/dom/process.ts
```diff
@@ -58,22 +158,46 @@ export async function processAllOfDom() {
...
 
-  // Wait for scrolling to finish using the scrollend event
   await new Promise<void>((resolve) => {
     let scrollEndTimer: number;
-    const handleScrollEnd = () => {
+    const handleScroll = () => {
       clearTimeout(scrollEndTimer);
       scrollEndTimer = window.setTimeout(() => {
-        window.removeEventListener("scroll", handleScrollEnd);
+        if (containerOrWindow instanceof Window) {
```
kamath: should not be an if statement; this should be one method

lib/dom/process.ts
```diff
@@ -58,22 +158,46 @@ export async function processAllOfDom() {
...
+          containerOrWindow.removeEventListener("scroll", handleScroll);
+        }
         resolve();
       }, 100);
     };
 
-    window.addEventListener("scroll", handleScrollEnd, { passive: true });
-    handleScrollEnd();
+    if (containerOrWindow instanceof Window) {
+      containerOrWindow.addEventListener("scroll", handleScroll, {
```
kamath: should also not be in an if statement

lib/dom/process.ts
```diff
@@ -83,32 +207,45 @@ export async function processElements(
...
-  const viewportHeight = calculateViewportHeight();
-  const chunkHeight = viewportHeight * chunk;
 
-  // Calculate the maximum scrollable offset
-  const maxScrollTop = document.documentElement.scrollHeight - viewportHeight;
+  const container = scrollableContainer ?? document.documentElement;
+  let viewportHeight: number;
+  let totalScrollHeight: number;
+
+  if (container instanceof Window) {
```
kamath: not an if statement

kamath (changes requested):
biggest thing is we probably need one unified abstract class called `StagehandContainer` that implements methods like `getHeight()`, `scrollTo()`, etc.

great work here, but these `if` statements aren't really sustainable as we keep adding to this kind of workflow

evals/tasks/extract_zillow.ts
seanmcguire12: I can beef this up a tad by making sure fields are filled within the object, but hardcoded expected values on zillow is going to be a nightmare to maintain.

kamath (approved):

PR #411, opened 2025-01-17
queue jobs based on labels: dont run concurrently

seanmcguire12:
# why
- dont overload BB
# what changed
- updated CI to queue jobs instead of running selected jobs all at the same time
# example
## we will do this:
Screenshot 2025-01-17 at 1 25 35 PM

## instead of this:
screenshot_2025-01-15_at_6 33 19___pm_480


kamath (approved):
lgtm, in CI it's running act and pausing before running other evals so i think gtg

PR #416, opened 2025-01-20
Revert back to sequential evals runs in CI

seanmcguire12:
# why
- Migrate from concurrency back to sequential evals to avoid overloading BB
# what changed
- we will now run evals in CI sequentially based on the labels that the developer chooses.
# note
- for eval sets that are **NOT** included in the labels chosen by the developer, the job will still _appear_ to have "run", but it wont actually run
- in other words, there will be a green check mark, but we will immediately `exit code 0` from that job


kamath (approved):
👑

PR #417, opened 2025-01-20
simplify stagehand method calls

sameelarif:
# why
Often times it's unnecessary to pass in Stagehand method options as an object, we want to simplify inputs to just a string. This allows syntax such as the following:
```ts
await page.act("search for browserbase");

// the default schema is:
// z.object({
//  extraction: z.string(),
// });
const { extraction } = await page.extract("the first result's title");

await page.observe("return the possible next steps from this page");
```

# what changed
Added new input types in the Stagehand page interface (`types/page`) and updated input handling in the `StagehandPage`class to parse the inputs accordingly.

# test plan
Run evals to ensure existing functionality has not been altered and add new evals that test the new syntax works as expected.

### DIFF
--- a/.changeset/hungry-scissors-mix.md
+++ b/.changeset/hungry-scissors-mix.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": minor
+---
+
+Simplify Stagehand method calls by allowing a simple string input instead of an options object.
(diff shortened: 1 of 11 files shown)


kamath (approved):
excellent work 🚢 :shipit:

miguelg719 (approved):
Just tested it works great on my end. Great turnaround!

sameelarif:
Evals passed, extract evals failed by 2%. merging

PR #418, opened 2025-01-21
Reference dist/ in examples and evals

kamath:
# why
There have been times where we make code changes, but type exports aren't done properly. By referencing `dist/` in examples and evals, we're forcing ourselves to use the publicly-available types unless explicitly evals-related.

This helps us prevent releases that don't properly export the necessary types

# what changed
- import paths now go to dist, which is post-build
- no changeset since this is a stagehand dev ex improvement

# test plan
evals/examples


examples/external_clients/ollama.ts
sameelarif, line 1: should we include these changes in a separate PR? doesn't really matter though

kamath: 🤷‍♂️ , honestly it's just a bunch of tedious small changes, let's just knock em out if it works

sameelarif (approved):

seanmcguire12 (approved):

PR #428, opened 2025-01-23
Remove vision

seanmcguire12:
# why
- vision doesn't work very well (at all)
# what changed
- still allow for useVision as a param, but give a warning if it is set to true
- do not use vision for action verification
# test plan
evals


.changeset/tame-gifts-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+"@browserbasehq/stagehand": major
```
kamath: whoa not a major this is a minor

lib/prompt.ts
```diff
@@ -363,7 +362,7 @@ You will be given:
 2. ${
     isUsingAccessibilityTree
       ? "a hierarchical accessibility tree showing the semantic structure of the page"
-      : "a numbered list of possible elements or an annotated image of the page"
```
kamath, line 366: whoa nice catch

kamath (approved):
🪓

.changeset/tame-gifts-fail.md
seanmcguire12: bro trying to pull a sneaky one?

seanmcguire12: ive never been good at changesets 🥴

PR #448, opened 2025-01-31
Improve reliability of radio button clicks

seanmcguire12:
# why
- clicking a radio button with playwright can be finnicky
- often, it is best to just click the `label` which corresponds to the radio button instead of trying to click the radio button itself
# what changed
- this PR introduces a step in `_performPlaywrightMethod` where we directly check if the target element is a radio button.
- if true, it will run through a series of checks to try and find the corresponding `label` for the radio button
- if a `label` is found, then the `label` is clicked instead of the radio button itself

### DIFF
--- a/.changeset/rich-pans-lay.md
+++ b/.changeset/rich-pans-lay.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+improve handling of radio button clicks
--- a/lib/handlers/actHandler.ts
+++ b/lib/handlers/actHandler.ts
@@ -413,13 +413,6 @@ export class StagehandActHandler {
             },
           },
         });
-        // try {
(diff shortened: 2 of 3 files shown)


kamath:
@seanmcguire12 should we add more evals here? i.e. specifically for radio buttons/checkboxes?

seanmcguire12:
> @seanmcguire12 should we add more evals here? i.e. specifically for radio buttons/checkboxes?

@kamath yeah ideally it is tested somehow, but i wasn't sure it warranted the addition of some evals considering we run 5 trials per model per eval. I spun up some hosted sites that outline the edge cases that this PR fixes, so I could easily throw them into evals, but personally I think unit tests would be best for something like this. Lmk your thoughts

kamath (approved):

PR #452, opened 2025-02-01
Anirudh/add o3

kamath:
# why
everyone else is doing it

# what changed
add o3 to `AvailableModel` + some auxiliary changes to models

# test plan
add o3-mini to evals (much cheaper than 4o so shouldn't add much cost)


miguelg719 (approved):
Looks good to me

PR #453, opened 2025-02-01
Fix: Ensure goto Uses Overridden Page in act/performPlaywrightMethod for Proper DOM Injection

### LINKED ISSUE #443: `goto` should be handled better in act/performPlaywrightMethod
right now, you can say `"go to instagram"` in `act`, but the resulting `page.goto` isn't the overridden page -- resulting in Stagehand DOM scripts not being injected. we should fix this in `actHandler` and `performPlaywrightMethod`

askadityapandey:
## Why
fixes #443
The current implementation of the `"goto"` command within the `act` handler calls `page.goto` on the original page instance. This prevents Stagehand’s DOM scripts from being injected as intended. This PR addresses the issue by explicitly handling `"goto"` commands to use the overridden page instance, ensuring that the DOM modifications occur correctly.

---

## What Changed
- **Dedicated Branch for `"goto"` Command:**  
  Added a specific branch in the `_performPlaywrightMethod` method to handle the `"goto"` command. The URL is extracted from the command arguments and passed directly to `this.stagehandPage.page.goto(url)`, ensuring that the overridden page (with the injected DOM scripts) is used.
  
- **Enhanced Logging and Error Handling:**  
  Introduced additional logging for the `"goto"` command execution to improve debugging and traceability. Error handling has been updated to catch and log exceptions appropriately during navigation.

- **DOM Settling After Navigation:**  
  After navigation, the method waits for the DOM to settle using `_waitForSettledDom`, guaranteeing that the page is fully loaded and the Stagehand scripts have been injected.

---

## Test Plan
1. **Basic Navigation Test:**  
   - Execute a command like `"goto https://example.com"` via Stagehand.
   - Confirm that the page navigates to the specified URL.
   - Verify that the Stagehand DOM scripts are injected correctly.

2. **Error Handling:**  
   - Test with invalid URLs or simulate network errors.
   - Check that the error is logged appropriately and that a `PlaywrightCommandException` is thrown.

3. **Regression Test for Other Methods:**  
   - Run existing tests for other actions (e.g., `"click"`, `"fill"`, `"press"`) to ensure their functionality remains unchanged.

4. **End-to-End Flow:**  
   - Execute a full action sequence that includes navigation along with other actions.
   - Verify that the overridden page instance is used consistently throughout the sequence and that the DOM modifications persist.

### DIFF
--- a/lib/handlers/actHandler.ts
+++ b/lib/handlers/actHandler.ts
@@ -200,36 +200,242 @@ export class StagehandActHandler {
     return actionCompleted;
   }
 
-  private async _performPlaywrightMethod(
-    method: string,
-    args: unknown[],
-    xpath: string,
-    domSettleTimeoutMs?: number,
-  ) {
-    const locator = this.stagehandPage.page.locator(`xpath=${xpath}`).first();
-    const initialUrl = this.stagehandPage.page.url();
+private async _performPlaywrightMethod(
+  method: string,
+  args: unknown[],
+  xpath: string,
+  domSettleTimeoutMs?: number,
+) {
+  // New branch: Handle "goto" commands separately.
+  if (method === "goto") {
+    try {
+      const url = args[0]?.toString();
+      this.logger({
+        category: "action",
+        message: "navigating to URL using goto",
+        level: 1,
+        auxiliary: {
+          url: { value: url, type: "string" },
+        },
+      });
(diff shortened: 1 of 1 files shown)


kamath:
my bad, didn't see this until now! thank you so much for your contribution, will look at this in depth probably later this week or this weekend. sincerely appreciate your hard work and detailed write up

PR #457, opened 2025-02-03
Eval server setup

shinpads:
## What was added
- Express server with `/test` endoint
- Add option to pass in cdp_url

## Why this is necessary
- Express server needed as an interface to run evals on
- cdp_url needed to run tests on our simulated browser environments


kamath:
marking draft until ready

PR #458, opened 2025-02-03
A11y empty xpath fix

miguelg719:
# why

Some useful roles were being cleaned on a11y. Xpath generation from backend node id was failing for text nodes.

# what changed

Updated `getAccessibilityTree()` to make sure it doesn't skip useful nodes. Improved `getXPathByResolvedObjectId()` to account for text nodes and not skip generation

# test plan


seanmcguire12 (approved):
nice one

PR #459, opened 2025-02-04
Hybrid of a11y tree & DOM for input to observe

seanmcguire12:
# why
- sometimes the `role` for an `AXNode` is something useless like `none` or `generic`. this isn't very helpful context for the LLM
# what changed
- if the `role` for an `AXNode` is `none` or `generic`, we replace it with the `tagName` of its corresponding DOM element

### before this PR:
Screenshot 2025-02-03 at 7 11 53 PM

### after this PR:
Screenshot 2025-02-03 at 7 14 09 PM

### DIFF
--- a/lib/a11y/utils.ts
+++ b/lib/a11y/utils.ts
@@ -32,12 +32,77 @@ export function formatSimplifiedTree(
  * 2. Collapses generic/none nodes with single child
  * 3. Keeps generic/none nodes with multiple children but cleans their subtrees
  */
-function cleanStructuralNodes(
+async function cleanStructuralNodes(
   node: AccessibilityNode,
-): AccessibilityNode | null {
+  page?: StagehandPage,
+  logger?: (logLine: LogLine) => void,
+): Promise<AccessibilityNode | null> {
   // Filter out nodes with negative IDs
   if (node.nodeId && parseInt(node.nodeId) < 0) {
     return null;
+  }
+
+  // If we have a backendDOMNodeId and the node is generic/none, try to get its tagName
+  if (
+    page &&
+    logger &&
+    node.backendDOMNodeId !== undefined &&
+    (node.role === "generic" || node.role === "none")
+  ) {
+    try {
+      // 1) Resolve the node to a Runtime object
+      const { object } = await page.sendCDP<{
+        object: { objectId?: string };
+      }>("DOM.resolveNode", {
+        backendNodeId: node.backendDOMNodeId,
+      });
+
+      if (object && object.objectId) {
+        try {
+          // Get the tagName for the node
+          const { result } = await page.sendCDP<{
+            result: { type: string; value?: string };
+          }>("Runtime.callFunctionOn", {
+            objectId: object.objectId,
+            functionDeclaration: `
+              function() {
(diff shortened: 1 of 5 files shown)


lib/a11y/utils.ts
```diff
@@ -137,26 +149,122 @@ export function buildHierarchicalTree(nodes: AccessibilityNode[]): TreeResult {
...
     const sources = nodes.map((node) => ({
       role: node.role?.value,
       name: node.name?.value,
       description: node.description?.value,
       value: node.value?.value,
       nodeId: node.nodeId,
+      backendDOMNodeId: node.backendDOMNodeId,
       parentId: node.parentId,
       childIds: node.childIds,
+      xpath: node.xpath,
```
miguelg719: for large websites it makes sense to compute the xpaths at runtime for the candidates selected by observe

lib/a11y/utils.ts
```diff
@@ -99,6 +107,10 @@ export function buildHierarchicalTree(nodes: AccessibilityNode[]): TreeResult {
       ...(hasValidName && { name: node.name }), // Only include name if it exists and isn't empty
       ...(node.description && { description: node.description }),
       ...(node.value && { value: node.value }),
+      ...(node.backendDOMNodeId !== undefined && {
```
miguelg719, line 188: isnt backendDOMNodeId === nodeId?

lib/a11y/utils.ts
```diff
@@ -137,26 +149,122 @@ export function buildHierarchicalTree(nodes: AccessibilityNode[]): TreeResult {
...
-    // Extract specific sources
+    // For each node with a backendDOMNodeId, resolve it
+    for (const node of nodes) {
+      // Convert the AX role to a plain string (since .value is optional)
+      const role = node.role?.value;
+
+      if (node.backendDOMNodeId !== undefined) {
+        try {
+          // 1) Resolve the node to a Runtime object
+          const { object } = await page.sendCDP<{
```
miguelg719: Suggestion: resolve only those with a generic / none roles

miguelg719 (approved):
This looks good to me, you're cooking here

lib/a11y/utils.ts
seanmcguire12: yeah nice catch, we arent precomputing them, i'll take them out of the AXNode type since they are unused

lib/a11y/utils.ts
seanmcguire12: no, nodeId is a11y specific (remember we used to have negative values for nodeIds)

lib/a11y/utils.ts
```diff
@@ -137,29 +231,45 @@ export function buildHierarchicalTree(nodes: AccessibilityNode[]): TreeResult {
...
+      logger,
+    );
+
+    logger({
+      category: "observation",
+      message: `got accessibility tree in ${Date.now() - startTime}ms`,
+      level: 1,
+    });
+
+    // fs.writeFileSync("../hybrid_tree.txt", hierarchicalTree.simplified);
```
kamath:
```suggestion

```

kamath (approved):

PR #463, opened 2025-02-04
Annotations for Scrollable elements in a11y-dom hybrid

seanmcguire12:
# why
- we want the LLM to be able to find scrollable containers in `observe` calls
- we don't have this level of granularity with the a11y tree alone
# what changed
- changed the `getMainScrollableElement` function in `process.ts` to `getScrollableElements`, which now optionally returns a list of _all_ the scrollable elements on the page unless the `topN` param is specified. 
- For usage in `processAllOfDom`, we set `topN` to `1` to maintain the existing behaviour.
- added a new function in `process.ts` called `getScrollableElementXpaths` which calls `getScrollableElements` and then calls `generateXpaths` on each of the returned elements. This function returns a list of xpaths for each scrollable element on the page.
- exposed the new `getScrollableElementXpaths` globally so that it can be called from `a11y/utils.ts`
- created a new function `findScrollableElementIds` in `a11y/utils.ts` which calls the `getScrollableElementXpaths` function from the browser context, evaluates the returned xpaths, and returns their corresponding `backendDOMId`'s
- updated `getAccessibilityTree` to prepend the string `Scrollable` to the `role` of all the scrollable elements in the a11y tree
# example
## LLM input before this PR:
Screenshot 2025-02-04 at 2 47 22 PM

## LLM input after this PR:
Screenshot 2025-02-04 at 2 48 04 PM

### DIFF
--- a/.changeset/chilled-apes-sneeze.md
+++ b/.changeset/chilled-apes-sneeze.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+create a11y + dom hybrid input for observe
--- a/lib/a11y/utils.ts
+++ b/lib/a11y/utils.ts
@@ -159,6 +159,9 @@ export async function getAccessibilityTree(
   await page.enableCDP("Accessibility");
 
   try {
+    // Identify which elements are scrollable and get their backendNodeIds
+    const scrollableBackendIds = await findScrollableElementIds(page);
+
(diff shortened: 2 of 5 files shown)


kamath:
behavior* 🇺🇸

seanmcguire12:
> behavior* 🇺🇸

delete

kamath (approved):

PR #476, opened 2025-02-09
`drawOverlay` observe option

sameelarif:
# why
We want to give users access to utils that highlight `observe()` results in the browser view.

# what changed
Added `drawOverlay: boolean` parameter to `observe()` and called util to clear drawings after the observe call.


miguelg719 (approved):
Looks good to me, getting an error with window.getScrollableElementXpaths() though from a prev PR? Are you getting this error too?

kamath (approved):

PR #479, opened 2025-02-11
Static evals environments

shinpads:
# why
- to have ground truth answers for evals

# what changed
- connect plato to the browser session which will serve the simulations
- had to return `connectUrl` from `Stagehand.init` to be able to pass it to plato


shinpads:
@kamath could you create a branch I could change this PR to merge to?

PR #480, opened 2025-02-11
added fallback try to actFromObserveResult

miguelg719:
# why
Instead of having to do try catches when act from observe result fails, we automatically try again using the method+description on regular act

# what changed
Added logic to actFromObserveResult to retry with regular act and indicate the retry to the user with a log

# test plan
Tested locally

### DIFF
--- a/lib/handlers/actHandler.ts
+++ b/lib/handlers/actHandler.ts
@@ -29,6 +29,7 @@ export class StagehandActHandler {
     [key: string]: { result: string; action: string };
   };
   private readonly userProvidedInstructions?: string;
+  private readonly selfHeal: boolean;
 
   constructor({
     verbose,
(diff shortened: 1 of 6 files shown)


lib/handlers/actHandler.ts
```diff
@@ -92,19 +92,40 @@ export class StagehandActHandler {
...
           trace: { value: err.stack, type: "string" },
           observeResult: { value: JSON.stringify(observe), type: "object" },
         },
       });
-      return {
-        success: false,
-        message: `Failed to perform act: ${err.message}`,
-        action: observe.description || `ObserveResult action (${method})`,
-      };
+      try {
```
kamath, line 125: let's put this behind an opt-out flag. we are telling people that `act` from `observe` has no LLM inference, so we should adhere to that. i think it's ok to default `true` though (i.e. default this behavior)

seanmcguire12 (approved):

kamath (approved):
this would be a minor in the changeset btw

PR #482, opened 2025-02-11
making the instruction more descriptive in observe_taxes

miguelg719:
# why
After pushing the dom+a11y changes evals for observe_taxes were becoming more inconsistent. Basically selecting the right thing but not the actual input element. Just added more description to account for this

# what changed
Changed prompt for eval observe_taxes

# test plan


seanmcguire12 (approved):
Approved ~~pending observe evals pass~~

PR #446, opened 2025-01-30
Feat/add langchain as llmclient

ishaksebsib:
# why

Adding support for Langchain as an LLM client. 

# what changed

- Added `LangchainClient`, extending `LLMClient`, to enable Langchain model support. 
- Created an example file demonstrating usage. 

# test plan

- Verified `LangchainClient` initializes and processes messages correctly. 
- Ran example file to confirm expected behavior.

PR #484, opened 2025-02-12
Ishaksebsib/add langchain external client

filip-michalsky:
from @ishaksebsib :

# why

Adding support for Langchain as an LLM client. 

# what changed

- Added `LangchainClient`, extending `LLMClient`, to enable Langchain model support. 
- Created an example file demonstrating usage. 

# test plan

- Verified `LangchainClient` initializes and processes messages correctly. 
- Ran example file to confirm expected behavior.


types/langchain-stubs.d.ts
sameelarif, line 1: can we keep this contained within the `examples` folder? we don't want non-stagehand types contained with other stagehand types in the `types` folder

package.json
```diff
@@ -77,6 +78,8 @@
   "dependencies": {
     "@anthropic-ai/sdk": "^0.27.3",
     "@browserbasehq/sdk": "^2.0.0",
+    "@langchain/core": "^0.3.40",
+    "@langchain/openai": "^0.4.4",
```
sameelarif: these should be in `devDependencies` since they're not directly stagehand dependencies

sameelarif (changes requested):

examples/langchain.ts
```diff
@@ -0,0 +1,58 @@
...
+import { LangchainClient } from "./external_clients/langchain";
+import { z } from "zod";
+
+async function example() {
+  const stagehand = new Stagehand({
+    env: "LOCAL",
+    debugDom: true,
+    enableCaching: false,
+    llmClient: new LangchainClient({
+      modelName: StagehandConfig.modelName,
```
kamath: Let's use strings here, StagehandConfig model name is only for 4o/sonnet

examples/langchain.ts
```diff
@@ -0,0 +1,58 @@
...
+      apiKey: StagehandConfig.modelClientOptions.apiKey,
```
kamath: same as above

examples/types/langchain-stubs.d.ts
```diff
@@ -0,0 +1,44 @@
+// This file provides stub types for the LangChain modules
+// so that our example can compile without having them installed.
+
+declare module "@langchain/openai" {
```
kamath: Do we need this? Examples are not part of compilation, and tsx doesn't require entire project compilation either

If absolutely necessary, this could be in external_clients/langchain.ts

package.json
```diff
@@ -9,6 +9,7 @@
     "2048": "npm run build-dom-scripts && tsx examples/2048.ts",
     "popup": "npm run build-dom-scripts && tsx examples/popup.ts",
     "example": "npm run build-dom-scripts && tsx examples/example.ts",
+    "langchain": "npm install @langchain/core @langchain/openai && npm run build-dom-scripts && tsx examples/langchain.ts",
```
kamath: install scripts shouldn't be here if they're in dev dependencies

examples/types/langchain-stubs.d.ts
alexdphan: @filip-michalsky

filip-michalsky: @kamath @alexdphan this was done since we did not include the langchain dependencies in package.json.

If we do include those dependencies in package.json now (which I thought we did not want), I think this file can be deleted as the linter should not complain anymore.

examples/external_clients/langchain.ts
```diff
@@ -0,0 +1,218 @@
...
+
+export class LangchainClient extends LLMClient {
+  public type = "langchain" as const;
+  private model: ChatOpenAI;
+
+  constructor({ modelName, apiKey }: { modelName: string; apiKey: string }) {
+    super(modelName as AvailableModel);
+    this.model = new ChatOpenAI({
+      modelName: modelName,
+      openAIApiKey: apiKey,
```
kamath: what if you're using Anthropic?

ishaksebsib: Hey @filip-michalsky, in PR #446, I used BaseChatModel in LangchainClient to keep it flexible for different models.
```
export class LangchainClient extends LLMClient {
	public type = "langchainClient" as const;
	private model: BaseChatModel;
```
I’m curious about the decision to change it to ChatOpenAI

alexdphan: Hi! It seems like the code is running into a number of linting issues; have you seen this on your end?

alexdphan: reverting back

filip-michalsky: @alexdphan fixed this, and great shoutout to @ishaksebsib for the trick with the base model which allows you to pass a subclass (anthropic, etc) to the LLM Client and this way is more flexible.

filip-michalsky:
@sameelarif should be good to go, pls approve

sameelarif (approved):

PR #495, opened 2025-02-17
avoid race conditions

sameelarif:
# why
We want to avoid having parallel calls to the API as this can result in unexpected behavior related to keeping the `Page` object updated. To prevent this, we throw a custom error  when an API call is made while a previous call is in-progress.

# what changed
Added an `API_STATE` variable to keep track of outbound requests.

# test plan
Evals and internal API tests.

### DIFF
--- a/lib/api.ts
+++ b/lib/api.ts
@@ -3,6 +3,7 @@ import zodToJsonSchema from "zod-to-json-schema";
 import {
   ApiResponse,
   ExecuteActionParams,
+  RaceConditionError,
   StagehandAPIConstructorParams,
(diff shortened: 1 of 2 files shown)


lib/api.ts
```diff
@@ -121,6 +126,12 @@ export class StagehandAPI {
     args,
     params,
   }: ExecuteActionParams): Promise<T> {
+    if (API_STATE.outgoingRequest) {
```
kamath, line 129: this is just on api execution -- what if someone is running `page.evaluate` while an `act` command is in flight?

PR #498, opened 2025-02-18
feat: add Bun runtime check

eastlondoner:
This PR adds a runtime check to detect if Stagehand is being initialized in the Bun environment. Since Playwright does not currently support Bun (see microsoft/playwright#27139), this check prevents initialization and provides a helpful error message directing users to use Node.js instead.\n\nChanges:\n- Add isRunningInBun() utility function to detect Bun runtime\n- Add runtime check in Stagehand's init() method\n- Throw descriptive error with reference to Playwright issue\n\nCloses microsoft/playwright#27139

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -30,7 +30,7 @@ import { StagehandAPI } from "./api";
 import { scriptContent } from "./dom/build/scriptContent";
 import { LLMClient } from "./llm/LLMClient";
 import { LLMProvider } from "./llm/LLMProvider";
-import { logLineToString } from "./utils";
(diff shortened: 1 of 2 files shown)


miguelg719:
Hi @eastlondoner, thanks for the PR! I tested both the SDK on main and your fork and the npm package on `npx create-browser-app` and got it running with bun; playwright included. Can you point out to a specific workflow where you're experiencing this issue? Is it using the latest bun version?

eastlondoner:
Ask @kamath for more details!

I also spoke to people on the bun team who said that playwright issue I linked is corrext

miguelg719:
merged in #509

PR #499, opened 2025-02-18
Changes for Groq demo

eastlondoner:
I wanted to use stagehand with Groq Qwen-2.5-32b - this is a model that's not as smart as Sonnet and it did not work initially. It ran into various problems, changes that I made to get this working are all included in this PR

These improvements make stagehand work better with "dumber" models like Qwen-2.5

- Add "You can only make one tool call per step, you will get the result of the tool call in the next step." to the act system prompt. This stops models from sending multiple tool calls (because stagehand only runs the first tool call) this reduces model confusion and improves speed and cost (fewer return tokens needed).

- Added a note to discourage trying the same actions repeatedly in the act system prompt

- Define the main available actions and communicate them to the LLM. This makes things much clearer for the LLM and helps the LLM not to 'hallucinate' tool calls.

- If the LLM attempts an unavailable action, return to it an error explaining what the available actions are (including actions available on the chosen element) this helps models to self-heal if they make a mistake.

- introduces a "goBack" action to use page.back() models want to use this to recover from mistakes

- moves the actions that the model has taken into "assistant" role message, this is more in line with expected chat interaction behaviour. IMO this helps the models understand that they took the actions and reduces repetition of the same action.

- adds a new return method for the model to return data to the caller. This solves a problem where if you ask Stagehand to return some information it has no way to do it.

- made llm completion support an array of messages rather than just grabbing the first one. This ensures response information from models is not lost

- adds handling for case where the model tool call includes a null element

- adds "commentary" to the steps carried out. This includes non-tool-call content from the model allowing it to keep its side of the conversation going or pass forward thinking tokens

### DIFF
--- a/lib/a11y/utils.ts
+++ b/lib/a11y/utils.ts
@@ -6,6 +6,8 @@ import {
   PlaywrightCommandMethodNotSupportedException,
   PlaywrightCommandException,
 } from "@/types/playwright";
+import { actMethods, getAvailableMethods } from "../actions";
+import { exhaustiveMatchingGuard } from "../utils";
 
 // Parser function for str output
 export function formatSimplifiedTree(
@@ -464,24 +466,54 @@ export async function performPlaywrightMethod(
     },
   });
 
-  if (method === "scrollIntoView") {
-    logger({
-      category: "action",
-      message: "scrolling element into view",
-      level: 2,
-      auxiliary: {
-        xpath: {
-          value: xpath,
-          type: "string",
-        },
-      },
-    });
-    try {
-      await locator
-        .evaluate((element: HTMLElement) => {
-          element.scrollIntoView({ behavior: "smooth", block: "center" });
-        })
-        .catch((e: Error) => {
+  if (actMethods.includes(method as (typeof actMethods)[number])) {
+    const action = method as (typeof actMethods)[number];
+    switch (action) {
+      case "goBack": {
+        await stagehandPage.goBack({
+          waitUntil: "domcontentloaded",
+        });
+        break;
+      }
+      case "scrollIntoView": {
+        logger({
+          category: "action",
+          message: "scrolling element into view",
+          level: 2,
+          auxiliary: {
+            xpath: {
+              value: xpath,
+              type: "string",
+            },
+          },
+        });
+        try {
+          await locator
+            .evaluate((element: HTMLElement) => {
+              element.scrollIntoView({ behavior: "smooth", block: "center" });
+            })
+            .catch((e: Error) => {
+              logger({
+                category: "action",
+                message: "error scrolling element into view",
+                level: 1,
+                auxiliary: {
+                  error: {
+                    value: e.message,
+                    type: "string",
+                  },
+                  trace: {
+                    value: e.stack,
+                    type: "string",
+                  },
+                  xpath: {
+                    value: xpath,
+                    type: "string",
+                  },
+                },
+              });
+            });
+        } catch (e) {
           logger({
             category: "action",
             message: "error scrolling element into view",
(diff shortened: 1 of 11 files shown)


lib/a11y/utils.ts
```diff
@@ -501,89 +533,173 @@ export async function performPlaywrightMethod(
...
+            },
+          });
+
+          throw new PlaywrightCommandException(e.message);
+        }
+        break;
+      }
+      default: {
+        throw exhaustiveMatchingGuard(action);
+      }
```
eastlondoner, line 702: this makes this a type safe switch statement. Every defined action must be implemented or this will not compile

lib/actions.ts
```diff
@@ -0,0 +1,23 @@
...
+const defaultMethods: Set<string> = new Set(getAvailableMethods({} as Locator));
+
+export const actMethods = [
+  "scrollIntoView",
+  "press",
+  "click",
+  "fill",
+  "type",
+  "goBack",
+] as const;
```
eastlondoner, line 12: this defines the additional actions so the LLM can know what they are

lib/prompt.ts
```diff
@@ -16,11 +17,13 @@ You will receive:
 
 
 ## Your Goal / Specification
-You have 2 tools that you can call: doAction, and skipSection. Do action only performs Playwright actions. Do exactly what the user's goal is. Do not perform any other actions or exceed the scope of the goal.
+You have 3 tools that you can call: doAction, returnPlan, returnResult, and skipSection. Do action performs Playwright actions and can record information found so far. Return Plan is used to return a plan for multiple steps. Return Result is used to return either an intermediate result or a final result. Skip Section is used to skip a section of the page. Do exactly what the user's goal is. Do not perform any other actions or exceed the scope of the goal. You can only make one tool call per step, you will get the result of the tool call in the next step.
```
eastlondoner: there are a couple of changes in here. A very important one is
> You can only make one tool call per step, you will get the result of the tool call in the next step.

lib/inference.ts
```diff
@@ -113,14 +121,37 @@ export async function act({
...
 
-    return JSON.parse(toolCalls[0].function.arguments);
+    if (toolCalls[0].function.name === "returnPlan") {
+      const { plan } = JSON.parse(toolCalls[0].function.arguments);
+      return {
+        result: plan,
+        completed: false,
+        commentary: "This is my plan for the next steps",
+      };
+    }
```
eastlondoner: this was an experiment but didn't make a meaningful difference

lib/llm/AnthropicClient.ts
```diff
@@ -227,6 +227,11 @@ export class AnthropicClient extends LLMClient {
       anthropicTools.push(toolDefinition);
     }
 
+    console.log("sending messages to anthropic", {
+      anthropicTools: JSON.stringify(anthropicTools),
+      systemMessage: JSON.stringify(systemMessage),
+      formattedMessages: JSON.stringify(formattedMessages),
+    });
```
eastlondoner: remove

lib/llm/AnthropicClient.ts
```diff
@@ -265,7 +270,10 @@ export class AnthropicClient extends LLMClient {
           message: {
             role: "assistant",
             content:
-              response.content.find((c) => c.type === "text")?.text || null,
+              response.content
+                .filter((c) => c.type === "text")
+                ?.map((c) => c.text)
+                .join("\n") || null,
```
eastlondoner, line 271: this was throwing away response content because it only returns the first instance of type text. Some models return multiple response elements

lib/llm/OpenAIClient.ts
```diff
@@ -341,6 +341,8 @@ export class OpenAIClient extends LLMClient {
 
     const response = await this.client.chat.completions.create(body);
 
+    console.log("openai response", JSON.stringify(response, null, 2));
+
```
eastlondoner: remove log

lib/prompt.ts
```diff
@@ -197,6 +226,47 @@ export const actTools: LLMTool[] = [
...
+    name: "returnPlan",
+    description: "return a plan for multiple steps",
+    parameters: {
+      type: "object",
+      required: ["plan"],
+      properties: {
+        plan: {
+          type: "string",
+          description: "The plan for the next steps",
+        },
```
eastlondoner: remove returnPlan it didn't achieve much

lib/prompt.ts
```diff
@@ -16,11 +17,13 @@ You will receive:
...
+If you have the information needed to complete the goal, return a final result using the returnResult tool setting completed to true.
 If the user's goal will be accomplished after running the playwright action, set completed to true. Better to have completed set to true if your are not sure.
 
 Note 1: If there is a popup on the page for cookies or advertising that has nothing to do with the goal, try to close it first before proceeding. As this can block the goal from being completed.
 Note 2: Sometimes what your are looking for is hidden behind and element you need to interact with. For example, sliders, buttons, etc...
+Note 3: Avoid repeating actions that have already been taken, try something different.
```
eastlondoner, line 26: Added this to discourage infinite retries

lib/handlers/actHandler.ts
```diff
@@ -1235,18 +1312,68 @@ export class StagehandActHandler {
...
       const elementId = response["element"];
-      const xpaths = selectorMap[elementId];
+
+      // If elementId is null, use the document body as the root element
+      let xpaths: string[];
+      if (elementId === null) {
+        xpaths = ["//body"];
+      } else {
+        xpaths = selectorMap[elementId] ?? [];
+      }
```
eastlondoner, line 1364: handle the case where the tool call sets the elementId to null

lib/handlers/actHandler.ts
```diff
@@ -1235,18 +1312,68 @@ export class StagehandActHandler {
         }
       }
 
+      if ("result" in response) {
```
eastlondoner, line 1315: new response type implementation

lib/handlers/actHandler.ts
```diff
@@ -1311,7 +1438,9 @@ export class StagehandActHandler {
 
         // If no XPath was valid, we cannot proceed
         if (!foundXpath || !locator) {
-          throw new Error("None of the provided XPaths could be located.");
+          if (method !== "goBack") {
+            throw new Error("None of the provided XPaths could be located.");
+          }
```
eastlondoner, line 1443: goBack does not require an element

lib/handlers/actHandler.ts
```diff
@@ -1338,7 +1467,15 @@ export class StagehandActHandler {
...
-          `  Reasoning: ${response.why}\n`;
+          (response.why && response.why?.toUpperCase() !== "NONE"
+            ? `  Reasoning: ${response.why}\n`
+            : "") +
+          (response.commentary && response.commentary?.toUpperCase() !== "NULL"
+            ? `  Commentary: ${response.commentary}\n`
+            : "") +
+          (response.findings && response.findings?.toUpperCase() !== "NULL"
+            ? `  Findings: ${response.findings}\n`
+            : "");
```
eastlondoner, line 1478: added commentary and findings for the model to communicate more information to its future self

lib/inference.ts
```diff
@@ -113,14 +121,32 @@ export async function act({
...
       return null;
     }
 
-    return JSON.parse(toolCalls[0].function.arguments);
+    if (toolCalls[0].function.name === "returnResult") {
+      const { result, completed } = JSON.parse(toolCalls[0].function.arguments);
+      return {
+        result,
+        completed,
+        commentary: modelCommentary,
```
eastlondoner, line 141: modelCommentary collects additional (non tool call) messages that the model records, these are often helpful to the model in future. particularly with a thinking model like r1 if this contains thinking tokens

PR #501, opened 2025-02-19
removing/commenting logs

miguelg719:
# why
We're logging data to the browser console through the process dom functions, should be all managed by stagehand logger.

# what changed
Removed unnecessary logs and commented out those of potential interest. If necessary, we can log the outputString/selectorMap though verbose logging on stagehand logger

# test plan


seanmcguire12:
these will be removed in #737

PR #505, opened 2025-02-20
fixing 5s timeout on actHandler

miguelg719:
# why
we're waiting too long after any action that could potentially redirect to a new page

# what changed
removed a timeout of 5s in `performPlaywrightMethod` on potential redirect actions

# test plan
ran evals for `observe_amazon_add_to_cart`  locally and on bb, but ci evals should pinpoint issues at a larger scale

### DIFF
--- a/lib/handlers/actHandler.ts
+++ b/lib/handlers/actHandler.ts
@@ -502,7 +502,6 @@ export class StagehandActHandler {
         },
       });
 
-      // NAVIDNOTE: Should this happen before we wait for locator[method]?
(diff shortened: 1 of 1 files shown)


lib/handlers/actHandler.ts
```diff
@@ -544,7 +543,6 @@ export class StagehandActHandler {
 
       await Promise.race([
```
kamath, line 544: we don't need a `Promise.race` if we're only awaiting one thing -- can we change this to `this.stagehandPage.waitForSettledDom` instead?

PR #508, opened 2025-02-20
stagehand.context pages as stagehand pages and supporting more than one page

miguelg719:
# why
Currently, Stagehand's context only supports managing a single page through `stagehand.page`. This limits users who need to work with multiple browser pages simultaneously, which is a common use case in web automation (e.g., comparing data across tabs, handling popups, or parallel processing).

# what changed
1. Enhanced Context Management
Created EnhancedContext interface that properly extends Playwright's BrowserContext
Modified context.newPage() to return our enhanced Page type with act/extract/observe methods
Modified context.pages() to return an array of enhanced Page instances

2. Page Tracking
Added WeakMap in StagehandContext to track StagehandPage mappings
Maintains memory efficiency by using WeakMap for automatic cleanup

3. Type System Improvements
Updated type definitions to properly reflect enhanced page capabilities
Fixed circular dependencies between StagehandPage and StagehandContext
Ensured type safety when working with multiple pages

Now you can call act/extract/observe on new pages:
```typescript
  const mainPage = stagehand.page;
  await mainPage.goto("https://example.com");

  const context = await stagehand.context;
  const newPage = await context.newPage();
  await newPage.goto("https://google.com");

  const observeRes = await newPage.observe({
    instruction: "click the google search button",
  });
```

# test plan
adding deterministic tests


lib/StagehandContext.ts
```diff
@@ -1,29 +1,112 @@
...
 import { Stagehand } from "./index";
+import { StagehandPage } from "./StagehandPage";
+import { Page } from "../types/page";
+
+// Define the enhanced context type that includes our modified methods
+export interface EnhancedContext
+  extends Omit<PlaywrightContext, "newPage" | "pages"> {
+  newPage(): Promise<Page>;
+  pages(): Page[];
+}
```
sameelarif: should probably move this to the `types` folder and export it from `index.ts`

miguelg719: yup on it

lib/StagehandContext.ts
```diff
@@ -1,29 +1,125 @@
...
+  }
+
+  private async createStagehandPage(
+    page: PlaywrightPage,
+  ): Promise<StagehandPage> {
+    const stagehandPage = await new StagehandPage(
+      page,
+      this.stagehand,
+      this,
+      this.stagehand.getLLMClient(),
```
kamath: would rather just make `llmClient`, `systemPrompt`, etc. `public readonly` access modifier and just do `this.stagehand.llmClient` in `StagehandPage` for consistency

evals/deterministic/tests/BrowserContext/multiPage.test.ts
```diff
@@ -0,0 +1,268 @@
+import { test, expect } from "@playwright/test";
+import { Stagehand } from "@/dist";
+import StagehandConfig from "@/evals/deterministic/stagehand.config";
+import { Page } from "../../../../types/page";
```
kamath: this should import from `dist` to ensure we're importing the built types to avoid a situation in which users can't import the right types

lib/StagehandPage.ts
```diff
@@ -69,7 +83,7 @@ export class StagehandPage {
       },
       on: () => {
         throw new Error(
-          "You seem to be referencing a page in an uninitialized `Stagehand` object. Ensure you are running `await stagehand.init()` on the Stagehand object before referencing the `page` object.",
```
kamath, line 72: ?

lib/StagehandPage.ts
```diff
@@ -209,13 +223,52 @@ export class StagehandPage {
...
     const page = this.intPage;
     const stagehand = this.stagehand;
-    this.intPage = new Proxy(page, {
-      get: (target, prop) => {
+
+    // Clean up the throwing methods before creating the new proxy
+    delete this.intPage.on;
+    delete this.intPage.act;
+    delete this.intPage.extract;
+    delete this.intPage.observe;
```
kamath: this doesn't look right, feels unsafe

lib/index.ts
```diff
@@ -377,6 +377,35 @@ export class Stagehand {
...
+    return this.userProvidedInstructions;
+  }
+
+  public getAPIClient(): StagehandAPI | undefined {
+    return this.apiClient;
+  }
+
+  public getWaitForCaptchaSolves(): boolean {
+    return this.waitForCaptchaSolves;
+  }
```
kamath: make these underlying variables `public readonly`

lib/index.ts
```diff
@@ -377,6 +377,35 @@ export class Stagehand {
...
+
+  public setActivePage(page: StagehandPage): void {
+    this.stagehandPage = page;
+  }
```
kamath: what if we make the input parameter an index? i.e. specify the page in `context.pages()` -- helps prevent edge cases

kamath (changes requested):

evals/deterministic/tests/BrowserContext/multiPage.test.ts
```diff
@@ -0,0 +1,268 @@
...
+    expect(await stagehand.page.title()).toBe("Page 1");
+
+    // Create another page and verify it becomes active
+    const anotherPage = await stagehand.context.newPage();
+    expect(stagehand.page).toBe(anotherPage);
+    expect(stagehand.page).not.toBe(newPage);
+  });
+
+  /**
+   * Test simultaneous operations on multiple pages THIS WILL FAIL
```
seanmcguire12: delete

kamath (approved):

PR #509, opened 2025-02-21
feat: add Bun runtime check to prevent initialization in Bun environment

miguelg719:
From @eastlondoner:

This PR adds a runtime check to detect if Stagehand is being initialized in the Bun environment. Since Playwright does not currently support Bun (see https://github.com/microsoft/playwright/issues/27139), this check prevents initialization and provides a helpful error message directing users to use Node.js instead.\n\nChanges:\n- Add isRunningInBun() utility function to detect Bun runtime\n- Add runtime check in Stagehand's init() method\n- Throw descriptive error with reference to Playwright issue\n\nCloses https://github.com/microsoft/playwright/issues/27139

**This PR will throw an error even on local bun runtimes which probably work**

### DIFF
--- a/.changeset/rotten-buttons-relate.md
+++ b/.changeset/rotten-buttons-relate.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+Bun runs will now throw a more informed error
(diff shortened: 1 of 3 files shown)


seanmcguire12 (approved):

KeKs0r:
Just out of curiosity, what specifically is missing? Since I had a project that was using playwright in bun without issues. Only when switching to stagehand, I am getting this runtime check. So I am curious which errors or things are actually blocking using Bun.

PR #518, opened 2025-02-25
use observe in act

sameelarif:
# why
Using `observe` results to determine the parameters passed into `act` has proven to be a much more efficient way to perform actions. This PR makes the `observe -> act` workflow the default, along with the option to use the old `act()` logic.

# what changed
`act()` now uses `observe()` under the hood.

# test plan
Combination & act evals. There should not be any performance degradations.


lib/StagehandPage.ts
```diff
@@ -487,6 +488,25 @@ export class StagehandPage {
       ? this.stagehand.llmProvider.getClient(modelName, modelClientOptions)
       : this.llmClient;
 
+    if (!slowDomBasedAct) {
```
miguelg719, line 491: We need to account for `selfHeal:true/false` It basically automatically tries the slowDom approach whenever act from observeResult throws an error

sameelarif: Shouldn't `selfHeal` not affect behavior when `slowDomBasedAct` is set to `true`?

sameelarif: There's already logic for this in `actHandler`. Since we're calling `actFromObserveResult` it should already handle self-healing right?

miguelg719: yeah you're right, let me help with prompting constraints

kamath:
image
ran act evals locally, major improvement

kamath (approved):

lib/handlers/actHandler.ts
```diff
@@ -136,11 +140,14 @@ export class StagehandActHandler {
             ? `${method} ${observe.description}`
             : observe.description;
         // Call act with the ObserveResult description
-        await this.stagehandPage.act(actCommand);
+        await this.stagehandPage.act({
+          action: actCommand,
+          slowDomBasedAct: true,
```
miguelg719, line 145: we should explore running through the list of observeResults (for ambiguity) before triggering old result

miguelg719 (approved):
need to go over other optimizations but happy with current version and self-healing

PR #520, opened 2025-02-25
added error thrown on act from a not-supported observe result

miguelg719:
# why
Now that we're appending iframes to observeResults if they are present on the page (#500), we need to avoid passing those through `performPlaywrightMethod`.

# what changed
Added an informed error to `actFromObserveResult` and returning `success:false` on passed observeResults with `method:'not-supported'`

# test plan


lib/handlers/actHandler.ts
```diff
@@ -85,6 +85,29 @@ export class StagehandActHandler {
     });
 
     const method = observe.method;
+    if (method === "not-supported") {
```
kamath, line 89: ideally should be a zod enum that we just validate here

PR #502, opened 2025-02-19
Adding Cerebras inference support

kwt00:
# why
Added Cerebras inference API support to enable using Cerebras's llama models (3.3-70b and 3.1-8b) with Stagehand. This expands model options beyond OpenAI and Anthropic.

# what changed
Added CerebrasClient.ts:
- Handles API communication with Cerebras
- Supports function calling
- Includes robust response parsing with fallbacks
- Handles caching and error logging
- Infrastructure Changes:

Updated LLMClient.ts to support Cerebras provider type:
- Added Cerebras models to model.ts type definitions
- Added Cerebras case to LLMProvider.ts
- Added CEREBRAS_API_KEY to environment variables

Documentation changes:
Added version 1.13.0 to CHANGELOG.md
Created cerebras-integration-details.md with technical details

# test plan
Run form_filling_sensible_cerebras.ts with 70b model (8b model is inconsistent due to tool calling struggles)

PR #523, opened 2025-02-26
Adding Cerebras inference support (#502)

kamath:
just added a changeset to #502


kamath:
merged without approval since #502 was the actual diff

PR #528, opened 2025-02-26
handle close session

seanmcguire12:
# why
- catch error if we attempt to close a session that has already been closed when using the API

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -711,7 +711,21 @@ export class Stagehand {
 
   async close(): Promise<void> {
     if (this.apiClient) {
-      await this.apiClient.end();
+      try {
+        await this.apiClient.end();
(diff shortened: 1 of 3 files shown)


lib/index.ts
```diff
@@ -711,7 +711,21 @@ export class Stagehand {
...
   async close(): Promise<void> {
     if (this.apiClient) {
-      await this.apiClient.end();
+      try {
+        await this.apiClient.end();
+      } catch (error) {
+        if (
+          error instanceof Error &&
+          error.message.includes("Attempted to close session")
+        ) {
```
sameelarif: better to use status codes here imo. maybe have `this.apiClient.end` return  the http response and check for the status code here

sameelarif (approved):

PR #536, opened 2025-02-28
Adding Groq inference support

sankalpgunturi:
# why
Added Groq inference API support to enable using Groq's llama models (3.3-70b versatile and specdec) with Stagehand. This will be an addition to the Stagehand's Gamet of supporting Inference APIs, just like [Cerebras](https://github.com/browserbase/stagehand/pull/502) (this PR is modeled after that one).

# what changed
## Added GroqClient.ts:
- Handles API communication with Groq
- Supports function calling
- Includes robust response parsing with fallbacks
- Handles caching and error logging

## Infrastructure Changes:
- Updated LLMClient.ts to support Groq provider type:
- Added Groq models to model.ts type definitions
- Added Groq case to LLMProvider.ts
- Added GROQ_API_KEY to environment variables

# test plan
- Added `form_filling_sensible_groq.ts`


sankalpgunturi:
Groq has support for following models:

image

However, not all of them worked out well for the form filling use case. I found the 3.3-70b versatile and specdec models from LLaMa to be pretty efficient with Stagehand.

kamath (approved):
thanks so much! merging this into an internal branch to add a changeset and run evals

PR #550, opened 2025-03-04
make slow act the default

sameelarif:
# why
We don't want to make slow act (`act()` that doesn't use an `ObserveResult`) the default for this release.

# what changed
Default `slowDomBasedAct` to `true`.

# test plan
E2E tests

### DIFF
--- a/.changeset/wild-hats-turn.md
+++ b/.changeset/wild-hats-turn.md
@@ -2,4 +2,4 @@
 "@browserbasehq/stagehand": minor
 ---
 
-`act()` now uses `observe()` under the hood. This results in significant performance improvements.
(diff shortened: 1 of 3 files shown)


kamath (approved):

.changeset/wild-hats-turn.md
```diff
@@ -2,4 +2,4 @@
 "@browserbasehq/stagehand": minor
 ---
 
-`act()` now uses `observe()` under the hood. This results in significant performance improvements.
+`act()` can now use `observe()` under the hood, resulting in significant performance improvements. To opt-in to this change, set `slowDomBasedAct: true` in `ActOptions`.
```
kamath, line 5:
```suggestion
`act()` can now use `observe()` under the hood, resulting in significant performance improvements. To opt-in to this change, set `slowDomBasedAct: false` in `ActOptions`.
```

kamath (approved):

PR #556, opened 2025-03-05
Add a timeout to dom based act

kamath:
# why
Customer request -- we can now add timeouts to prevent infinite recursion on `act()`

# what changed
You can now add a timeout to `page.act()` with `timeoutMs` and at a global level at `actTimeoutMs` (default `60_000`)

# test plan
added two evals, one for global timeout, one for each act timeout

### DIFF
--- a/lib/StagehandPage.ts
+++ b/lib/StagehandPage.ts
@@ -487,7 +487,7 @@ export class StagehandPage {
       useVision, // still destructure this but will not pass it on
       variables = {},
       domSettleTimeoutMs,
-      slowDomBasedAct = false,
+      slowDomBasedAct = true,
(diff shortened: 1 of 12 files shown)


lib/StagehandPage.ts
```diff
@@ -487,7 +487,8 @@ export class StagehandPage {
       useVision, // still destructure this but will not pass it on
       variables = {},
       domSettleTimeoutMs,
-      slowDomBasedAct = true,
+      slowDomBasedAct = false,
```
seanmcguire12:
```suggestion
      slowDomBasedAct = true,
```

seanmcguire12 (approved):

PR #563, opened 2025-03-06
support scrolling

seanmcguire12:
# why
- needed for CU/vision
# what changed
- handle scrolling inside `performPlaywrightMethod`
# example usage
```
await stagehand.page.goto("https://aigrant.com/");
await stagehand.page.act({
  action: "scroll 50% down the page.",
  slowDomBasedAct: false,
});

### DIFF
--- a/lib/handlers/actHandler.ts
+++ b/lib/handlers/actHandler.ts
@@ -448,6 +448,91 @@ export class StagehandActHandler {
 
         throw new PlaywrightCommandException(e.message);
       }
+    } else if (method === "scrollTo" || method === "scroll") {
+      this.logger({
+        category: "action",
+        message: "scrolling element vertically to specified percentage",
+        level: 2,
+        auxiliary: {
+          xpath: { value: xpath, type: "string" },
+          coordinate: { value: JSON.stringify(args), type: "string" },
+        },
+      });
+
+      try {
+        const [yArg = "0%"] = args as string[];
+
+        await this.stagehandPage.page.evaluate(
+          ({ xpath, yArg }) => {
+            /**
+             * Helper to evaluate an XPath and return the first node found.
+             */
+            function getNodeFromXpath(theXpath: string) {
+              return document.evaluate(
+                theXpath,
+                document.documentElement,
+                null,
+                XPathResult.FIRST_ORDERED_NODE_TYPE,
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: lib/handlers/actHandler.ts [hunk 1]
```diff
@@ -448,6 +449,78 @@ export class StagehandActHandler {
 
         throw new PlaywrightCommandException(e.message);
       }
+    } else if (method === "scrollTo" || method === "scroll") {
```
miguelg719: I don't see the llm calling scroll by pixel. If anything by % or maybe like we had in chunking something like 'scroll down'. It would be interesting to add some sort of feedback loop like 'successfully scrolled all the way to the bottom' so the agent doesn't get stuck. It seems performPlaywrightMethod doesn't return anything?

sameelarif:
tested and confirmed scrolling with `act` works on my end! did run into some issues though:
1. when `slowDomBasedAct` is `true`, the scrolling works but is then ignored by `verifyActCompletion`, which scrolls down and back to the top of the page.
2. when `slowDomBasedAct` is `false`, the underlying `observe` does not always return a valid `_performPlaywrightMethod` method. For example, it returned `.evaluate` instead of `scroll`/`scrollTo`/`mouse.wheel`. This may be an underlying issue with our schema for the `observe` LLM call response, but it's causing scrolling to fail here.

seanmcguire12:
> tested and confirmed scrolling with `act` works on my end! did run into some issues though:
>     1. when `slowDomBasedAct` is `true`, the scrolling works but is then ignored by `verifyActCompletion`, which scrolls down and back to the top of the page.
> 2. when `slowDomBasedAct` is `false`, the underlying `observe` does not always return a valid `_performPlaywrightMethod` method. For example, it returned `.evaluate` instead of `scroll`/`scrollTo`/`mouse.wheel`. This may be an underlying issue with our schema for the `observe` LLM call response, but it's causing scrolling to fail here.

yeah its not supposed to work if `slowDomBasedAct` is true. should prob add docs on this. nice find on the `_performPlaywrightMethod` issue. if only #569 was in, then i could easily debug 👀 @kamath

sameelarif (approved):

miguelg719 (approved):

PR #566, opened 2025-03-06
expose a11y

kamath:
# why

# what changed

# test plan

### DIFF
--- a/examples/example.ts
+++ b/examples/example.ts
@@ -17,6 +17,7 @@ async function example() {
   /**
    * Add your code here!
    */
+  // delete before pushing to main
   const tree = await stagehand.page.getAccessibilityTree();
   console.log(tree);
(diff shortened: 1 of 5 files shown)


examples/example.ts
```diff
@@ -17,6 +17,9 @@ async function example() {
   /**
    * Add your code here!
    */
+  // delete before pushing to main
+  const tree = await stagehand.page.getAccessibilityTree();
+  console.log(tree);
```
miguelg719, line 22: ideally what we log/write to a file is tree.simplified (the txt version passed into the llm)

kamath: i kinda like both? allows the user to play with it

miguelg719: yea, the whole tree is kinda overkill but with this structure both are returned

PR #568, opened 2025-03-10
add connect url from api

kamath:
# why
We shouldn't be mapping regions internally

# what changed
Get connect URL from API (with BB SDK 2.4.0)

# test plan
simple resume session in example


lib/index.ts
```diff
@@ -97,6 +91,7 @@ async function getBrowser(
       try {
         const sessionStatus =
           await browserbase.sessions.retrieve(browserbaseSessionID);
+        console.log("GOT SESSION STATUS", sessionStatus);
```
miguelg719: this.logger?

package.json
```diff
@@ -82,8 +82,8 @@
     "zod": "^3.23.8"
   },
   "dependencies": {
-    "@anthropic-ai/sdk": "0.39.0",
-    "@browserbasehq/sdk": "^2.0.0",
+    "@anthropic-ai/sdk": "^0.27.3",
```
miguelg719:
```suggestion
    "@anthropic-ai/sdk": "0.39.0",
```

miguelg719:
still missing changes

PR #573, opened 2025-03-11
return act result in actFromObserve

seanmcguire12:
# why
- not returning result from `page.act` in `actFromObserve`


kamath (approved):
technically this PR is mine but i'll let you have it

seanmcguire12:
> technically this PR is mine but i'll let you have it

stop the cap

PR #576, opened 2025-03-12
wait for page navigation after key press

seanmcguire12:
# why
- key presses can trigger page navigation, but we currently only handle this after we execute a `click` function. 
- this PR ensures we handle possible page navigation after `press` in addition to `click`


miguelg719 (approved):
Lgtm, we should explore improving handle navigation entirely though for pages that take too long to load (like CUA not waiting for networkidle to take actions on the page)

PR #584, opened 2025-03-15
remove healthcheck ping

miguelg719:
# why
Healthcheck is unneccessary from the client

# what changed
Routes changed, deprecating healthcheck ping

# test plan


kamath (approved):
approved pending changeset + PR description

sameelarif (approved):

PR #586, opened 2025-03-16
operator handler

sameelarif:
# why
Naturally, we want to provide an agent that uses Staghand's native capabilities. This PR adds a default agent that doesn't utilize computer-use models.

# what changed
Added a `StagehandOperatorHandler` class which provides the logic for action planning and execution.

# test plan
E2E

# TODO

- [x] Export response schemas from `types`.
- [x] Finalize which primitives are exposed to the API. 
- [x] If the goal of the agent is data extraction, return it in the `AgentResult["message"]` field.
- [x] In the next prompt, provide the result of the previous method.

### DIFF
--- a/lib/handlers/operatorHandler.ts
+++ b/lib/handlers/operatorHandler.ts
@@ -1,23 +1,22 @@
-import {
-  AgentAction,
-  AgentExecuteOptions,
-  AgentResult,
-  ActionExecutionResult,
-} from "@/types/agent";
+import { AgentAction, AgentExecuteOptions, AgentResult } from "@/types/agent";
 import { LogLine } from "@/types/log";
-import { OperatorResponse, operatorResponseSchema } from "@/types/operator";
+import {
+  OperatorResponse,
+  operatorResponseSchema,
+  OperatorSummary,
+  operatorSummarySchema,
+} from "@/types/operator";
 import { LLMParsedResponse } from "../inference";
 import { ChatMessage, LLMClient } from "../llm/LLMClient";
 import { buildOperatorSystemPrompt } from "../prompt";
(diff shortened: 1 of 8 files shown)


lib/handlers/operatorHandler.ts
```diff
@@ -0,0 +1,188 @@
...
+import { LLMParsedResponse } from "../inference";
+import { ChatMessage, LLMClient } from "../llm/LLMClient";
+import { buildOperatorSystemPrompt } from "../prompt";
+import { StagehandPage } from "../StagehandPage";
+
+export class StagehandOperatorHandler {
+  private stagehandPage: StagehandPage;
+  private logger: (message: LogLine) => void;
+  private llmClient: LLMClient;
+  messages: ChatMessage[];
```
kamath: In a fast follow PR, let's use stagehand.history here

kamath (commented):
Code generally lgtm from reviewing on my phone, will review properly later

miguelg719 (approved):
This is great

kamath (approved):

lib/handlers/operatorHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+        this.messages.push({
+          role: "user",
+          content: [
+            {
+              type: "text",
+              text: messageText,
+            },
+            {
+              type: "image_url",
+              image_url: { url: `data:image/png;base64,${base64Image}` },
```
miguelg719, line 90: anthropic's format:
```ts
messages: [
      {
          "role": "user",
          "content": [
              {
                  "type": "image",
                  "source": {
                      "type": "base64",
                      "media_type": image_media_type,
                      "data": image_data,
                  },
              },
...
```

PR #588, opened 2025-03-16
add agent API support

sameelarif:
# why
To support offloading agentic tasks to the API.

# what changed
Added logic for sending requests to the API with the agent config and execution options.

# test plan
E2E


miguelg719 (approved):
tested locally w/ api

PR #589, opened 2025-03-16
Override screenshot logic

miguelg719:
# why
Adding CDP support for screenshots 

# what changed
Stagehand override for `screenshot()`. There are multiple benefits to using CDP for screenshots rather than automation frameworks: speed, memory, reliability... For more details: check [these docs](https://docs.browserbase.com/features/screenshots#why-use-cdp-for-screenshots%3F)
# test plan

### DIFF
--- a/.changeset/free-pots-move.md
+++ b/.changeset/free-pots-move.md
@@ -1,5 +1,5 @@
 ---
-"@browserbasehq/stagehand": minor
+"@browserbasehq/stagehand": patch
 ---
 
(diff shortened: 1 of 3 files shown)


.changeset/free-pots-move.md
```diff
@@ -0,0 +1,5 @@
+---
+"@browserbasehq/stagehand": minor
```
kamath:
```suggestion
"@browserbasehq/stagehand": patch
```

this isn't a user-facing change

PR #595, opened 2025-03-17
add supported methods to `observe` prompt 

seanmcguire12:
# why
- we should include a list of 'officially supported' methods to the `observe` prompt so that the LLM has a better understanding of what method to suggest
- this lays the foundation for expanding beyond playwright: for example, we can just add new methods to the list of supported methods that we pass to the LLM
# what changed
- added a prompt that contains a list of 'officially supported' methods
- this prompt is added when we are calling `observe` from within `act`
- this prompt is also added when the `returnAction` flag is set within `observeOptions`

### DIFF
--- a/.changeset/cool-lemons-report.md
+++ b/.changeset/cool-lemons-report.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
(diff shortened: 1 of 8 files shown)


kamath:
anirudh says hello from stagehand mcp

PR #601, opened 2025-03-20
custom error classes

seanmcguire12:
# why
- easier to debug
# what changed
- added custom error classes for stagehand and stagehand API

### DIFF
--- a/types/stagehandErrors.ts
+++ b/types/stagehandErrors.ts
@@ -1,0 +1,85 @@
+export class StagehandError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = this.constructor.name;
+  }
+}
+
+export class StagehandEnvironmentError extends StagehandError {
+  constructor(
+    currentEnvironment: string,
+    requiredEnvironment: string,
+    feature: string,
+  ) {
+    super(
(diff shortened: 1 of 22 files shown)


types/stagehandErrors.ts
```diff
@@ -0,0 +1,97 @@
+export class StagehandError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = this.constructor.name;
+  }
+}
+
```
kamath, line 7: let's make a `StagehandDefaultError` that extends `StagehandError` with the message `"Hey! We're sorry you ran into an error. If you need help, please open a Github issue or send us a Slack message (link to both)"`

kamath: ^ this should go into `init`, act/extract/observe global

sameelarif (commented):
there are still a lot of instances of Stagehand throwing a default `Error`. if you do a global search for `throw new Error` you'll see a bunch in `lib/llm` and more elsewhere

sameelarif (approved):

kamath (approved):
need to add errors to api and check for those errors and throw the same errors client-side

PR #604, opened 2025-03-21
delete old CU models - won't be supported

miguelg719:
# why

# what changed

# test plan


kamath (approved):
get those contrib numbers

PR #611, opened 2025-03-28
added stagehand flag in user metadata

miguelg719:
# why
To know which BB sessions are created with Stagehand

# what changed
Added a user-metadata key:value pair `{'stagehand_created': 'true'}`

# test plan

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -186,7 +186,7 @@ async function getBrowser(
         ...browserbaseSessionCreateParams,
         userMetadata: {
           ...(browserbaseSessionCreateParams?.userMetadata || {}),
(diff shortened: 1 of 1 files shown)


lib/index.ts
```diff
@@ -184,6 +184,10 @@ async function getBrowser(
       const session = await browserbase.sessions.create({
         projectId,
         ...browserbaseSessionCreateParams,
+        userMetadata: {
+          ...(browserbaseSessionCreateParams?.userMetadata || {}),
+          stagehand_created: "true",
```
kamath:
```suggestion
          stagehand: "true",
```

kamath (approved):

PR #613, opened 2025-03-29
history fixes

sameelarif:
# why
observeAct wasn't being stored in history

# what changed
Added a `addToHistory` call

# test plan

### DIFF
--- a/lib/StagehandPage.ts
+++ b/lib/StagehandPage.ts
@@ -56,7 +56,7 @@ export class StagehandPage {
   private _history: Array<HistoryEntry> = [];
 
   public get history(): ReadonlyArray<HistoryEntry> {
-    return this._history;
+    return Object.freeze([...this._history]);
   }
 
   constructor(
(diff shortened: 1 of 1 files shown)


lib/StagehandPage.ts
```diff
@@ -55,7 +55,7 @@ export class StagehandPage {
   private _history: Array<HistoryEntry> = [];
 
   public get history(): ReadonlyArray<HistoryEntry> {
-    return this._history;
+    return Object.freeze([...this._history]);
```
kamath, line 58: what does this change?

lib/StagehandPage.ts
```diff
@@ -552,12 +557,15 @@ export class StagehandPage {
       });
 
       // `useVision` is no longer passed to the handler
```
kamath:
```suggestion
```

kamath (commented):
can you add an eval here as well

lib/StagehandPage.ts
sameelarif: should enforce that `this._history` is immutable instead of just declaring it as that type

kamath (approved):

PR #615, opened 2025-03-30
add hooks to agent

sameelarif:
# why

# what changed

# test plan

### DIFF
--- a/lib/handlers/operatorHandler.ts
+++ b/lib/handlers/operatorHandler.ts
@@ -6,30 +6,33 @@ import {
   OperatorSummary,
   operatorSummarySchema,
 } from "@/types/operator";
-import { LLMParsedResponse } from "../inference";
-import { ChatMessage, LLMClient } from "../llm/LLMClient";
-import { buildOperatorSystemPrompt } from "../prompt";
-import { StagehandPage } from "../StagehandPage";
-import { ObserveResult } from "@/types/stagehand";
+import { AgentConfig, ObserveResult } from "@/types/stagehand";
 import {
   StagehandError,
   StagehandMissingArgumentError,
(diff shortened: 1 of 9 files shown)


lib/handlers/operatorHandler.ts
```diff
@@ -46,106 +49,121 @@ export class StagehandOperatorHandler {
...
-        );
-      }
-      let extractionResult: unknown | undefined;
-      if (result.method === "extract") {
-        extractionResult = await this.stagehandPage.page.extract(
-          result.parameters,
-        );
+        let playwrightArguments: ObserveResult | undefined;
+        if (result.method === "act") {
+          [playwrightArguments] = await this.stagehandPage.page.observe(
```
miguelg719, line 122: shouldn't the stagehandPage object already have act/extract observe?

lib/handlers/operatorHandler.ts
```diff
@@ -46,106 +49,121 @@ export class StagehandOperatorHandler {
...
 
-      await this.executeAction(result, playwrightArguments, extractionResult);
+      const finalResult: AgentResult = {
+        success: true,
+        message: await this.getSummary(options.instruction),
+        actions,
+        completed:
+          actions.length > 0
+            ? (actions[actions.length - 1].taskCompleted as boolean)
+            : false,
```
miguelg719, line 157: This was a good catch/suggestion, but I'm p sure we should throw an error here not return false. The previous code was kinda catching this without being informative

miguelg719: just need to confirm cases where the agent sends a message without taking an action?

vuminhluan:
Amazing work! I really need this feature.

Really looking forward to using it soon!!!

PR #617, opened 2025-03-30
a11y extract

seanmcguire12:
# why
- fast
# what changed
- use a11y tree
# test plan
- evals

### DIFF
--- a/.changeset/forty-symbols-crash.md
+++ b/.changeset/forty-symbols-crash.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+use a11y tree for default extract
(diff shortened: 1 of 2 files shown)


kamath (approved):

MrChrisRodriguez:
@seanmcguire12 , I'm just becoming familiar with this, but does it really make sense to replace the old method with the a11y method (as opposed to introducing it as a third option?)

With the a11y method, we're no longer able to easily extract hyperlink information (like the href of an anchor tag) since that frequently isn't explicitly made available to the accessibility tree. In the old DOM method we'd have a lot more input tokens, but we would have access to that data (which is the point of using the domExtract method instead of textExtract)

Please correct me if I'm way off base or misunderstanding!

seanmcguire12:
@MrChrisRodriguez Hey Christian, I appreciate the feedback, and you're not off base at all! The thought process here was that we wanted to prioritize speed (got a lot of feedback from people saying extract was too slow). We could have added a third option, but we didn't want to overwhelm people with too many confusing parameters. 

We have plans to add support for link extraction very soon, so we can give users the best of both worlds!

If you have any more feedback about what you liked about the old version of extract/what you'd might want in the future, please feel free to join the [slack channel](https://stagehand.dev/slack)!!

PR #620, opened 2025-03-30
Add support for custom OpenAI provider

kamath:
# why
Sometimes you want to wrap the entire OpenAI object, like with Braintrust or LiteLLM

# what changed
Add support for passing in an OpenAI object as an `llmClient` like so:

```
import { CustomOpenAIClient } from "./external_clients/customOpenAI";

    ...StagehandConfig,
    llmClient: new CustomOpenAIClient({
      modelName: "llama3.3",
      client: new OpenAI({
        apiKey: "ollama",
        baseUrl: "http://localhost:11434"
      }),
    }),
  });
```

# test plan
Add `hn_customOpenAI` eval

### DIFF
--- a/examples/external_clients/customOpenAI.ts
+++ b/examples/external_clients/customOpenAI.ts
@@ -34,21 +34,7 @@ export class CustomOpenAIClient extends LLMClient {
   public type = "openai" as const;
   private client: OpenAI;
 
-  constructor({
-    modelName,
-    client,
-    enableCaching = false,
-  }: {
-    modelName?: string;
-    client?: OpenAI;
-    enableCaching?: boolean;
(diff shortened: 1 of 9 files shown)


examples/external_clients/customOpenAI.ts
seanmcguire12, line 1: do we want to remove the Ollama client or should we keep it and just make the customOpenAI client a new one?

kamath: i did remove

kamath: no need to have both

seanmcguire12 (approved):

PR #625, opened 2025-03-31
give act its own inference call

seanmcguire12:
# why
- we are calling `observeHandler` inside `actHandler`. This means that the downstream inference call is just an observe inference call. 
- this means that when someone calls `act`, the token metrics are just being added to `observe` token metrics instead of `act`
- it also makes the code confusing
# what changed
- gave `act` its own inference call
# test plan
- evals

# notes
- updated the act schema description (which is given to the LLM) so that it knows the pattern we expect to handle for variable parsing

### DIFF
--- a/evals/tasks/hn_langchain.ts
+++ b/evals/tasks/hn_langchain.ts
@@ -9,7 +9,7 @@ export const hn_langchain: EvalFunction = async ({ logger }) => {
     logger,
     llmClient: new LangchainClient(
       new ChatOpenAI({
-        model: "gpt-4o",
+        model: "gpt-4o-mini",
       }),
     ),
   });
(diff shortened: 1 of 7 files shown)


evals/tasks/hn_langchain.ts
```diff
@@ -9,7 +9,7 @@ export const hn_langchain: EvalFunction = async ({ logger }) => {
     logger,
     llmClient: new LangchainClient(
       new ChatOpenAI({
-        model: "gpt-4o",
```
sameelarif, line 12: did you mean to leave this in

sameelarif (approved):

evals/tasks/hn_langchain.ts
seanmcguire12: yes this was intentional

Issue #592, opened 2025-03-17
Stagehand env set to "BROWSERBASE", but API key is undefined—defaults to LOCAL instead of throwing an error


valentindoering:
When the stagehand env is set to "BROWSERBASE" in the constructor without a undefined API key, Stagehand defaults to "LOCAL" instead of issuing an error or warning.

kamath:
Thanks for the feedback! PRs welcome here :)

yash1744:
It looks like in `stagehand.init()` at this point [L674](https://github.com/browserbase/stagehand/blob/main/lib/index.ts#L674)  `this.env` is modified to `LOCAL` as in [L621](https://github.com/browserbase/stagehand/blob/main/lib/index.ts#L621) because the API key is undefined, which in turn makes [L80-L81](https://github.com/browserbase/stagehand/blob/main/lib/index.ts#L80-L81) ineffective as `env` is `LOCAL` at the time of matching those conditions.  

The solution for this would be passing the `intEnv` variable to `getBrowser` and matching `intEnv` instead of `env` at [L80-L81](https://github.com/browserbase/stagehand/blob/main/lib/index.ts#L80-L81) .

omlahore:
Looks like this was already resolved  — nice work!

I had been working on a fix as well but just saw the PR now. All good on my end, just flagging this so the issue can be closed if everything’s now in place

PR #629, opened 2025-04-02
fixed output message when apikey is not present

yash1744:
# why 
solves issue #592. 

# what changed
Added `intEnv` in `getBrowser` method in order to know about initial environment and replaced `env` with `intEnv`  for checking the environment correctly and outputting the message accordingly.

# test plan


omlahore:
Looks like this was already resolved in #642 — nice work!

I had been working on a fix as well but just saw the PR now. All good on my end, just flagging this so the issue can be closed if everything’s now in place.

monadoid:
Thanks for your contribution - closing this PR as getBrowser no longer exists in v3.

PR #630, opened 2025-04-02
Make type and modelName more flexible in LLMClient.ts

mbleigh:
# why

Right now `modelName` is strongly typed to only allow the pre-specified model names in AvailableModel. I'm building a custom LLMClient that doesn't use those models.

# what changed

This PR updates it to also accept an arbitrary string, and also updates type to use the `(string & {})` [trick](https://stackoverflow.com/questions/76142160/literal-string-union-autocomplete-in-typescript-3-8-3) to enable autocomplete with a string fallback.

# test plan

Types-only change, should be okay? :)

### DIFF
--- a/lib/llm/LLMClient.ts
+++ b/lib/llm/LLMClient.ts
@@ -86,8 +86,8 @@ export interface CreateChatCompletionOptions {
 }
 
 export abstract class LLMClient {
-  public type: "openai" | "anthropic" | "cerebras" | "groq" | string;
-  public modelName: AvailableModel;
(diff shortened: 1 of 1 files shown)


kamath:
thanks so much for taking the time to try stagehand! curious why you wouldn't want a string for `modelName`?

mbleigh:
> thanks so much for taking the time to try stagehand! curious why you wouldn't want a string for `modelName`?

I do want a string for model name - that's what I'm adding here. The current implementation *only* allows model name to be one of the enumerated models specified. I'm adding support for an arbitrary string in addition to those.

kamath (approved):

PR #633, opened 2025-04-04
fix the getBrowser function

miguelg719:
# why
To avoid redundant calls to BB api

# what changed
Make one single call instead of 2 for getting session status and rest of session data

# test plan
- [x] tested locally


lib/index.ts
```diff
@@ -114,18 +106,28 @@ async function getBrowser(
...
+          category: "init",
+          message: "connect:",
+          level: 1,
+          auxiliary: {
+            connectUrl: {
+              value: connectUrl,
+              type: "string",
+            },
+          },
+        });
```
miguelg719: need to double check this might be transpiled from other work

sameelarif (approved):

PR #636, opened 2025-04-05
token usage for act

seanmcguire12:
# why
- add token usage for act
# what changed
- pass `fromAct` parameter from within act handler -> observe handler -> observe inference if `observe` is being called inside `act`
# test plan
- evals


kamath (approved):
i see the drawbacks here, but ig we can deal with that once it starts to become a pain

PR #638, opened 2025-04-05
Added support for Invalid LLM API KEYS

yash1744:
# why
Previously it is displaying a  `StagehandDefaultError` when LLM api keys are not present which does not provide concise details about the error.

# what changed
It now displays invalid LLM API keys with a clean structured message. 
Added error handling in ```createChatCompletion``` method for various LLM Clients ( OpenAI, Grok, Anthropic, Cerebras ).
Created new ```InvalidLLMKeyError``` class for custom Invalid api message

# test plan

### DIFF
--- a/lib/llm/AnthropicClient.ts
+++ b/lib/llm/AnthropicClient.ts
@@ -14,7 +14,10 @@ import {
   LLMClient,
   LLMResponse,
 } from "./LLMClient";
-import { CreateChatCompletionResponseError } from "@/types/stagehandErrors";
+import {
+  CreateChatCompletionResponseError,
+  InvalidLLMKeyError,
+} from "@/types/stagehandErrors";
 
 export class AnthropicClient extends LLMClient {
   public type = "anthropic" as const;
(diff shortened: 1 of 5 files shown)


kamath:
thanks so much for contributing to stagehand! instead of modifying each client individually, can we instead change `LLMClient` such that it can call `createChatCompletion` and catch it with an error, then throw a `ChatCompletionError`?

also, please add a changeset - just run `npx changeset`, and this is a `minor` PR

yash1744:
Yeah, that sounds like a good idea. But since `createChatCompletion` is an abstract method, we can’t directly instantiate it
Instead, I can create a new method like `safeCreateChatCompletion` that wraps the logic and handles errors. It can throw a common `ChatCompletionError` instead of introducing a new `InvalidLLMKey` class and currently two methods in `operatorHandler.ts` file point to `createChatCompletion` which we need to modify it to new one.

However, even with that, we’ll still need to check which client is being used — like OpenAI, Groq, Cerebras (which are similar using OPEN AI client), or Anthropic, which has a different API and error pattern and handle both errors. 

So in the future, if we add a new client, we’ll need to update this part again to support its error format.

Do you want me to move in this way!

PR #642, opened 2025-04-06
fix: throw error if BROWSERBASE env is set without API key (fixes #592)

### LINKED ISSUE #592: Stagehand env set to "BROWSERBASE", but API key is undefined—defaults to LOCAL instead of throwing an error

omlahore:
Fixes #592

# why
Previously, the fallback to LOCAL mode was silent, which could be misleading in production. Now, the constructor and `getBrowser()` both throw informative errors to help developers catch misconfigurations early.

# what changed
This PR adds an explicit error when `env: "BROWSERBASE"` is used without a valid `BROWSERBASE_API_KEY`.

# test plan
Tested with `npm run example` and verified the error is thrown as expected.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -79,14 +79,11 @@ async function getBrowser(
 ): Promise<BrowserResult> {
   if (env === "BROWSERBASE") {
     if (!apiKey) {
-      logger({
-        category: "init",
-        message:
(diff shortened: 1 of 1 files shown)


kamath (approved):
thanks for your contribution! continuing this in #633

Issue #643, opened 2025-04-07
Stagehand Agent mode should allow providing baseURL for OpenAICUAClient


majkiw:
# Background
Hi, when configuring `Stagehand` itself, you can override llmClient options, or even the whole `llmClient` like this:
```
const stagehand = new Stagehand({
  browserbaseSessionID: browserSession.sessionId,
  modelClientOptions: {
    apiKey: "",
    baseURL: "",
  },
  llmClient: CustomOpenAIClient()
```
💯 

# Problem
However when running `stagehand.agent` you can only set the `apiKey` via:
```
const agent = stagehand.agent({
  options: {
    apiKey: process.env.OPENAI_API_KEY,
  },
```

Because all the other parameters, most importantly `baseURL`, are ignored in the `OpenAICUAClient`:
```
    // Store client options for reference
    this.clientOptions = {
      apiKey: this.apiKey,
    };

    // Initialize the OpenAI client
    this.client = new OpenAI(this.clientOptions);
```

# Ask
Could `stagehand.agent` support the whole `export type ClientOptions = OpenAIClientOptions | AnthropicClientOptions;` options, just like `Stagehand` constructor does?
At the very least, the `baseURL` (that nota bene seems to be passed in the `AnthropicCUAClient.ts`)

For bonus points: allow for a full `llmClient` override, just like in the `Stagehand` constructor to support custom instrumentation.

majkiw:
I went ahead and created the minimal PR that respects `baseURL` in the `OpenAICUAClient.ts` (matching the `AnthropicCUAClient.ts`): https://github.com/browserbase/stagehand/pull/644

bpetty-interlace:
Isn't this required to work with Azure OpenAI?
This seems like an important enhancement.

Kamilczak020:
This is also required for any agent setup that sits behind a proxy - as of right now, it does not respect it.
This is true for OpenAI as well as other model SDKs, such as Anthropic

PR #644, opened 2025-04-07
OpenAICUAClient.ts reads `baseUrl` from `clientOptions`

majkiw:
# why
https://github.com/browserbase/stagehand/issues/643

# what changed
OpenAICUAClient now respects `baseURL` passed in the options

### DIFF
--- a/lib/agent/OpenAICUAClient.ts
+++ b/lib/agent/OpenAICUAClient.ts
@@ -43,6 +43,7 @@ export class OpenAICUAClient extends AgentClient {
       (clientOptions?.apiKey as string) || process.env.OPENAI_API_KEY || "";
     this.organization =
       (clientOptions?.organization as string) || process.env.OPENAI_ORG;
+    this.baseURL = (clientOptions?.baseURL as string) || undefined;
 
     // Get environment if specified
     if (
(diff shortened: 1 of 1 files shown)


kamath:
Thanks so much for opening this PR! Ideally this would take the entire `clientOptions` type -- happy to take that on unless you want to add that to this PR

majkiw:
Thanks @kamath!
I thought I'd add that, too, but I ran into some backward compatibility questions.
Probably better if I just pass it to you then.

As I mentioned in the [issue](https://github.com/browserbase/stagehand/issues/643), ideally, we would be able to pass the whole `llmClient` to support custom instrumentation for better observability (just like for the `Stagehand` constructor).

But for now, to unblock our pilot implementation, just that `baseURL` would be very helpful 🙏

packages/core/lib/v3/agent/AnthropicCUAClient.ts
```diff
@@ -53,13 +55,7 @@ export class AnthropicCUAClient extends AgentClient {
     }
 
     // Store client options for reference
-    this.clientOptions = {
-      apiKey: this.apiKey,
-    };
-
-    if (this.baseURL) {
-      this.clientOptions.baseUrl = this.baseURL;
```
majkiw, line 68: This probably didn't work as it was setting `baseUrl` instead of `baseURL`

PR #652, opened 2025-04-09
feat: handle key press

hitesh-ag1:
# why
enable actions using key press commands on screen

# what changed
- Added press key, val in `SupportedPlaywrightAction` enum
- Modified instruction prompt under `buildActObservePrompt`
- Added two examples - `examples/google_enter.ts`, `examples/try_wordle.ts`

# test plan
- Tested key presses in google search and wordle


kamath:
merged in #661! congrats on your first contribution!

PR #668, opened 2025-04-15
Stagehand Evaluator & first agent evals

miguelg719:
# why
Deterministic evals require high maintenance. A balanced approach is what most frontier labs use for evaluating models: **_LLMs as judges_.** Although not as deterministic as our current evals, this new category scales better while randomizing the potential judging error. Introducing:
### Stagehand Evaluator

# what changed
Created a new class `Evaluator` which reuses a stagehand object to leverage the `llmClient` functionality, sending a screenshot (currently, maybe text-only in the future) asking a question or set of questions (batch). 

Example usage:
```ts
  const evaluator = new Evaluator(stagehand);

  // single evaluation
  const result = await evaluator.evaluate({
    question: "Is the form name input filled with 'John Smith'?"
  });

  // batch evaluate
  const results = await evaluator.batchEvaluate({
    questions: [
      "Is the form name input filled with 'John Smith'?",
      "Is the form email input filled with 'john.smith@example.com'?",
      "Is the 'Are you the domain owner?' option selected as 'No'?",
    ],
    strictResponse: true, // whether or not to throw an error on evaulations !== YES | NO
  });
```

Evaluator will then return a response in the following format:
```
{
   evaluation: "YES" | "NO",
   reasoning:  "reasoning about the evaluation process"
}
```

Also added a few evals for agent that use `Evaluator` as the judge

# test plan
- [x] Tested with local evals
- [x] Tested with BB evals

### DIFF
--- a/evals/evaluator.ts
+++ b/evals/evaluator.ts
@@ -6,69 +6,15 @@
  */
 
 import { AvailableModel, ClientOptions, Stagehand } from "@/dist";
+import { LLMResponseError } from "@/types/stagehandErrors";
 import dotenv from "dotenv";
+import {
+  EvaluateOptions,
+  EvaluationResult,
+  BatchEvaluateOptions,
+} from "@/types/evaluator";
 
 dotenv.config();
-
-export interface EvaluateOptions {
-  /**
-   * The question to ask about the task state
-   */
-  question: string;
-  /**
-   * Custom system prompt for the evaluator
-   */
-  systemPrompt?: string;
-  /**
-   * Delay in milliseconds before taking the screenshot
-   * @default 1000
-   */
-  screenshotDelayMs?: number;
-  /**
-   * Whether to throw an error if the response is not a clear YES or NO
-   * @default false
-   */
-  strictResponse?: boolean;
-}
-
-export interface BatchEvaluateOptions {
-  /**
-   * Array of questions to evaluate
-   */
-  questions: string[];
-  /**
-   * Custom system prompt for the evaluator
-   */
-  systemPrompt?: string;
-  /**
-   * Delay in milliseconds before taking the screenshot
-   * @default 1000
-   */
-  screenshotDelayMs?: number;
-  /**
-   * Whether to throw an error if any response is not a clear YES or NO
-   * @default false
-   */
-  strictResponse?: boolean;
-  /**
-   * The reasoning behind the evaluation
-   */
-  reasoning?: string;
-}
-
-/**
- * Result of an evaluation
- */
-export interface EvaluationResult {
-  /**
-   * The evaluation result ('YES', 'NO', or 'INVALID' if parsing failed or value was unexpected)
-   */
-  evaluation: "YES" | "NO" | "INVALID";
-  /**
-   * The reasoning behind the evaluation
-   */
-  reasoning: string;
-}
 
 export class Evaluator {
   private stagehand: Stagehand;
@@ -117,7 +63,7 @@ export class Evaluator {
     );
 
     const response = await llmClient.createChatCompletion({
-      logger: () => {},
+      logger: this.stagehand.logger,
       options: {
         messages: [
           { role: "system", content: systemPrompt },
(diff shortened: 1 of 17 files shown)


evals/initStagehand.ts
```diff
@@ -33,6 +34,15 @@ const StagehandConfig = {
...
   disablePino: true,
+  browserbaseSessionCreateParams: {
+    projectId: process.env.BROWSERBASE_PROJECT_ID!,
+    browserSettings: {
+      viewport: {
+        width: 1024,
+        height: 768,
+      },
+    },
+  },
```
miguelg719, line 45: can this mess with existing evals? should we make it an agent category property?

evals/evaluator.ts
```diff
@@ -0,0 +1,359 @@
...
+ * usually represented as a screenshot.
+ * The evaluator will reply with YES or NO given the state of the provided task.
+ */
+
+import { AvailableModel, ClientOptions, Stagehand } from "@/dist";
+import dotenv from "dotenv";
+
+dotenv.config();
+
+export interface EvaluateOptions {
```
miguelg719: need to move types

peytoncasper:
My only comment would be that in general for LLM-as-a-Judge you should enforce JSON output with a schema rather than freeform text. This helps drive more determinism in my experience even if its just

 {"passed": True}

You can also then pass in response_mime_type: "application/json" in Gemini's case which does token filtering. I believe this is one of the largest drivers of Gemini's data extraction capabilities. 

Other than that, great idea

EDIT

Just realized you are doing that essentially but providing YES | NO as options inside a JSON object. Apologies.

miguelg719:
Thanks for the pointer on the mime type @peytoncasper! I was hoping to integrate that next into the gemini client abstraction

evals/initStagehand.ts
seanmcguire12: should be fine

evals/evaluator.ts
```diff
@@ -0,0 +1,359 @@
...
+        JSON.parse(cleanedResponse);
+
+      // Validate structure
+      if (
+        typeof parsedResult !== "object" ||
+        parsedResult === null ||
+        typeof parsedResult.evaluation !== "string" ||
+        typeof parsedResult.reasoning !== "string"
+      ) {
+        throw new Error(
```
seanmcguire12: can we add some custom errors?

evals/index.eval.ts
```diff
@@ -322,6 +388,7 @@ const generateFilteredTestcases = (): Testcase[] => {
             logger,
             llmClient,
             useTextExtract: shouldUseTextExtract,
+            modelName: input.modelName,
```
seanmcguire12, line 380: why do we need to pass in the modelName? is this bc agent wont work with external client providers like aiSDK etc?

evals/evaluator.ts
miguelg719: [f53fed6](https://github.com/browserbase/stagehand/pull/668/commits/f53fed61c3b7e008da6074e128521140f57e440d)

evals/index.eval.ts
miguelg719: It's because the modelName is defined inside the `agent` constructor, whereas LLM models are set in the Stagehand constructor. For example, in `google_flights.ts`:
```ts
export const google_flights: EvalFunction = async ({
  debugUrl,
  sessionUrl,
  stagehand,
  logger,
  modelName, // <-- This is needed here
}) => {
  await stagehand.page.goto("https://google.com/travel/flights");

  const agent = stagehand.agent({
    model: modelName, // <-- so that we can pass it here
...
```

evals/evaluator.ts
seanmcguire12, line 1: nit: should prob put these prompts & prompt building functions in `prompts.ts`, could do this in a fast follow

seanmcguire12 (approved):

evals/evaluator.ts
miguelg719: sg made a note for this

PR #669, opened 2025-04-15
default to stagehand LLM clients for evals

seanmcguire12:
# why
- we shouldn't use aiSDK (or any other external LLM clients) for evals by default, especially not in CI. We should be testing against our own custom clients
- we still want to be able to wrapped external clients so that we can easily get telemetry in braintrust
# what changed
- added a new CLI arg `useExternalClients` that defaults to false. if you want to run evals with external clients (to get telemetry, you can set `useExternalClients=true`
# test plan
- this is it


miguelg719 (approved):

kamath (approved):

kamath (changes requested):
accidentally approved, `-useExternalClient=true` isn't working for me

PR #678, opened 2025-04-18
set `force: true` for `locator.fill`

seanmcguire12:
# why
- we should still fill input boxes even if they do not pass playwrights built in actionability checks
- e.g., dropdown is expanded, and the element we want to fill is hidden underneath it
# what changed
- set `force: true` inside `locator.fill()`
- removed `keyboard.type()` since it is unlikely to make a difference for antibot
# test plan
- added an eval

### DIFF
--- a/.changeset/sweet-glasses-hope.md
+++ b/.changeset/sweet-glasses-hope.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+allow form filling when form is not top-most element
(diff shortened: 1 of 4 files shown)


miguelg719 (commented):
Is this working for cases where it's not topmost? I.e does force fill work different than force click

seanmcguire12:
> Is this working for cases where it's not topmost? I.e does force fill work different than force click

@miguelg719 good question. yeah force fill will just directly set the value of the element. ie, its not coordinate based. I also added an eval for this where the element to fill is covered by a dropdown

miguelg719 (approved):

PR #694, opened 2025-04-22
support telemetry for cua agent

seanmcguire12:
# why
- need to track tokens on cua agents
# what changed
- added token usage and inference time to `stagehand.metrics` for cua agent
# test plan
- tested locally by cross referencing with token metrics on openai dashboard


miguelg719 (approved):
Tested locally with a single agent and two agent instances

Issue #689, opened 2025-04-20
Use safeParse instead of parse for zod schemas to improve error reporting


jds2501:
The error reporting on zod parsing failures right now reports "Invalid response schema," which made debugging some stagehand test failures difficult. We can improve error handling by using a pattern like this:

const result = schema.safeParse(data);

if (!result.success) {
  // log or rethrow with detail
  throw new MyCustomError(result.error.format());
}

const parsed = result.data;

This would allow you to get more explicit guidance on why Zod schema parsing failed - example:

{
  name: { _errors: ['Required'] },
  age: { _errors: ['Expected number, received string'] },
  _errors: []
}

kamath:
Thanks for the feedback!! We can certainly take this on, unless someone else wants to handle this :)

CtrlAltGiri:
I'm happy to take this!

CtrlAltGiri:
Have a first iteration here: #690

CtrlAltGiri:
@kamath can you review this please?

PR #703, opened 2025-04-24
Issue[#689] Solved

akshatsrivastava11:
# why
Current error handling for Zod parsing failures logs only "Invalid response schema" without context, which makes debugging difficult — especially in automated test scenarios like stagehand. Developers have to dig deeper into the code or add temporary logs to identify the root issue.

# what changed
Replaced schema.parse(data) with schema.safeParse(data) to gracefully handle validation errors.

When parsing fails, the error is formatted using result.error.format() and thrown as a MyCustomError.

This provides a structured, detailed breakdown of validation issues for each field.

### DIFF
--- a/evals/args.ts
+++ b/evals/args.ts
@@ -83,13 +83,20 @@ if (parsedArgs.leftover.length > 0) {
       process.exit(1);
     }
     try {
-      EvalCategorySchema.parse(filterByCategory);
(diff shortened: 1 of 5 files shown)


seanmcguire12:
Hey thanks for your work here! We should have this covered in #741. Closing as duplicate.

PR #714, opened 2025-05-01
Allow AI SDK native clients to pass an API key

miguelg719:
# why
AI SDK will try to get the api key directly from `env` based on their convention, which disregards the constructor's `modelClientOptions`. This PR fixes that

# what changed
Added a conditional in `LLMProvider.ts` that checks if the api key is explicitly defined, and creates an instance of an AI SDK provider based on that api key. 

# test plan


lib/llm/LLMProvider.ts
```diff
@@ -17,19 +17,19 @@ import { GoogleClient } from "./GoogleClient";
...
+import { azure, createAzure } from "@ai-sdk/azure";
+import { groq, createGroq } from "@ai-sdk/groq";
+import { cerebras, createCerebras } from "@ai-sdk/cerebras";
+import { togetherai, createTogetherAI } from "@ai-sdk/togetherai";
+import { mistral, createMistral } from "@ai-sdk/mistral";
+import { deepseek, createDeepSeek } from "@ai-sdk/deepseek";
+import { perplexity, createPerplexity } from "@ai-sdk/perplexity";
 import { ollama } from "ollama-ai-provider";
-import { AISDKProvider } from "@/types/llm";
+import { AISDKProvider, AISDKCreator } from "@/types/llm";
```
miguelg719: happy to name creator differently, or maybe even override the defaults to conditionally load an api key

lib/llm/LLMProvider.ts
```diff
@@ -45,6 +45,19 @@ const AISDKProviders: Record<string, AISDKProvider> = {
...
+  google: createGoogleGenerativeAI,
+  xai: createXai,
+  azure: createAzure,
+  groq: createGroq,
+  cerebras: createCerebras,
+  togetherai: createTogetherAI,
+  mistral: createMistral,
+  deepseek: createDeepSeek,
+  perplexity: createPerplexity,
+};
```
miguelg719, line 60: if we override, then this is not needed. For example, AI SDK returns `openai`, `anthropic`, ... and so on, which are essentially constants returned from `createOpenAI`, `createAnthropic`,... in the same package. It is possible to define:
```ts
const openai = createOpenAI({})
```

lib/index.ts
```diff
@@ -667,16 +671,7 @@ export class Stagehand {
...
-        LLMProvider.getModelProvider(this.modelName) === "openai"
-          ? process.env.OPENAI_API_KEY || this.llmClient.clientOptions.apiKey
-          : LLMProvider.getModelProvider(this.modelName) === "anthropic"
-            ? process.env.ANTHROPIC_API_KEY ||
-              this.llmClient.clientOptions.apiKey
-            : LLMProvider.getModelProvider(this.modelName) === "google"
-              ? process.env.GOOGLE_API_KEY ||
-                this.llmClient.clientOptions.apiKey
-              : undefined;
+      const modelApiKey = this.modelClientOptions?.apiKey;
```
kamath, line 703: does this work if you don't explicitly pass in `apiKey` to `modelClientOptions`?

miguelg719: yes, but not for api

miguelg719: it does now

kamath (approved):

PR #731, opened 2025-05-06
return hybrid tree for `extract()` with no arguments

seanmcguire12:
# why
- migrating off of browser side/injected scripts
# what changed
- `extract()` with no arguments now returns the dom+a11y hybrid tree instead of a text rendered webpage
# test plan
- tested locally

### DIFF
--- a/.changeset/blue-beds-thank.md
+++ b/.changeset/blue-beds-thank.md
@@ -1,5 +1,5 @@
 ---
-"@browserbasehq/stagehand-lib": patch
+"@browserbasehq/stagehand-lib": minor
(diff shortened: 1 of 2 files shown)


.changeset/blue-beds-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+"@browserbasehq/stagehand-lib": patch
```
kamath:
```suggestion
"@browserbasehq/stagehand-lib": minor
```

kamath (approved):

PR #746, opened 2025-05-09
Docs: add link extraction to docs

seanmcguire12:
# why
- explain how to extract links 
# what changed
- added snippet explaining that you need to use `z.string().url()`

Screenshot 2025-05-09 at 1 26 05 PM

### DIFF
--- a/docs/reference/extract.mdx
+++ b/docs/reference/extract.mdx
@@ -6,8 +6,6 @@ icon: 'brain-circuit'
 
 `extract()` grabs structured text from the current page using [zod](https://github.com/colinhacks/zod). Given instructions and `schema`, you will receive structured data.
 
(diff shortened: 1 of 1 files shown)


docs/reference/extract.mdx
```diff
@@ -6,8 +6,6 @@ icon: 'brain-circuit'
 
 `extract()` grabs structured text from the current page using [zod](https://github.com/colinhacks/zod). Given instructions and `schema`, you will receive structured data.
 
-<Tip>We strongly suggest you set `useTextExtract` to `true` if you are extracting data from a longer body of text.</Tip>
```
kamath, line 9: need to update targeted extract reference as well, currently has text extract and says selector only works on text extract

kamath (approved):

PR #768, opened 2025-05-19
wait for dom settle with CDP network events

seanmcguire12:
# why
- the original browser side mutation observer implementation would cause this error: `page.evaluate: Execution context was destroyed, most likely because of a navigation`
   - the mutation observer had a timeout of 2 seconds, and if a navigation occurred during these two seconds, the `page.evaluate` would throw an error 
- we can avoid this by monitoring network events on the node side instead of relying on browser side scripts
# what changed
- removed browser side mutation observer logic in favour of listening to network events with CDP
# test plan
- evals


lib/StagehandPage.ts
```diff
@@ -427,71 +428,165 @@ ${scriptContent} \
...
+   *       `Page.frameStoppedLoading` fires the corresponding Document request is
+   *       removed immediately (covers iframes whose network events never close).
+   *   5.  A **stalled-request sweep timer** runs every 500 ms.  If a *Document*
+   *       request has been open for ≥ 2 s it is forcibly removed; this prevents
+   *       ad/analytics iframes from blocking the wait forever.
+   *   6.  When `inflight` becomes empty the helper starts a 500 ms timer.
+   *       If no new request appears before the timer fires, the promise
+   *       resolves → **DOM is considered settled**.
+   *   7.  A global guard (`timeoutMs` or `stagehand.domSettleTimeoutMs`,
+   *       default ≈ 30 s) ensures we always resolve; if it fires we log how many
```
miguelg719, line 457: note: lower the default timeout

miguelg719 (approved):

Issue #37, opened 2024-06-08
Support iframes


jeremypress:
Currently we don't process iframes, tarsier might have an approach that works for us here https://github.com/reworkd/tarsier/blob/ab12ad5fc303527b3247bded3b2ab5033a55133b/tarsier/tag_utils.ts#L178

neoromantic:
Tried to use Stagehand for form automations and immediately hit problems with iframes.

In many cases iFrames are used to embed various form fields, most importantly, credit card forms. 

While I can work around it with playwright directly, it seems to me that this feature is quite important.

Sudhamsha:
I'm working on automating forms on legacy insurer websites and immediately hit this issue! 

@neoromantic would you be able to share your workaround? Thanks!

lx3133584:
Is there any progress on this issue?

filip-michalsky:
I believe there is a fall back to computer use model now if an iframe is encountered

cmaycumber:
> I believe there is a fall back to computer use model now if an iframe is encountered

At least for me it's prohibitively expensive.

seanmcguire12:
completed in PR #778

PR #778, opened 2025-05-28
iframes

seanmcguire12:
## why
- addresses stagehand's longest standing issue: #37 
## What changed

This PR adds full support for interacting with nested `iframe`s in
Stagehand—across `extract`, `observe`, and `act` by:

- **Building a combined accessibility tree** (and URL/XPath mappings)
spanning the main document and any nested iframes.
- **Tracking frame‑scoped element IDs** so that element identifiers are
globally unique across frames. This is done because `backendNodeId`'s are not guaranteed to be unique across OOPIF's (out of process iframes).
- **Extending CDP session management** to correctly target Out‑Of‑Process
iframes (OOPIFs) and fallback to same‑process frames.
- **Introducing a “deep” XPath‑based locator** that can step into `<iframe>`
 elements when performing Playwright actions.
- **Updating Zod schema transforms** to expect and handle the new
“frameId-backendId” string format for element IDs.
- **Adding new error types** for improved diagnostics when iframe resolution
 or XPath lookups fail.
- **Updating internal types** (`context.ts`, `stagehand.ts`,
`stagehandErrors.ts`) and utility modules (`a11y/utils.ts`, `utils.ts`,
handlers) to accommodate frame‑aware operations.
- **Adding three end‑to‑end eval tasks** testing iframe support:
    - `iframe_hn` (extract)
    - `iframe_same_proc` (act)
    - `iframe_form_filling` (act).

### 1. Frame‑Scoped Element IDs (EncodedId)

- **lib/StagehandPage.ts**
   -  Introduced `encodeWithFrameId(…)`, `ordinalForFrameId(…)`, and
`resetFrameOrdinals()` to assign and track per‑frame ordinals.
   - CDP session caching moved to a `WeakMap<Page|Frame, CDPSession>` so we
 can open sessions against arbitrary frames.
- **types/context.ts**
   - Defined ``EncodedId = ${number}-${number}`` for
“frameOrdinal-backendNodeId” IDs.
   - Updated `TreeResult` to key `xpathMap`/`idToUrl` by `EncodedId`.

### 2. Combined Accessibility Tree Across Frames

- **lib/a11y/utils.ts**
   - Added `getAccessibilityTreeWithFrames()` which walks the CDP frame
tree, captures accessibility sub‑trees for each frame, and concatenates them
into a single “combinedTree” string plus combined URL/XPath maps keyed by
`EncodedId`.
   - Updated `formatSimplifiedTree()` to emit the new `encodedId` in tree
lines.
   - Updated `buildBackendIdMaps()` to traverse nested frame DOM nodes
(OOPIF and same‑process iframes) and include the frame’s `frameId` when encoding
 backend IDs.

### 3. Deep XPath Locator for Frame Actions

- **lib/handlers/handlerUtils/actHandlerUtils.ts**
      Added `deepLocator(root, rawXPath)` which splits an XPath on `<iframe>`
steps to descend into `FrameLocator`s automatically before applying the
remainder of the path.
- **lib/handlers/actHandler.ts**
      Uses `deepLocator()` instead of a flat `page.locator(...)` so that
Playwright actions can target elements inside iframes when `options.iframes` is
set.

### 4. Frame‑Aware Extract & Observe Handlers

- **lib/handlers/extractHandler.ts**
   - Imports and uses `getAccessibilityTreeWithFrames()` when `iframes:
true`; otherwise falls back to the legacy single‑frame tree.
   - Passes through the `iframes` flag into its internal calls.
- **lib/handlers/observeHandler.ts**
   - Similarly leverages `getAccessibilityTreeWithFrames()` and builds its
element‑to‑XPath mapping from the combined tree.
   - Removes the old ad‑hoc iframe injection logic in favor of the unified
combined tree approach.

### 5. Inference Schema & Element ID Changes

- **lib/inference.ts**
   - Changed the Observe LLM output schema so `elementId` is now a
**string** matching the regex `/^\d+-\d+$/` (frame‑ID plus backend‑ID) instead
of a raw number.
   - Updated parsing logic to no longer coerce to Number.

### 6. Zod URL‑Field Transformation

- **lib/utils.ts**
   - Renamed `makeIdNumberSchema()` → `makeIdStringSchema()` to emit
`z.string().regex(/^\d+-\d+$/)` for fields that were formerly `string().url()`,
so extracted URL placeholders match the new `EncodedId` format.
   - Updated `injectUrls()` to map both numeric and frame‑ID strings back
into real URLs once extraction is complete.

to do:
- [x] clean up code
- [x] add JSDoc
- [x] add comprehensive PR desc
- [x] parameterize iframe traversal. ie, need to set `iframes: true`
- [ ] ignore iframes without content
- [x] add evals

### DIFF
--- a/lib/StagehandPage.ts
+++ b/lib/StagehandPage.ts
@@ -20,7 +20,7 @@ import { StagehandObserveHandler } from "./handlers/observeHandler";
 import { ActOptions, ActResult, GotoOptions, Stagehand } from "./index";
 import { LLMClient } from "./llm/LLMClient";
 import { StagehandContext } from "./StagehandContext";
-import { EnhancedContext } from "../types/context";
+import { EncodedId, EnhancedContext } from "../types/context";
 import { clearOverlays } from "./utils";
 import {
   StagehandError,
@@ -54,6 +54,9 @@ export class StagehandPage {
     PlaywrightPage | Frame,
     CDPSession
   >();
+  private fidOrdinals: Map<string | undefined, number> = new Map([
+    [undefined, 0],
+  ]);
 
   constructor(
     page: PlaywrightPage,
@@ -120,6 +123,29 @@ export class StagehandPage {
         userProvidedInstructions,
       });
     }
+  }
+
+  public ordinalForFrameId(fid: string | undefined): number {
+    if (fid === undefined) return 0;
+
+    const cached = this.fidOrdinals.get(fid);
+    if (cached !== undefined) return cached;
+
+    const next = this.fidOrdinals.size; // 1, 2, 3, …
+    if (next > 99) throw new Error("Too many frames – widen EncodedId format");
+    this.fidOrdinals.set(fid, next);
+    return next;
+  }
+
+  public encodeWithFrameId(
+    fid: string | undefined,
+    backendId: number,
+  ): EncodedId {
+    return `${this.ordinalForFrameId(fid)}-${backendId}` as EncodedId;
+  }
+
+  public resetFrameOrdinals(): void {
+    this.fidOrdinals = new Map([[undefined, 0]]);
   }
 
   private async ensureStagehandScript(): Promise<void> {
--- a/lib/StagehandPage.ts
+++ b/lib/StagehandPage.ts
@@ -31,6 +31,8 @@ import {
   MissingLLMConfigurationError,
   HandlerNotInitializedError,
   StagehandDefaultError,
+  ExperimentalApiConflictError,
+  ExperimentalNotConfiguredError,
 } from "../types/stagehandErrors";
 import { StagehandAPIError } from "@/types/stagehandApiErrors";
 import { scriptContent } from "@/lib/dom/build/scriptContent";
@@ -67,6 +69,9 @@ export class StagehandPage {
     api?: StagehandAPI,
     waitForCaptchaSolves?: boolean,
   ) {
+    if (stagehand.experimental && api) {
+      throw new ExperimentalApiConflictError();
+    }
     this.rawPage = page;
     // Create a proxy to intercept all method calls and property access
     this.intPage = new Proxy(page, {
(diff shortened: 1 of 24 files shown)


types/context.ts
```diff
@@ -53,15 +56,70 @@ export type DOMNode = {
...
+  parentFrame?: Frame;
+  /** CDP frame identifier for this snapshot; used to generate stable EncodedIds. */
+  frameId?: string;
+}
+
+export type EncodedId = `${number}-${number}`;
+
+export interface RichNode extends AccessibilityNode {
+  encodedId?: EncodedId;
+}
```
miguelg719, line 125: should we move these types to page types

lib/utils.ts
```diff
@@ -419,13 +424,17 @@ function makeIdNumberSchema(orig: z.ZodString): z.ZodNumber {
...
+    '(e.g. "0-432").';
   const composed =
     userDesc.trim().length > 0
       ? `${base} that follows this user-defined description: ${userDesc}`
       : base;
 
-  return z.number().describe(composed);
+  return z
+    .string()
+    .regex(/^\d+-\d+$/)
```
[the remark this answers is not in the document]
miguelg719: I like this comment

lib/StagehandPage.ts
```diff
@@ -16,7 +20,7 @@ import { StagehandObserveHandler } from "./handlers/observeHandler";
 import { ActOptions, ActResult, GotoOptions, Stagehand } from "./index";
 import { LLMClient } from "./llm/LLMClient";
 import { StagehandContext } from "./StagehandContext";
-import { EnhancedContext } from "../types/context";
+import { EncodedId, EnhancedContext } from "../types/context";
```
miguelg719, line 23: if we move to page types we just pull from there

lib/utils.ts
```diff
@@ -396,9 +394,16 @@ export function injectUrls(
...
     if (path.length === 1) {
       const fieldValue = record[key];
-      if (typeof fieldValue === "number") {
-        const mappedUrl = idToUrlMapping[String(fieldValue)];
-        record[key] = mappedUrl ?? ``;
+
+      const id =
+        typeof fieldValue === "number"
+          ? String(fieldValue)
+          : typeof fieldValue === "string" && /^\d+-\d+$/.test(fieldValue)
```
miguelg719: could replace the regex with the constant

miguelg719:
working through it, another note before approving is adding this behind `experimental:true`

PR #786, opened 2025-05-30
handle reroute for rollout

miguelg719:
# why

# what changed

# test plan

### DIFF
--- a/.changeset/rude-lions-call.md
+++ b/.changeset/rude-lions-call.md
@@ -1,0 +1,6 @@
+---
+"@browserbasehq/stagehand-lib": patch
+"@browserbasehq/stagehand": patch
+---
+
+Handle reroute to account for rollout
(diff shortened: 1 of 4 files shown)


.changeset/rude-lions-call.md
```diff
@@ -0,0 +1,6 @@
+---
+"@browserbasehq/stagehand-lib": patch
```
kamath:
```suggestion
```

types/api.ts
```diff
@@ -29,6 +29,7 @@ export interface StartSessionParams {
 
 export interface StartSessionResult {
   sessionId: string;
+  unavailable?: boolean;
```
kamath: this is a little unintuitive -- having a true/false field for `apiEnabled` from the API would be cleaner than a true/undefined `unavailable` field

.changeset/rude-lions-call.md
miguelg719: without it changes wouldn't publish to npm

PR #796, opened 2025-06-04
Experimental flag for new features

miguelg719:
# why

# what changed

# test plan

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -399,6 +399,7 @@ export class Stagehand {
   private _browser: Browser | undefined;
   private _isClosed: boolean = false;
   private _history: Array<HistoryEntry> = [];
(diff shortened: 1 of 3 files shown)


lib/index.ts
```diff
@@ -399,6 +399,7 @@ export class Stagehand {
   private _browser: Browser | undefined;
   private _isClosed: boolean = false;
   private _history: Array<HistoryEntry> = [];
+  public readonly experimental: boolean; // could be also made private?
```
miguelg719: depends on preference

seanmcguire12: delete this comment

miguelg719:
```suggestion
  public readonly experimental: boolean;
```

PR #801, opened 2025-06-10
default use API true when using browserbase

miguelg719:
# why

# what changed

# test plan

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -637,7 +637,7 @@ export class Stagehand {
     this.experimental = experimental;
     if (this.experimental) {
       this.stagehandLogger.warn(
-        "Experimental mode is enabled. This is a beta feature and may break at any time.",
+        "Experimental mode is enabled. This is a beta feature and may break at any time. Enabling experimental mode will disable the API",
       );
       // Disable API mode in experimental mode
       this.usingAPI = false;


lib/index.ts
```diff
@@ -640,6 +635,13 @@ export class Stagehand {
     this.selfHeal = selfHeal;
     this.disablePino = disablePino;
     this.experimental = experimental;
+    if (this.experimental) {
+      this.stagehandLogger.warn(
+        "Experimental mode is enabled. This is a beta feature and may break at any time.",
+      );
```
miguelg719, line 641: indicate that api will be false in the warning

evals/test-results/.last-run.json
```diff
@@ -0,0 +1,4 @@
+{
```
kamath: do we need?

evals/deterministic/tests/browserbase/sessions.test.ts
```diff
@@ -45,10 +45,11 @@ test.describe("Browserbase Sessions", () => {
     expect(page.url()).toBe(
       "https://docs.stagehand.dev/get_started/introduction",
     );
-    await stagehand.close();
   });
   test("resumes a session via CDP URL", async () => {
     const session = await browserbase.sessions.retrieve(sessionId);
+    console.log(session);
+    console.log(session.connectUrl);
```
kamath: use expect here

lib/api.ts
```diff
@@ -58,6 +58,13 @@ export class StagehandAPI {
       throw new StagehandAPIError("modelApiKey is required");
     }
     this.modelApiKey = modelApiKey;
+
+    const region = browserbaseSessionCreateParams?.region;
+    if (region && region !== "us-west-2") {
+      throw new StagehandAPIError(
+        "Only us-west-2 region is currently supported for API mode. Try updating your region or set useAPI to false.",
```
kamath: this shouldn't throw error

miguelg719: [08f1396](https://github.com/browserbase/stagehand/pull/801/commits/08f139628c58fde182fa767c273630ad364b0a25)

evals/deterministic/tests/browserbase/sessions.test.ts
miguelg719: [08f1396](https://github.com/browserbase/stagehand/pull/801/commits/08f139628c58fde182fa767c273630ad364b0a25)

evals/test-results/.last-run.json
miguelg719: [08f1396](https://github.com/browserbase/stagehand/pull/801/commits/08f139628c58fde182fa767c273630ad364b0a25)

lib/api.ts
seanmcguire12: @kamath wouldn't it be better to throw here? i personally would not want to spend time guessing why code is slow. the only problem i could see is if you for some reason we're unable to set it to `us-west-2`

kamath (approved):

PR #803, opened 2025-06-12
Enable session affinity for cache optimization

miguelg719:
# why

# what changed

# test plan

### DIFF
--- a/lib/api.ts
+++ b/lib/api.ts
@@ -27,6 +27,7 @@ import {
   StagehandResponseBodyError,
   StagehandResponseParseError,
 } from "../types/stagehandApiErrors";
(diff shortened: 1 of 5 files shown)


lib/api.ts
```diff
@@ -58,6 +62,11 @@ export class StagehandAPI {
       throw new StagehandAPIError("modelApiKey is required");
     }
     this.modelApiKey = modelApiKey;
+
+    const region = browserbaseSessionCreateParams?.region;
+    if (region && region !== "us-west-2") {
+      return { sessionId: browserbaseSessionID ?? null, available: false };
+    }
```
miguelg719: this might've transpiled from #801

PR #809, opened 2025-06-16
better logging for `AI_NoObjectGeneratedError`

seanmcguire12:
# why
- AI SDK exposes useful debugging info for `AI_NoObjectGeneratedError`. Link to docs [here](https://ai-sdk.dev/docs/reference/ai-sdk-errors/ai-no-object-generated-error#checking-for-this-error)
- we should expose this info in our logs
# what changed
- wrapped the `generateObject` call in a try/catch
- if we get the `AI_NoObjectGeneratedError`, we log additional debugging info about the error, such as:
     - error.cause
     - error.response
     - error.usage
     - error.finishReason
     - error.text
# test plan
- regressions

### DIFF
--- a/lib/llm/aisdk.ts
+++ b/lib/llm/aisdk.ts
@@ -8,6 +8,7 @@ import {
   generateText,
   ImagePart,
   LanguageModel,
+  NoObjectGeneratedError,
   TextPart,
 } from "ai";
 import { CreateChatCompletionOptions, LLMClient } from "./LLMClient";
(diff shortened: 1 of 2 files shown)


lib/llm/aisdk.ts
```diff
@@ -156,19 +157,59 @@ export class AISdkClient extends LLMClient {
...
+          messages: formattedMessages,
+          schema: options.response_model.schema,
+        });
+      } catch (err) {
+        if (NoObjectGeneratedError.isInstance(err)) {
+          this.logger?.({
+            category: "AISDK error",
+            message: err.message,
+            level: 0,
+            auxiliary: {
```
kamath, line 174: can just do `{ value: JSON.stringify(err), type: "object" }`

kamath (approved):

PR #819, opened 2025-06-20
try playwright click and fall back to JS click

### LINKED ISSUE #795: page.act on an ObserveResult with click method does not replicate page.click
From this:

```
const [action] = await page.observe(step);
```

I get an `ObserveResult`

```
{
  description: "A menuitem button labeled 'Rewards' in the menubar, which can be clicked to navigate to the Rewards section.",
  method: "click",
  selector: "xpath=/html/body[1]/div[1]/div[1]/div[1]/div[2]/button[6]"
}
```

Doing `await page.act(action);` does not click the button.

But it worked as expected when I tested using `await page.click(action.selector)`. Is `page.act()` doing some conversion or processing on the selector?

seanmcguire12:
# why
- this should fix #795 
- JS click events can be blocked by website owners
- we need to try with a coordinate click (with playwright) then retry with a JS click event
- ideally, we would try with a JS click first (because it is faster), but there is no way to know if a JS click has failed 
# what changed
- updated the `clickElement` function to start with a regular playwright click and fallback to a JS click 
- the playwright click has a 3.5 second timeout so that we fail fast
# test plan
- `act`, `regression`, and `combination` evals
- added a new eval: `no_js_click.ts` which simulates a website that blocks JS click events

### DIFF
--- a/lib/handlers/handlerUtils/actHandlerUtils.ts
+++ b/lib/handlers/handlerUtils/actHandlerUtils.ts
@@ -372,10 +372,8 @@ export async function clickElement(ctx: MethodHandlerContext) {
     },
   });
 
-  let clickSucceeded = false;
   try {
     await locator.click({ timeout: 3_500 });
-    clickSucceeded = true;
   } catch (e) {
     logger({
       category: "action",
@@ -389,9 +387,7 @@ export async function clickElement(ctx: MethodHandlerContext) {
         args: { value: JSON.stringify(args), type: "object" },
       },
     });
-  }
 
-  if (!clickSucceeded) {
     try {
       await locator.evaluate((el) => (el as HTMLElement).click());
     } catch (e) {


lib/handlers/handlerUtils/actHandlerUtils.ts
```diff
@@ -372,24 +372,43 @@ export async function clickElement(ctx: MethodHandlerContext) {
...
+        auxiliary: {
+          error: { value: (e).message, type: "string" },
+          trace: { value: (e).stack, type: "string" },
+          xpath: { value: xpath, type: "string" },
+          method: { value: "click", type: "string" },
+          args: { value: JSON.stringify(args), type: "object" },
+        },
+      });
+      throw new StagehandClickError(xpath, (e).message);
+    }
```
miguelg719, line 407: should we bring this into the catch block instead and remove the clickSucceeded bool?

seanmcguire12: yeah good call

miguelg719 (approved):

Issue #794, opened 2025-06-03
(Local browser session reuse) No option to persist Stagehand / Playwright userDataDir


camille-readbean:
It seems that [`stagehand.close()`](https://github.com/browserbase/stagehand/blob/2ec8e5a0f9cfc3aca153b805bed1988dfe5dec2e/lib/index.ts#L795-L827) deletes the userDataDir.  

```ts
    if (this.contextPath) {
      try {
        fs.rmSync(this.contextPath, { recursive: true, force: true });
      } catch (e) {
        console.error("Error deleting context directory:", e);
      }
    }
```
  
I was led to the impression that setting userDataDir will allow me to be able to store cookies etc, i.e. letting me reuse the session and context of the browser. However it seems calling `stagehand.close()` will wipe and delete the data in userDataDir.

It would be good if theres a flag to set persistence, or other mechanism to allow storage.  
  
A very important use case is to authenticate once and reuse the sessions in future workflows, saving time and resources spent on authentication.

c-reiter:
Bump, as I have the same issue. Providing a `userDataDir` in `localBrowserLaunchOptions` somehow doesn't forward the property to Chromium and still uses the `.browser_profile` directory which gets deleted after `stagehand.close()`

kushagrob:
I am also getting the same issue

miguelg719:
hey @kushagrob @c-reiter @camille-readbean ! This just shipped with the new release of stagehand! Let us know if it works for you

PR #825, opened 2025-06-23
Persist userDataDir if specified in localBrowserLaunchOptions

kushagrob:
# why

As per https://github.com/browserbase/stagehand/issues/794, the local userDataDir automatically gets deleted after `stagehand.close()`, making it difficult to reuse the Stagehand browser context locally. 

# what changed

Change `stagehand.close()`so it only deletes the profile folder when Stagehand itself created a temp one. Now a user-supplied userDataDir survives `stagehand.close()`, while temp dirs are still cleaned up.

Let me know if you want to make it a flag instead. 

# test plan

Added to a current test case and ran <code>npm run e2e:local -- tests/local/create.test.ts</code> as well as <code>npm run e2e</code>

### DIFF
--- a/evals/deterministic/tests/local/create.test.ts
+++ b/evals/deterministic/tests/local/create.test.ts
@@ -26,6 +26,7 @@ test.describe("Local browser launch options", () => {
       localBrowserLaunchOptions: {
         userDataDir: customUserDataDir,
         headless: true,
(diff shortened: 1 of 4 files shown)


the-roaring:
we don't want this to be default behavior so I added it as a flag

Issue #829, opened 2025-06-24
_refreshPageFromAPI invalidly selecting wrong API key


ZacharyHampton:
When initializing a stagehand session, it properly authenticates with Browserbase via the given API key. However, when calling `_refreshPageFromAPI`, it defaults to using `BROWSERBASE_API_KEY`, not the predefined key given via the stagehand config.

We need to use multiple api keys in our project, so we cannot define to the environment variable, rather explicitly to the config. This creates an error when stagehand calls to `_refreshPageFromAPI`, as it is not defined.

PR #830, opened 2025-06-24
_refreshPageFromAPI auth fix

ZacharyHampton:
# why
#829

# what changed
Conditionally added config defined api key when the env variable is false

# test plan


lib/StagehandPage.ts
```diff
@@ -192,7 +192,7 @@ ${scriptContent} \
     }
 
     const browserbase = new Browserbase({
-      apiKey: process.env.BROWSERBASE_API_KEY,
+      apiKey: process.env.BROWSERBASE_API_KEY || this.stagehand.apiKey,
```
miguelg719: Let's prioritize the explicitly defined apikey
```suggestion
      apiKey: this.stagehand.apiKey ?? process.env.BROWSERBASE_API_KEY ,
```

miguelg719 (changes requested):

ZacharyHampton:
from not being so in the weeds of the codebase, im not so sure what the best practice is to pass the apiKey through (build failing due to the field we are trying to access being private)

if you dont mind ill leave it up to you @miguelg719

miguelg719:
got it, fixing now thanks for the heads up!

PR #836, opened 2025-06-25
Optionally persist userDataDir with preserveUserDataDir flag

the-roaring:
Closes (#825)

# why
some users want to preserve the userDataDir between runs in local (per https://github.com/browserbase/stagehand/issues/794 and https://github.com/browserbase/stagehand/pull/825)

# what changed
- new flag in localBrowserLaunchOptions called preserveUserDataDir

# test plan
- added relevant test cases in tests/local/create.test.ts


miguelg719 (approved):
approved pending docs + changeset

PR #849, opened 2025-06-28
[chore] update cua models for evals

seanmcguire12:
# why
- date stamped openai model is more performant
- date stamped claude model was not working

### DIFF
--- a/evals/taskConfig.ts
+++ b/evals/taskConfig.ts
@@ -102,7 +102,7 @@ const DEFAULT_EVAL_MODELS = process.env.EVAL_MODELS
 
 const DEFAULT_AGENT_MODELS = process.env.EVAL_AGENT_MODELS
   ? process.env.EVAL_AGENT_MODELS.split(",")
-  : ["computer-use-preview", "claude-3-7-sonnet-20250219"];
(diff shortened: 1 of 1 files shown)


ctfrob:
Thanks for opening this. I’ve been hitting the same model-name mismatch on the cloud service:

* `computer-use-preview-2025-03-11` → “model not supported”
* `claude-3-7-sonnet-latest` → “model not whitelisted”

Locally (`env: "LOCAL"`) both names work, so the block is coming from the server-side allow-list. Updating the eval config is a great first step, but could you also add these aliases to the runtime schema (`packages/core/src/llm/*`) and redeploy the Browserbase validator? Right now any call that includes the new OpenAI project keys or the latest Claude alias dies before it reaches the LLM provider.

It's basically giving me a whack-a-mole response right now - listing model names stochastically (3-4 variations with different date suffixes, etc.)

Happy to provide logs or a minimal repro if that helps. Appreciate the quick attention to this.

miguelg719 (approved):

seanmcguire12:
@ctfrob thanks for the feedback! We'll make sure to include the model updates in the next api deployment. Aiming for early this week

PR #856, opened 2025-06-30
set default download behaviour

seanmcguire12:
# why
- users do not want to have to send CDP commands to set this manually
- we can handle this by:
   - setting the default downloads path to "downloads" 
   - allowing this to be overwritten if the user defines a downloads directory in `localBrowserLaunchOptions
# what changed
- added an `enableDownloads` const inside `init` to set the default download behaviour
# test plan
- still need to add tests

### DIFF
--- a/.changeset/eighty-papers-shout.md
+++ b/.changeset/eighty-papers-shout.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+set download behaviour by default


lib/index.ts
```diff
@@ -789,6 +789,47 @@ export class Stagehand {
...
+
+        const applyToPage = async (page: Page): Promise<void> => {
+          const pageSession: CDPSession =
+            await this.context.newCDPSession(page);
+          await pageSession.send("Page.setDownloadBehavior", {
+            behavior: "allow",
+            downloadPath: downloadsDir,
+          });
+        };
+
```
miguelg719: I think we should add this to the page class wdyt

lib/index.ts
```diff
@@ -789,6 +789,47 @@ export class Stagehand {
...
+        if (this._browser) {
+          const browserSession: CDPSession =
+            await this._browser.newBrowserCDPSession();
+          await browserSession.send("Browser.setDownloadBehavior", {
+            behavior: "allow",
+            downloadPath: downloadsDir,
+            eventsEnabled: true,
+          });
+          return;
+        }
```
miguelg719: Is this a problem with env local? Ideally we can feed this into the getBrowser function `env:BROWSERBASE` and eventually move it out to its own file like we did in python

lib/index.ts
seanmcguire12: did it this way because `localBrowserLaunchOptions` is private, ie: not accessible from the `StagehandPage` class

lib/index.ts
seanmcguire12: yeah I couldn't get it to work with local dev

lib/index.ts
miguelg719: we can make it `protected`

miguelg719 (approved):

PR #864, opened 2025-07-09
temporarily patch custom defined llm clients on api

miguelg719:
# why

# what changed

# test plan

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -614,6 +614,13 @@ export class Stagehand {
 
     if (llmClient) {
       this.llmClient = llmClient;
+      this.logger({
+        category: "init",
+        message:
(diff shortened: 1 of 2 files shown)


lib/index.ts
```diff
@@ -614,6 +614,13 @@ export class Stagehand {
 
     if (llmClient) {
       this.llmClient = llmClient;
+      this.logger({
+        category: "init",
+        message:
+          "Custom LLM clients are currently not supported in API mode, share your feedback in slack! https://stagehand.dev/slack",
```
miguelg719:
```suggestion
          "Custom LLM clients are currently not supported in API mode",
```

seanmcguire12 (approved):

PR #867, opened 2025-07-10
add new models

sameelarif:
# why
we need to add support for new computer use models

# what changed
added to the list of supported models for CUA. also updated the legacy model lists until we fully migrate to AI SDK

# test plan

### DIFF
--- a/lib/agent/AnthropicCUAClient.ts
+++ b/lib/agent/AnthropicCUAClient.ts
@@ -1,4 +1,4 @@
-import Anthropic from "@anthropic-ai/sdk";
+import Anthropic, { ClientOptions } from "@anthropic-ai/sdk";
 import { LogLine } from "@/types/log";
 import {
   AgentAction,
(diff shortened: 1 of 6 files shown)


lib/agent/AnthropicCUAClient.ts
```diff
@@ -44,21 +44,13 @@ export class AnthropicCUAClient extends AgentClient {
...
     this.baseURL = (clientOptions?.baseURL as string) || undefined;
 
-    // Get thinking budget if specified
-    if (
-      clientOptions?.thinkingBudget &&
-      typeof clientOptions.thinkingBudget === "number"
-    ) {
-      this.thinkingBudget = clientOptions.thinkingBudget;
-    }
-
```
sameelarif, line 54: removed since the key did not exist on the client options type

lib/agent/OpenAICUAClient.ts
```diff
@@ -43,15 +43,21 @@ export class OpenAICUAClient extends AgentClient {
...
     this.organization =
       (clientOptions?.organization as string) || process.env.OPENAI_ORG;
+    this.baseURL = (clientOptions?.baseURL as string) || undefined;
 
     // Get environment if specified
-    if (
-      clientOptions?.environment &&
-      typeof clientOptions.environment === "string"
-    ) {
-      this.environment = clientOptions.environment;
```
sameelarif, line 52: removed since the environment key did not exist on the client options type

PR #872, opened 2025-07-12
[feat]: adding openrouter as a client to run evals on grok 4

Kylejeong2:
# why

to run grok 4 in stagehand

# what changed

added openrouter client to run grok before the api comes out 

# test plan

running evals


Kylejeong2:
Openrouter is provided in the AI-sdk so this isn't needed.

PR #876, opened 2025-07-14
[chore]: run a full stack eval every time there is a minor release

Kylejeong2:
# why
Evaluating stagehand on all the evals to have the most up to date evaluation information for stagehand.dev/evals. This makes it so everytime there is a minor release, it automatically runs the full evaluation suite. 

Can be changed to the major release? or "odd"/"even" minor releases. 

# what changed

if there aren't any modifiers to the eval, then it is a full stack eval and will be named as such (for later retrieval) 

# test plan

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,6 +15,9 @@ jobs:
   release:
     name: Release
     runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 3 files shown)


Kylejeong2:
Needs to be fully rethought and planned.

PR #880, opened 2025-07-14
fixed deepLocator() pathing bug

Mochael:
# why
The logic in deepLocator was consolidating all the double // -> single /s, meaning that if I tried to do any sort of XPATH without specifying direct descendants, the query would break. So for example when I set my selector to `selector: xpath=//*[@id='myID'] , it turns the locator into 'xpath=/*[@id=\'myID\']' removing the 2 slashes. This breaks my search because it's only looking for the ID in the root element, not across the descendants of the whole document (which is what I wanted)

# what changed
I changed the logic so that it preserves the number of slashes in the xpath instead of splitting on "/", dropping all the slashes and then re-adding them

# test plan
I tested locally but the stagehand team should verify that this works on whatever suite of tests they typically try. They should also probably add a test to catch this error like a search on non-direct descendants.

### DIFF
--- a/lib/handlers/handlerUtils/actHandlerUtils.ts
+++ b/lib/handlers/handlerUtils/actHandlerUtils.ts
@@ -9,37 +9,52 @@ import { StagehandClickError } from "@/types/stagehandErrors";
 const IFRAME_STEP_RE = /^iframe(\[[^\]]+])?$/i;
 
 export function deepLocator(
-  root: Page | FrameLocator,
-  rawXPath: string,
-): Locator {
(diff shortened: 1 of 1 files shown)


m3kessler1:
Ran into this same bug, this is a big deal for repeatability.   Have to use full xpath to items, can't reference by id.  Came to see if someone else resolved.   Please merge this in asap.

seanmcguire12:
closing as this was fixed in stagehand v3, both with [locator](https://docs.stagehand.dev/v3/references/locator) & [deepLocator](https://docs.stagehand.dev/v3/references/deeplocator)

PR #897, opened 2025-07-19
[fix] ensure selfHeal respects arguments

miguelg719:
# why
On self healing we were finding the selector, but nor remembering the arguments previously passed; so either they were empty or the LLM was hallucinating them. This PR fixes this

# what changed
Updated the logic from selfHealing to reuse the previously passed arguments and method

# test plan
- [x] Added evals to act and regression (`evals/tasks/heal_*.ts`)

### DIFF
--- a/lib/handlers/actHandler.ts
+++ b/lib/handlers/actHandler.ts
@@ -307,9 +307,10 @@ export class StagehandActHandler {
   private async _performPlaywrightMethod(
     method: string,
     args: unknown[],
-    xpath: string,
+    rawXPath: string,
     domSettleTimeoutMs?: number,
   ) {
+    const xpath = rawXPath.replace(/^xpath=/i, "").trim();
     const locator = deepLocator(this.stagehandPage.page, xpath).first();
     const initialUrl = this.stagehandPage.page.url();
 


evals/initStagehand.ts
```diff
@@ -48,6 +48,7 @@ const StagehandConfig = {
       },
     },
   },
+  selfHeal: true,
```
miguelg719, line 51: @seanmcguire12 👀, 👍 / 👎 ?

evals/tasks/heal_scroll_50.ts
```diff
@@ -0,0 +1,60 @@
...
+  try {
+    await stagehand.page.goto(
+      "https://browserbase.github.io/stagehand-eval-sites/sites/aigrant/",
+    );
+    await stagehand.page.act({
+      description: "Scroll 50% down the page",
+      selector: "/html/body/div/div/button",
+      arguments: ["50"],
+      method: "scrollTo",
+    });
```
miguelg719: does this make sense to self heal at all?

lib/handlers/actHandler.ts
```diff
@@ -147,10 +147,35 @@ export class StagehandActHandler {
...
+          return {
+            success: false,
+            message: `Failed to self heal act: No observe results found for action`,
+            action: actCommand,
+          };
+        }
+        const element: ObserveResult = observeResults[0];
+        await this._performPlaywrightMethod(
+          // override previously provided method and arguments
+          observe.method,
```
miguelg719, line 168: this will also be overriden

PR #905, opened 2025-07-22
removed all images except the latest two in AnthropicCuaClient

tkattkat:
# why
part of STG-586

Currently, we leave all of the images for anthropic cua client within the LLM's context as the task progresses

# what changed

We now remove all screenshots aside from the last two when experimental flag is set to true within stagehand config 

# test plan

tested locally

### DIFF
--- a/lib/agent/AnthropicCUAClient.ts
+++ b/lib/agent/AnthropicCUAClient.ts
@@ -13,6 +13,7 @@ import {
 } from "@/types/agent";
 import { AgentClient } from "./AgentClient";
 import { AgentScreenshotProviderError } from "@/types/stagehandErrors";
+import { compressConversationImages } from "./imageCompressionUtils";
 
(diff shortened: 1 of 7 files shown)


lib/agent/AnthropicCUAClient.ts
```diff
@@ -30,12 +31,14 @@ export class AnthropicCUAClient extends AgentClient {
...
   private actionHandler?: (action: AgentAction) => Promise<void>;
   private thinkingBudget: number | null = null;
+  private experimental: boolean = false;
 
   constructor(
     type: AgentType,
     modelName: string,
     userProvidedInstructions?: string,
     clientOptions?: Record<string, unknown>,
+    experimental?: boolean,
```
miguelg719, line 41: does the openai client also need this in the constructor if it's passed through?

PR #907, opened 2025-07-23
[api] store frame ID map

seanmcguire12:
# why
- to enable multi-tab handling on the Stagehand API
# what changed
- added a mapping of CDP Frame ID -> Page objects
   - frame IDs are now included in the payload for API request so that the API knows which page to take the action on
# test plan
- `act` evals
- `regression` evals
- `extract` evals 
- `observe` evals


types/stagehand.ts
```diff
@@ -152,6 +154,7 @@ export interface ObserveOptions {
   onlyVisible?: boolean;
   drawOverlay?: boolean;
   iframes?: boolean;
+  frameId?: string;
```
miguelg719, line 157: should we add to agent as well?

lib/api.ts
```diff
@@ -123,7 +123,7 @@ export class StagehandAPI {
     if (!options.schema) {
       return this.execute<ExtractResult<T>>({
         method: "extract",
-        args: {},
+        args: { ...options },
```
miguelg719, line 126: this might break extract no params?

PR #910, opened 2025-07-23
improve cua key mapping logic

tkattkat:
# why

Currently we have a handful of keys that fail to be used, and redundant logic around key handling. It is now consolidated to one location, and properly handles the failing keys 
part of STG-587

# what changed

consolidated key handling for CUA agents to one util 

# test plan

tested locally

### DIFF
--- a/.changeset/huge-suns-dress.md
+++ b/.changeset/huge-suns-dress.md
@@ -1,5 +1,0 @@
----
-"@browserbasehq/stagehand": patch
----
-
-Improve cua key mapping logic


.changeset/huge-suns-dress.md
miguelg719, line 1: can we delete this changeset

tkattkat: yea

tkattkat: deleted

miguelg719 (approved):

PR #913, opened 2025-07-24
don't require `experimental: true` for iframes

seanmcguire12:
# why
- to enable setting the `iframes` parameter without setting `experimental: true`
# what changed
- removed errors that are thrown when `iframes` is defined and `experimental` is not set to `true`
# test plan
- `regression` evals


miguelg719 (approved):
changeset

PR #961, opened 2025-08-12
Add more agent evals

tkattkat:
part of STG-653

# why

Adds more evals to agent

# what changed

Added ~ 15 new evals 

# test plan
- tested locally
- tested on browserbase


evals/initStagehand.ts
```diff
@@ -104,12 +105,26 @@ export const initStagehand = async ({
...
+  const agent = stagehand.agent(agentConfig);
+
   return {
     stagehand,
     stagehandConfig: config,
     logger,
     debugUrl,
     sessionUrl,
     modelName,
+    agent,
```
seanmcguire12, line 128: nice

evals/tasks/agent/all_recipes.ts
```diff
@@ -0,0 +1,51 @@
...
+    await stagehand.page.goto("https://www.allrecipes.com/");
+
+    const agentResult = await agent.execute({
+      instruction:
+        "Search for a recipe for Beef Wellington on Allrecipes that has at least 200 reviews and an average rating of 4.5 stars or higher. List the main ingredients required for the dish.",
+      maxSteps: 20,
+    });
+    logger.log(agentResult);
+
+    const success = agentResult.success;
```
seanmcguire12: how are we thinking about evaluating success here? 

longer term I would think it would be more trustworthy (though certainly more rigid) to rely on expected values that are tied to the page state somehow, like checking if `page.url()` is what we expect. 

in the short term, what do you think about leaning more on `evaluator`?

types/evals.ts
```diff
@@ -13,6 +13,7 @@ export type StagehandInitResult = {
   sessionUrl: string;
   stagehandConfig: ConstructorParams;
   modelName: AvailableModel;
+  agent: ReturnType<Stagehand["agent"]>;
```
seanmcguire12: do we have an actual agent type

PR #963, opened 2025-08-12
Evaluator bug fix   stg 667

tkattkat:
# why

 Currently we do not properly handle passing images from evaluator to the llmclient 

# what changed

- we now provide images in the correct format to the llm from evaluator
- switched to generate object in order to consolidate complex parsing logic in evaluator 

# test plan

tested locally & on browserbase

### DIFF
--- a/types/evaluator.ts
+++ b/types/evaluator.ts
@@ -28,6 +28,10 @@ export interface BatchEvaluateOptions {
    * @default 1000
    */
   screenshotDelayMs?: number;
+  /**
+   * The reasoning behind the evaluation
+   */
+  reasoning?: string;
 }
 
 /**


.changeset/pretty-jokes-own.md
```diff
@@ -0,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+Properly handle images in evaluator + clean up response parsing logic
```
miguelg719, line 5: for future reference, eval related changes don't require changesets, only changes in the core logic (lib, types...) Doesn't hurt though

tkattkat: 🫡

Issue #899, opened 2025-07-21
Can't use AWS Bedrock custom Llm on BROWSERBASE Env


ulwiisibaq:
Hi, I seems to always failed using stagehand with bedrock on BrowserBase env, the stagehand init would always expect to pass apiKey, even though we're using bedrock custom llm client. 

is not supported yet to use bedrock on browserbase env? or is there any flag for it to skip the api key validation? 
using LOCAL env, it works fine.  

below are the sample of our constructor (the commented one are our trial with dummy api key)

```
  const stagehand = new Stagehand({
    ...StagehandConfig,
    // modelName: "aisdk",
    // modelClientOptions: {
    //   apiKey: "test",
    // },
    llmClient: new AISdkClient({
      model: bedrock("anthropic.claude-3-5-sonnet-20240620-v1:0"),
    }),
  });
```

error: `"StagehandAPIError: modelApiKey is required`

gingeekrishna:
Hi @ulwiisibaq!

This issue **has been fixed**. Here's a full summary:

### The fix

The `"modelApiKey is required"` error was removed in [#1953](https://github.com/browserbase/stagehand/pull/1953) (merged April 6, 2026). `modelApiKey` is now fully optional — custom LLM clients like AWS Bedrock handle their own authentication and don't need to pass an API key to Stagehand.

**Available since `@browserbasehq/stagehand@3.4.0`** (May 11, 2026). The current latest is `3.6.0`.

### What to do

Upgrade to `3.4.0` or later. Your existing configuration works without any changes:

```typescript
const stagehand = new Stagehand({
  env: "BROWSERBASE",
  llmClient: new AISdkClient({
    model: bedrock("anthropic.claude-3-5-sonnet-20240620-v1:0"),
  }),
  // No modelApiKey or dummy apiKey needed
});
```

If you're stuck on an older version and can't upgrade immediately, passing a placeholder as a temporary workaround:

```typescript
modelClientOptions: { apiKey: "placeholder" }
```

### Docs

PR [#966](https://github.com/browserbase/stagehand/pull/966) adds a troubleshooting entry to the Models page covering this exact error, so future users searching for it will find the answer directly in the docs.

PR #966, opened 2025-08-14
fix: allow custom LLM clients to skip modelApiKey validation in BrowserBase

gingeekrishna:
This PR fixes issue #899 by updating [api.ts] to allow custom LLM clients (such as AWS Bedrock) to skip the [modelApiKey] validation in the BrowserBase environment. The API key requirement is now enforced only for built-in providers (OpenAI, Anthropic, Google, AISDK), enabling custom LLMs to work without an API key. This resolves the error and supports broader LLM integration.

### DIFF
--- a/packages/docs/v3/configuration/models.mdx
+++ b/packages/docs/v3/configuration/models.mdx
@@ -914,6 +914,30 @@ The following models work without the `provider/` prefix in the model parameter 
 ## Troubleshooting
 
 <AccordionGroup>
+<Accordion title="Error: modelApiKey is required (custom LLM clients)">
+**Error:** `StagehandAPIError: modelApiKey is required`
+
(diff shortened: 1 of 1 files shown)


gingeekrishna:
Thanks for the review! I wanted to provide an update on the current state of this PR.

After investigating the merge conflicts, I found that the upstream repo has been significantly restructured into a monorepo (via PR #1117). As a result, `lib/api.ts` no longer exists — the equivalent file is now located at `packages/core/lib/v3/api.ts`.

Additionally, it appears the upstream has independently addressed the `modelApiKey` being optional (there's even a dedicated test `api-optional-model-api-key.test.ts` confirming this). The `"modelApiKey is required"` error no longer exists in the new `api.ts`.

Given this, I'm going to update this PR to:
1. Rebase on the latest `main` with the new monorepo structure
2. Apply the fix to `packages/core/lib/v3/api.ts` if it isn't already fully covered upstream

I'll push the updated changes shortly. If the fix is already fully covered by the upstream changes, I'll close this PR accordingly and reference the upstream fix.

PR #969, opened 2025-08-19
add streamText to aisdk llm client

tkattkat:
# why

- Currently we have no way to utilize streamtext, only generateObject, and generateText

# what changed

- Streamtext can now be utilized through createChatCompletion by passing stream:true 

# test plan

- tested locally

### DIFF
--- a/lib/llm/LLMClient.ts
+++ b/lib/llm/LLMClient.ts
@@ -60,6 +60,8 @@ export interface ChatCompletionOptions {
   tool_choice?: "auto" | "none" | "required";
   maxTokens?: number;
   requestId?: string;
+  maxSteps?: number;
+  stream?: boolean;
 }
 
 export type LLMResponse = {
--- a/lib/llm/LLMClient.ts
+++ b/lib/llm/LLMClient.ts
@@ -12,6 +12,7 @@ import {
   embedMany,
   experimental_transcribe,
   experimental_generateSpeech,
+  ToolSet,
 } from "ai";
 
 export interface ChatMessage {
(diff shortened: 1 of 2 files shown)


lib/llm/aisdk.ts
```diff
@@ -265,20 +287,11 @@ export class AISdkClient extends LLMClient {
...
 
-    const tools: Record<string, CoreTool> = {};
-
-    for (const rawTool of options.tools ?? []) {
-      tools[rawTool.name] = {
-        description: rawTool.description,
-        parameters: rawTool.parameters,
-      };
-    }
-
```
tkattkat, line 276: tools were being mapped to the incorrect format for ai sdk, we now properly handle them by just passing them directly

lib/llm/LLMClient.ts
```diff
@@ -57,9 +58,12 @@ export interface ChatCompletionOptions {
     schema: ZodType;
   };
   tools?: LLMTool[];
+  aiSDKTools?: ToolSet;
```
tkattkat, line 61: added new aiSdkTools property to easily properly utilize tools with ai sdk without modifying all of the old llm clients that use tools in the LLMTool format. Ideally we move to use just the ToolSet type once remove those clients

lib/llm/aisdk.ts
```diff
@@ -157,8 +157,29 @@ export class AISdkClient extends LLMClient {
...
+          ? {
+              openai: {
+                textVerbosity: "low",
+                reasoningEffort: "minimal",
+              },
+            }
+          : undefined,
+      });
+
+      return stream as unknown as T;
```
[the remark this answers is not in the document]
tkattkat: this is fine. since this will only be used internally, we can just pass generic streamtext type when it is utilized

PR #974, opened 2025-08-21
add details on extract with no params & extract with just a prompt

seanmcguire12:
# why

# what changed

# test plan

### DIFF
--- a/docs/basics/extract/overview.mdx
+++ b/docs/basics/extract/overview.mdx
@@ -2,7 +2,7 @@
 title: Overview
 ---
 
-`extract()` grabs structured text from the current page using structured schemas. Given instructions and `schema`, you will receive structured data.
+`extract` grabs structured data from a webpage. You can define your schema with [zod](https://github.com/colinhacks/zod) (TypeScript) or [pydantic](https://github.com/pydantic/pydantic) (Python). If you do not want to define a schema, you can also call `extract` [with no parameters](#extract-with-no-parameters), or call `extract` with just a [natural language prompt](#extract-with-just-a-prompt).
 
 <Note>
 For TypeScript, the extract schemas are defined using zod schemas.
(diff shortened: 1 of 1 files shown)


docs/basics/extract/overview.mdx
```diff
@@ -2,7 +2,7 @@
 title: Overview
 ---
 
-`extract()` grabs structured text from the current page using structured schemas. Given instructions and `schema`, you will receive structured data.
```
Kylejeong2, line 5: is this how we're deciding to start the overview page?

docs/basics/extract/overview.mdx
```diff
@@ -46,7 +46,7 @@ To extract links or URLs, in the TypeScript version of Stagehand, you'll need to
 In Python, you'll need to define it as `HttpUrl`.
 </Note>
 
-Here is how an `extract` call might look for extracting a link or URL.
+Here is how an `extract` call might look for extracting a link or URL. This also works for image links.
```
Kylejeong2, line 50: should we have a section about how this works for image links? i think it's worth highlighting how it maps an id to the image url and that's how it's being fetched

docs/basics/extract/overview.mdx
```diff
@@ -171,26 +171,87 @@ apartments = await page.extract(
...
   <ParamField path="iframes" type="boolean" optional>
       Set `iframes: true` if the extraction content exists within an iframe.
   </ParamField>
 
-  <ParamField path="useTextExtract" type="boolean" deprecated>
-    This field is now **deprecated** and has no effect.
-  </ParamField>
-
   <ParamField path="selector" type="string" optional>
       An xpath that can be used to reduce the scope of an extraction. If an xpath is passed in, `extract` will only process
```
Kylejeong2, line 259: at the end we should have some kind of cards to link to other places, probably some best practices and the extract vs observe best practice page

Kylejeong2 (changes requested):
some minor changes, also would make sure that the codeblocks look right on the actual docs page

docs/basics/extract/overview.mdx
seanmcguire12: will do this in a later PR

PR #989, opened 2025-08-22
Sc/docs revamp

sophie2chance2:
Edit the configuration pages and add cost optimization guide


docs/best_practices/cost_optimization.mdx
```diff
@@ -0,0 +1,258 @@
+---
+title: Cost Optimization  
+sidebarTitle: Cost Optimization
+description: Minimize costs while maintaining automation performance
+---
+
+Cost optimization in Stagehand involves balancing LLM inference costs with browser infrastructure costs. This guide provides practical strategies to reduce your automation expenses.
```
Kylejeong2: are we balancing inference costs with infra costs? since if inference costs go up that doesn't necessarily mean that infra costs go down etc. 

maybe should be balancing LLM inference costs and browser infrastructure costs.

docs/best_practices/cost_optimization.mdx
```diff
@@ -0,0 +1,258 @@
...
+
+## Quick Wins
+
+Start with these simple optimizations that can reduce costs:
+
+### 1. Use the Right Model for the Job
+
+Don't use premium models for simple tasks. See our [evaluation results](https://stagehand.dev/evals) for model performance and cost comparisons across different task types.
```
Kylejeong2: "We don't recommend using larger, more premium models for simple tasks."

docs/best_practices/cost_optimization.mdx
```diff
@@ -0,0 +1,258 @@
...
+### 3. Optimize Browser Sessions
+
+Reuse sessions when possible and set appropriate timeouts. See [Browser Configuration](/configuration/browser) for details:
+
+<CodeGroup>
+```typescript TypeScript
+const stagehand = new Stagehand({
+  env: "BROWSERBASE",
+  browserbaseSessionCreateParams: {
+    timeout: 1800, // 30 minutes instead of default 1 hour
```
Kylejeong2, line 74: i thought timeout was in milliseconds?

docs/best_practices/cost_optimization.mdx
```diff
@@ -0,0 +1,258 @@
...
+  Reduce costs with smart action caching and observe patterns
+</Card>
+</CardGroup>
+
+### 3. Optimize Browser Sessions
+
+Reuse sessions when possible and set appropriate timeouts. See [Browser Configuration](/configuration/browser) for details:
+
+<CodeGroup>
+```typescript TypeScript
```
Kylejeong2, line 70: is this Typescript or TypeScript? 

wondering if it shows up the same, i've been writing Typescript for the other docs.

docs/best_practices/cost_optimization.mdx
```diff
@@ -0,0 +1,258 @@
...
+    timeout: 1800, // 30 minutes instead of default 1 hour
+    keepAlive: true, // Keep session alive between tasks
+  }
+});
+```
+```python Python
+stagehand = Stagehand(
+    env="BROWSERBASE",
+    browserbase_session_create_params={
+        "timeout": 1800,  # 30 minutes instead of default 1 hour
```
Kylejeong2, line 83: same thing as above

docs/best_practices/cost_optimization.mdx
```diff
@@ -0,0 +1,258 @@
...
+
+### Intelligent Model Switching
+
+Automatically fall back to cheaper models for simple tasks:
+
+<CodeGroup>
+```typescript TypeScript
+async function smartAct(page: Page, prompt: string) {
+  // Use models from least to most expensive based on task complexity
+  // See stagehand.dev/evals for performance comparisons
```
Kylejeong2: wanna move this above the function definition

docs/best_practices/cost_optimization.mdx
```diff
@@ -0,0 +1,258 @@
...
+    } catch (error) {
+      console.log(`Falling back to ${model}...`);
+    }
+  }
+}
+```
+```python Python
+async def smart_act(page, prompt: str):
+    # Use models from least to most expensive based on task complexity
+    # See stagehand.dev/evals for performance comparisons
```
Kylejeong2: since its in both python and typescript you can move this up above

docs/configuration/browser.mdx
```diff
@@ -185,13 +182,75 @@ print(f"Session ID: {stagehand.session_id}")
...
+```
+```python Python
+result = await stagehand.init()
+print(result)
+```
+</CodeGroup>
 
-Resume a previously created Browserbase session:
+The returned object contains:
+```
```
Kylejeong2: note that this is a json

Kylejeong2 (changes requested):
few nits, but looks good otherwise

docs/best_practices/cost_optimization.mdx
sophie2chance2: its in seconds!

PR #1010, opened 2025-08-25
update agent evals

tkattkat:
# why

agent evals failing due to max step limits being hit 

# what changed

increases max steps of various agent evals 

# test plan

### DIFF
--- a/evals/tasks/agent/all_recipes.ts
+++ b/evals/tasks/agent/all_recipes.ts
@@ -14,7 +14,7 @@ export const all_recipes: EvalFunction = async ({
     const agentResult = await agent.execute({
       instruction:
         "Search for a recipe for Beef Wellington on Allrecipes that has at least 200 reviews and an average rating of 4.5 stars or higher. List the main ingredients required for the dish.",
-      maxSteps: 20,
+      maxSteps: 30,
     });
 
     const { evaluation, reasoning } = await evaluator.evaluate({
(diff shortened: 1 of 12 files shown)


evals/tasks/agent/all_recipes.ts
```diff
@@ -14,7 +14,7 @@ export const all_recipes: EvalFunction = async ({
     const agentResult = await agent.execute({
       instruction:
         "Search for a recipe for Beef Wellington on Allrecipes that has at least 200 reviews and an average rating of 4.5 stars or higher. List the main ingredients required for the dish.",
-      maxSteps: 20,
+      maxSteps: 30,
```
filip-michalsky, line 17: Should we make the max steps an ENV or a local config vs hard coding?

filip-michalsky (changes requested):
max_steps in 10+ files seems to me should be a parameter somewhere?

evals/tasks/agent/all_recipes.ts
tkattkat: probably better to do when we move format to use a json file for all of the evals otherwise we are adding a new json file that still defines them in 10+ places with no real benefit

filip-michalsky: cool we can do a fast follow PR once all of these are merged

filip-michalsky (approved):

Issue #999, opened 2025-08-22
UnsupportedModelError when using a custom llmClient with an OpenAI-compatible endpoint


ikala-qa:
Bug Description
When attempting to use Stagehand with a custom LLM by providing a custom llmClient object to the constructor, the SDK throws an UnsupportedModelError. This occurs even when following the official documentation's example for "OpenAI-compatible APIs", which involves wrapping an OpenAI client instance in a custom class that extends LLMClient.

The core issue seems to be that the Stagehand constructor does not correctly bypass its internal provider checks when a custom llmClient is provided, leading to a crash before the custom client's methods are ever called.

To Reproduce

Setup Environment:

In a Playwright project, install @browserbasehq/stagehand@2.4.3 and openai.
Create the three files below in a test directory.
File 1: [CustomOpenAIClient.js](code-assist-path:/Users/jesse/project/qa-automation-pw/tests/ui/demo/CustomOpenAIClient.js)

Create a custom wrapper class for the OpenAI client, as instructed by the documentation.
```
// CustomOpenAIClient.js
import { LLMClient } from "@browserbasehq/stagehand";

/**
 * This is a simplified, JavaScript-version of the CustomOpenAIClient template.
 * It is designed to wrap an OpenAI-compatible client and make it usable by Stagehand.
 */
export class CustomOpenAIClient extends LLMClient {
  type = "openai";
  client;
  modelName;

  constructor({ modelName, client }) {
    super(modelName);
    this.client = client;
    this.modelName = modelName;
  }

  /**
   * This is the core method Stagehand calls to interact with the LLM.
   * @param {object} options - The options for the chat completion.
   * @returns {Promise<object>} The response from the LLM, formatted for Stagehand.
   */
  async createChatCompletion({ options }) {
    const { messages } = options;

    const body = {
      messages: messages,
      model: this.modelName,
      stream: false,
    };

    try {
        const response = await this.client.chat.completions.create(body);

        // Return the specific format that Stagehand expects
        return {
          data: response.choices[0].message.content,
          usage: {
            prompt_tokens: response.usage?.prompt_tokens ?? 0,
            completion_tokens: response.usage?.completion_tokens ?? 0,
            total_tokens: response.usage?.total_tokens ?? 0,
          },
        };
    } catch (error) {
        console.error("Error in CustomOpenAIClient:", error);
        throw error;
    }
  }
}
```
[...]

PR #1012, opened 2025-08-25
Patch disabling api whenever custom LLM clients are provided

miguelg719:
# why
Custom LLM clients are throwing an error on api mode

# what changed
We added logic for disabling api whenever custom LLM clients are provided, but there was a reassignment (re-enabling api mode) downstream

# test plan


miguelg719:
fixes #999

PR #1014, opened 2025-08-25
replace operator agent with base of new agent 

tkattkat:
# why

replace operator agent with new one

# what changed

- removed operator agent,
- added new agent handler 

# test plan
tested locally + ran evals


lib/agent/tools/act.ts
```diff
@@ -0,0 +1,36 @@
...
+          - type "John" into the first name input
+          - type "Doe" into the last name input`),
+    }),
+    execute: async ({ parameters }) => {
+      const [observeResult] = await page.observe({
+        modelName: "google/gemini-2.5-flash",
+        instruction: parameters,
+      });
+      if (observeResult) {
+        const isIframe = observeResult.description === "an iframe";
```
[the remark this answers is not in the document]
tkattkat: this is where we will implement iframe handling

lib/handlers/stagehandAgentHandler.ts
```diff
@@ -0,0 +1,224 @@
...
+        );
+      }
+
+      if (!isAISDKBacked(this.llmClient)) {
+        throw new Error(
+          "StagehandAgentHandler requires an AISDK-backed LLM client. Ensure your model is configured like 'openai/gpt-4.1-mini' or another AISDK provider.",
+        );
+      }
+      const baseModel: LanguageModel = this.llmClient.getLanguageModel();
+      const wrappedModel = wrapLanguageModel({
```
filip-michalsky, line 76: is the idea to put more middleware logic here later after the agent does each step?

lib/handlers/stagehandAgentHandler.ts
```diff
@@ -0,0 +1,224 @@
...
+
+      const result = await this.llmClient.generateText({
+        model: wrappedModel,
+        system: systemPrompt,
+        messages,
+        tools: allTools,
+        maxSteps,
+        temperature: 1,
+        toolChoice: "auto",
+        onStepFinish: async (event) => {
```
filip-michalsky, line 94: since these are async, if the LLM wants to call multiple tools - do we support parallel tool calling?

lib/handlers/stagehandAgentHandler.ts
```diff
@@ -0,0 +1,224 @@
...
+              if (event.text.length > 0) {
+                collectedReasoning.push(event.text);
+                this.logger({
+                  category: "agent",
+                  message: `reasoning: ${event.text}`,
+                  level: 1,
+                });
+              }
+
+              if (toolCall.toolName === "close") {
```
filip-michalsky, line 114: the danger here is that if we have an array of tool calls coming from the AI sdk, the close tool may not be the last in the array, which could cause our agent to terminate prematurely

lib/handlers/stagehandAgentHandler.ts
```diff
@@ -0,0 +1,224 @@
...
+      return {
+        success: completed,
+        message: finalMessage || "Task execution completed",
+        actions,
+        completed,
+        usage: result.usage
+          ? {
+              input_tokens: result.usage.promptTokens || 0,
+              output_tokens: result.usage.completionTokens || 0,
+              inference_time_ms: 0,
```
filip-michalsky: why is this hardcoded as 0?

lib/handlers/stagehandAgentHandler.ts
```diff
@@ -0,0 +1,224 @@
...
+            }
+          : undefined,
+      };
+    } catch (error) {
+      const errorMessage =
+        error instanceof Error ? error.message : String(error);
+      this.logger({
+        category: "agent",
+        message: `Error executing agent task: ${errorMessage}`,
+        level: 0,
```
filip-michalsky, line 168: logging level should be env var?

lib/handlers/stagehandAgentHandler.ts
```diff
@@ -0,0 +1,224 @@
...
+        messages,
+        tools: allTools,
+        maxSteps,
+        temperature: 1,
+        toolChoice: "auto",
+        onStepFinish: async (event) => {
+          this.logger({
+            category: "agent",
+            message: `Step finished: ${event.finishReason}`,
+            level: 2,
```
filip-michalsky, line 98: logging level should be env var?

lib/agent/tools/fillform.ts
```diff
@@ -0,0 +1,68 @@
...
+    parameters: z.object({
+      fields: z
+        .array(
+          z.object({
+            action: z
+              .string()
+              .describe(
+                'Description of the typing action, e.g. "type foo into the bar field"',
+              ),
+            value: z.string().describe("Text to type into the target field"),
```
filip-michalsky, line 45: am I crazy or I just don't see where the values are being passed in here? also the credit card form fill is an important use case, does this tool NOT send sensitive PII data to the LLM and uses masking?

lib/handlers/stagehandAgentHandler.ts
```diff
@@ -0,0 +1,224 @@
...
+5. If the task cannot be completed, use "close" with taskComplete: false
+
+TOOLS OVERVIEW:
+- screenshot: Take a compressed JPEG screenshot for quick visual context (use sparingly)
+- ariaTree: Get an accessibility (ARIA) hybrid tree for full page context (preferred for understanding layout and elements)
+- act: Perform a specific atomic action (click, type, etc.)
+- extract: Extract structured data
+- goto: Navigate to a URL
+- wait/navback/refresh: Control timing and navigation
+- scroll: Scroll the page x pixels up or down
```
filip-michalsky, line 210: we need to document that if we want to update more tools, we need to update this prompt string

lib/agent/tools/ariaTree.ts
```diff
@@ -0,0 +1,35 @@
...
+  tool({
+    description:
+      "gets the accessibility (ARIA) tree from the current page. this is useful for understanding the page structure and accessibility features. it should provide full context of what is on the page",
+    parameters: z.object({}),
+    execute: async () => {
+      const { page_text } = await page.extract();
+      const pageUrl = page.url();
+
+      let content = page_text;
+      const MAX_TOKENS = 70000;
```
filip-michalsky: why is this hard coded?

lib/agent/tools/ariaTree.ts
```diff
@@ -0,0 +1,35 @@
...
+
+      const estimatedTokens = Math.ceil(content.length / 4);
```
filip-michalsky, line 17: why are you dividing by 4?

lib/agent/tools/extract.ts
```diff
@@ -0,0 +1,92 @@
+import { tool } from "ai";
+import { z } from "zod";
+import { Page } from "@/types/page";
+
+/**
+ * Evaluates a Zod schema string and returns the actual Zod schema
+ * Uses Function constructor to evaluate the schema string in a controlled way
+ */
+function evaluateZodSchema(schemaStr: string): z.ZodTypeAny {
```
filip-michalsky: maybe we might want to add an LLM based schema healer here?

lib/agent/tools/goto.ts
```diff
@@ -0,0 +1,15 @@
...
+import { Page } from "@/types/page";
+
+export const createGotoTool = (page: Page) =>
+  tool({
+    description: "Navigate to a specific URL",
+    parameters: z.object({
+      url: z.string().describe("The URL to navigate to"),
+    }),
+    execute: async ({ url }) => {
+      await page.goto(url, { waitUntil: "load" });
```
filip-michalsky: no timeout on waiting?

lib/handlers/stagehandAgentHandler.ts
tkattkat: on every step the context gets ran through middleware yea

lib/handlers/stagehandAgentHandler.ts
tkattkat: if the close tool is called by agent it will always be last

lib/handlers/stagehandAgentHandler.ts
tkattkat: no it is just a number everywhere in codebase

lib/agent/tools/goto.ts
tkattkat: we could add one but I figure the default of 30s is fine?

lib/agent/tools/fillform.ts
tkattkat: the values get passed by the llm based on the schema, and are within the params used in execute function. in the future we should probably add some form of support for variables in agent though somehow

lib/handlers/stagehandAgentHandler.ts
tkattkat: no its just hardcoded number

lib/handlers/stagehandAgentHandler.ts
tkattkat: parallel tool calling is automatically handled by ai sdk and enabled by default

lib/agent/tools/ariaTree.ts
tkattkat: a token is roughly 4 characters

lib/agent/tools/ariaTree.ts
tkattkat: because it will not change

lib/agent/tools/extract.ts
tkattkat: agent will automatically do this if it fails, but it essentially never fails to pass proper schema. could be good to add this in the future but tbd on whether we even really want to use extract in agent aside from tasks that extract, and return the json in final response which is planned thing to add in near future using syntax similar to this 


```agent.execute({
  instruction: "find the top selling products on amazon",
  schema: z.object({
    products: z.array({
      name: z.string(),
      price_usd: z.number(),
    }),
  })
})

File: lib/handlers/stagehandAgentHandler.ts (hunk 9 above)
Comment (tkattkat) [reply lib/handlers/stagehandAgentHandler.ts:210]: I think in future we likely will not have these in actual system prompt at all but still unsure tbh

File: lib/handlers/stagehandAgentHandler.ts (hunk 5 above)
Comment (tkattkat) [reply lib/handlers/stagehandAgentHandler.ts]: updated to use same logic as before for inference time

File: lib/handlers/stagehandAgentHandler.ts (hunk 9 above)
Comment (tkattkat) [reply lib/handlers/stagehandAgentHandler.ts:210]: - added comment above system prompt

**Review by filip-michalsky — APPROVED** [review]
feedback addressed, agent runs locally, and braintrust traces were examined too.

File: lib/agent/tools/act.ts [hunk 14]
```diff
@@ -0,0 +1,38 @@
+import { tool } from "ai";
+import { z } from "zod";
+import { Page } from "@/types/page";
+
+export const createActTool = (page: Page, executionModel?: string) =>
+  tool({
+    description: "Perform an action on the page (click, type, etc)",
+    parameters: z.object({
+      action: z.string()
+        .describe(`Describe what to click in a short, specific phrase that mentions the element type.
```
miguelg719: describe what to click seems vague, let's add more substance

lib/agent/tools/act.ts
```diff
@@ -0,0 +1,38 @@
...
+          - type "John" into the first name input
+          - type "Doe" into the last name input`),
+    }),
+    execute: async ({ action }) => {
+      const [observeResult] = executionModel
+        ? await page.observe({
+            instruction: action,
+            modelName: executionModel,
+          })
+        : await page.observe(action);
```
miguelg719: why not call page.act()?

lib/agent/tools/act.ts
```diff
@@ -0,0 +1,38 @@
...
+      if (observeResult) {
+        const isIframe = observeResult.description === "an iframe";
+        if (isIframe) {
+          return {
+            success: false,
+            error: "Iframe encountered",
+          };
+        }
```
miguelg719, line 44: prob worth passing iframes:true/false as an optional tool param

lib/agent/tools/ariaTree.ts
```diff
@@ -0,0 +1,35 @@
...
+
+      if (estimatedTokens > MAX_TOKENS) {
+        const maxCharacters = MAX_TOKENS * 4;
+        content =
+          content.substring(0, maxCharacters) +
+          "\n\n[CONTENT TRUNCATED: Exceeded 70,000 token limit]";
```
miguelg719, line 23: I'm curious, did we hit this log on any evals?

lib/agent/tools/extract.ts
```diff
@@ -0,0 +1,92 @@
+import { tool } from "ai";
+import { z } from "zod";
```
miguelg719: can't have this, need to specify the import from "zod/v3" see https://github.com/browserbase/stagehand/pull/944

miguelg719: same for all other imports

lib/agent/tools/extract.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+}
+
+export const createExtractTool = (page: Page, executionModel?: string) =>
+  tool({
+    description: `Extract structured data from the current page based on a provided schema.
+    
+    USAGE GUIDELINES:
+    - Keep schemas MINIMAL - only include fields essential for the task
+    - All fields in the schema should be optional (.optional())
```
miguelg719: why?

lib/agent/tools/extract.ts
```diff
@@ -0,0 +1,92 @@
...
+        const result = executionModel
+          ? await page.extract({
+              instruction,
+              schema: schemaObject,
+              modelName: executionModel,
+            })
+          : await page.extract({
+              instruction,
+              schema: schemaObject,
+            });
```
miguelg719: should be cleaner, you can use ?? to check. Also if you pass an executionModel from a different provider (for example claude-4 is driving and gemini executing) it will not pass the correct provider api key, might need modelClientOptions

lib/agent/tools/goto.ts
miguelg719: page.goto has waitForSettleDom by default, I would remove the waitUntil param

lib/agent/tools/goto.ts
```diff
@@ -0,0 +1,15 @@
...
+      return { success: true, url };
```
miguelg719: Also, might be worth wrapping in a try/catch in case the goto timeout is hit? so the agent can try another url

lib/handlers/stagehandAgentHandler.ts
```diff
@@ -0,0 +1,229 @@
...
+
+STRATEGY:
+- Prefer ariaTree to understand the page before acting; use screenshot for quick confirmation.
+- Keep actions atomic and verify outcomes before proceeding.
+
+For each action, provide clear reasoning about why you're taking that step.`;
+  }
+
+  private createTools() {
+    const page = this.stagehandPage.page;
```
miguelg719: I'm curious if it would be better to pass the stagehand object downstream rather than just the page. Reasoning:
- if we want to add CUA as another tool for the driving agent, we would need to call stagehand.agent()
- might be easier to pass stagehand.logger downstream

lib/handlers/stagehandAgentHandler.ts
```diff
@@ -0,0 +1,229 @@
...
+import { CoreMessage, wrapLanguageModel } from "ai";
+import { LanguageModel } from "ai";
+import { processMessages } from "../agent/utils/messageProcessing";
+import { createAgentTools } from "../agent/tools";
+import { ToolSet } from "ai";
+function isAISDKBacked(client: LLMClient): client is LLMClient & {
+  type: "aisdk";
+  getLanguageModel: () => LanguageModel;
+} {
+  return client?.type === "aisdk" && "getLanguageModel" in client;
```
miguelg719: don't you just need llmClient.type? don't understand this function

lib/agent/tools/act.ts
tkattkat: is there a way to check for an iframe with act?

lib/agent/tools/extract.ts
tkattkat: minimal schema fields - suggestion from sean 
optional schema fields - extract can fail so it is better to use safer options with agent

lib/agent/tools/act.ts
sameelarif: i think this is a good use case for observe. it's acting as a prepatory step before act

lib/agent/tools/ariaTree.ts
```diff
@@ -0,0 +1,35 @@
...
+      const { page_text } = await page.extract();
+      const pageUrl = page.url();
+
+      let content = page_text;
+      const MAX_TOKENS = 70000;
+
+      const estimatedTokens = Math.ceil(content.length / 4);
+
+      if (estimatedTokens > MAX_TOKENS) {
+        const maxCharacters = MAX_TOKENS * 4;
```
sameelarif: nit but would just define `MAX_CHARACTERS` instead since this isn't a "true" token count

lib/handlers/stagehandAgentHandler.ts
tkattkat: moved getLanguageModel to llmclient so we can remove this while having proper typing

lib/agent/tools/ariaTree.ts
tkattkat: no but the limit can definitely be exceeded easily depending on the website

lib/agent/tools/act.ts
tkattkat: What would you suggest instead? This has always worked well for me

lib/agent/tools/extract.ts
```diff
@@ -0,0 +1,87 @@
...
+ * Uses Function constructor to evaluate the schema string in a controlled way
+ */
+function evaluateZodSchema(schemaStr: string): z.ZodTypeAny {
+  try {
+    // Create a function that returns the evaluated schema
+    // We pass z as a parameter to make it available in the evaluated context
+    const schemaFunction = new Function("z", `return ${schemaStr}`);
+    return schemaFunction(z);
+  } catch (error) {
+    console.warn("Failed to evaluate schema string, using z.any():", error);
```
miguelg719: logger

lib/agent/tools/screenshot.ts
```diff
@@ -0,0 +1,35 @@
...
+      "Takes a screenshot of the current page. Use this tool to learn where you are on the page, or to get context of elements on the page",
+    parameters: z.object({}),
+    execute: async () => {
+      const screenshotBuffer = await page.screenshot({
+        fullPage: false,
+        type: "jpeg",
+      });
+      const pageUrl = page.url();
+
+      console.log(`Screenshot size: ${screenshotBuffer.length} bytes`);
```
miguelg719: logger

lib/agent/tools/screenshot.ts
```diff
@@ -0,0 +1,35 @@
...
+
+      return {
+        base64: screenshotBuffer.toString("base64"),
+        timestamp: Date.now(),
+        pageUrl,
+      };
+    },
+    experimental_toToolResultContent: (result) => {
+      console.log(`Base64 length: ${result.base64.length} characters`);
```
miguelg719: logger

lib/agent/tools/extract.ts
miguelg719: not resolved

lib/agent/tools/extract.ts
miguelg719: @seanmcguire12 is this the case? I have almost never had to set optional on an extract schema

lib/agent/tools/act.ts
miguelg719: that's what act does really 
https://github.com/browserbase/stagehand/blob/ab5d6ede19aabc059badc4247f1cb2c6c9e71bae/lib/handlers/actHandler.ts#L207-L310
But supports things like variables and has a more elaborate system prompt
https://github.com/browserbase/stagehand/blob/ab5d6ede19aabc059badc4247f1cb2c6c9e71bae/lib/prompt.ts#L150-L176

lib/agent/tools/extract.ts
```diff
@@ -0,0 +1,105 @@
...
+    
+    EXAMPLES:
+    1. Extract a single value:
+       instruction: "extract the product price"
+       schema: "z.object({ price: z.number().optional() })"
+    
+    2. Extract multiple fields:
+       instruction: "extract product name and price"
+       schema: "z.object({ name: z.string().optional(), price: z.number().optional() })"
+
```
seanmcguire12, line 55: could we add a note note a about extracting links

lib/handlers/stagehandAgentHandler.ts
```diff
@@ -0,0 +1,225 @@
...
+        success: false,
+        actions,
+        message: `Failed to execute task: ${errorMessage}`,
+        completed: false,
+      };
+    }
+  }
+
+  // in the future if we continue to describe tools in system prompt, we need to make sure to update them in here when new tools are added or removed. still tbd on whether we want to keep them in here long term.
+  private buildSystemPrompt(
```
miguelg719, line 181: Note for future, we should move this to prompts.ts

miguelg719 (approved):

seanmcguire12 (approved):

PR #1015, opened 2025-08-25
add evaluate text to evaluator

tkattkat:
# why

adds "evaluateText" to evaluator

# what changed

evals can now use the evaluateText method to use the llms messages to evaluate if it came across the proper information required by the eval. This removes the dependency on extract within evals + makes it easier to evaluate within the webvoyager / gaia evals when they are implemented 

# test plan
tested locally

### DIFF
--- a/evals/evals.config.json
+++ b/evals/evals.config.json
@@ -515,10 +515,6 @@
     {
       "name": "agent/all_recipes",
       "categories": ["agent"]
-    },
-    {
-      "name": "agent/google_shopping",
-      "categories": ["agent"]
     }
   ]
 }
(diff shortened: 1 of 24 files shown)


miguelg719:
What do you think about changing EvaluateOptions to take either text or screenshot (or both) and use those in .evaluate / .batchEvaluate rather than separate functions? @tkattkat

tkattkat:
> What do you think about changing EvaluateOptions to take either text or screenshot (or both) and use those in .evaluate / .batchEvaluate rather than separate functions? @tkattkat

yea that would probably be cleaner and remove redundant code. will do

PR #1023, opened 2025-08-26
fix: set viewport to 1024x768 to make cua models work on advanced stealth

Kylejeong2:
# why

The viewport changes to 1288x711 when using advanced stealth, which causes computer use agents to mis-click etc. since they're trained on 1024x768.

# what changed

Added a functions to ensure viewport is set to the right size for CUA models.

# test plan

tested on evals, watched live sessions to make sure it was resizing correct (if needed) and advanced stealth didn't mess up the viewport

### DIFF
--- a/lib/handlers/agentHandler.ts
+++ b/lib/handlers/agentHandler.ts
@@ -119,6 +119,18 @@ export class StagehandAgentHandler {
     // Update viewport and URL for any client type
     this.updateClientViewport();
     this.updateClientUrl();
+
+    // For CUA models, ensure the browser viewport is set to the expected size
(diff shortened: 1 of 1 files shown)


miguelg719:
okay what we need to do here is: 
- on the agentHandler, we get the viewport, height, and devicePixelRatio using page.evaluate()
- set those before sending to the LLM (these won't change, so we can just override currentViewport)
- update **all** coordinate-based actions to multiply by the devicePixelRatio

cc: @Kylejeong2

miguelg719:
the question is, do we want this only on advanced stealth?

PR #1046, opened 2025-09-02
add gpt 5 handling to operator handler

tkattkat:
# why

when using operator handler with gpt-5, the temperature is not being set to 1 

# what changed

the temperature now gets set to 1 when gpt-5 is being used 

# test plan


miguelg719 (approved):
changeset

PR #1048, opened 2025-09-03
azure/bedrock api integration

sameelarif:
# why
Need to support Azure/Bedrock and all other AI SDK providers that have more than just an `apiKey` parameter.

# what changed
Allowed free input of `modelClientOptions` and added logic for authenticating with AWS for Bedrock usage.

# test plan


lib/api.ts
```diff
@@ -261,6 +265,12 @@ export class StagehandAPI {
       "x-sdk-version": STAGEHAND_VERSION,
     };
 
+    // Add modelClientOptions as a header if provided
+    if (modelClientOptions) {
+      defaultHeaders["x-model-client-options"] =
+        JSON.stringify(modelClientOptions);
```
miguelg719: why send it as a header? modelClientOptions already gets sent in the payload. A stringified json in a header is probably not the move

examples/example-bedrock.ts
```diff
@@ -0,0 +1,101 @@
...
+ * - anthropic.claude-3-5-sonnet-20240620-v1:0 (best for extraction/observation)
+ * - anthropic.claude-3-haiku-20240307-v1:0 (faster, good for basic tasks)
+ * - openai.gpt-oss-120b-1:0 (good for navigation and simple operations)
+ */
+
+async function runBedrockExample() {
+  // Initialize Stagehand with AWS Bedrock
+  const stagehand = new Stagehand({
+    env: "BROWSERBASE",
+    modelName: "bedrock/us.anthropic.claude-sonnet-4-20250514-v1:0",
```
filip-michalsky, line 47: Please don't hard code the model name as some users will not have this model name - they have to go to AWS dashboard anyway to get the token/access creds

filip-michalsky (changes requested):

examples/example-bedrock.ts
filip-michalsky: I couldnt run the example bc of that

lib/api.ts
```diff
@@ -191,6 +185,12 @@ export class StagehandAPI {
     const queryString = urlParams.toString();
     const url = `/sessions/${this.sessionId}/${method}${queryString ? `?${queryString}` : ""}`;
 
+    this.logger({
+      category: "execute",
+      message: `Executing ${method} with args: ${JSON.stringify(args)}`,
+      level: 1,
```
miguelg719: maybe level 2

lib/StagehandPage.ts
```diff
@@ -747,6 +747,7 @@ ${scriptContent} \
             const result = await this.api.act({
               ...observeResult,
               frameId: this.rootFrameId,
+              modelClientOptions: this.stagehand["modelClientOptions"],
```
miguelg719, line 750: is this required?

lib/StagehandPage.ts
```diff
@@ -844,7 +846,10 @@ ${scriptContent} \
       if (!instructionOrOptions) {
         let result: ExtractResult<T>;
         if (this.api) {
-          result = await this.api.extract<T>({ frameId: this.rootFrameId });
+          result = await this.api.extract<T>({
+            frameId: this.rootFrameId,
+            modelClientOptions: this.stagehand["modelClientOptions"],
```
miguelg719, line 851: is this required?

lib/utils.ts
```diff
@@ -493,6 +494,153 @@ export function loadApiKeyFromEnv(
...
+    }
+  }
+
+  // No authentication method found
+  logger({
+    category: "init",
+    message:
+      "No Amazon Bedrock authentication credentials found. Please provide credentials via modelClientOptions (accessKeyId/secretAccessKey or bearerToken) or environment variables (AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY or AWS_BEARER_TOKEN_BEDROCK)",
+    level: 0,
+  });
```
miguelg719, line 639: should we throw here

lib/StagehandPage.ts
sameelarif: We will need this once we add self-healing

lib/StagehandPage.ts
sameelarif: Yes because otherwise it doesn't get sent to the API. We need this param on all api calls now

lib/utils.ts
sameelarif: It would throw once they make an LLM call. We let people use stagehand without an LLM so I figured we just print a warning for now. Down to change it tho

PR #1050, opened 2025-09-04
add agent reasoning to evaluator, improve prompts & paramaterize max …

tkattkat:
…steps

# why

- Evaluator often is too strict in its evaluation, resulting in false positives 
- max step limits in evals are very brittle and can cause false negatives on tasks that exceed the current limit
- Evaluator can sometimes view screenshots as a single source of truth resulting in false negatives. Providing the agents reasoning alongside these heps mitigate this
# what changed

- paramaterized max steps to allow for easy configuration across all evals through env 
- adjusted evaluator prompting 
- added "agent reasoning" param which can be used to pass in the agents reasoning alongside bulk screenshots in a format evaluator can understand well 

# test plan
tested locally

### DIFF
--- a/evals/evaluator.ts
+++ b/evals/evaluator.ts
@@ -55,6 +55,7 @@ export class Evaluator {
       screenshot = true,
       systemPrompt,
       screenshotDelayMs = 250,
+      agentReasoning,
     } = options;
     if (!question) {
       throw new Error("Question cannot be an empty string");
@@ -69,12 +70,12 @@ export class Evaluator {
         question,
         screenshots: screenshot,
         systemPrompt,
+        agentReasoning,
       });
     }
 
     // Single screenshot case (existing logic)
-    const defaultSystemPrompt = `You are an expert evaluator that confidently returns YES or NO given a question and the state of a task (in the form of a screenshot, or an answer). Provide a detailed reasoning for your answer.
-          Be critical about the question and the answer, the slightest detail might be the difference between yes and no. for text, be lenient and allow for slight variations in wording. we will be comparing the agents trajectory to see if it contains the information we were looking for in the answer.
+    const defaultSystemPrompt = `You are an expert evaluator that confidently returns YES or NO based on if the original goal was achieved. You have access to  ${screenshot ? "a screenshot" : "the agents reasoning and actions throughout the task"} that you can use to evaluate the tasks completion. Provide detailed reasoning for your answer.
           Today's date is ${new Date().toLocaleDateString()}`;
 
     await new Promise((resolve) => setTimeout(resolve, screenshotDelayMs));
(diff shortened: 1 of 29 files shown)


filip-michalsky:
can we pls rename the branch so I can check out locally :D

evals/evaluator.ts
```diff
@@ -69,12 +70,12 @@ export class Evaluator {
...
         screenshots: screenshot,
         systemPrompt,
+        agentReasoning,
       });
     }
 
     // Single screenshot case (existing logic)
-    const defaultSystemPrompt = `You are an expert evaluator that confidently returns YES or NO given a question and the state of a task (in the form of a screenshot, or an answer). Provide a detailed reasoning for your answer.
-          Be critical about the question and the answer, the slightest detail might be the difference between yes and no. for text, be lenient and allow for slight variations in wording. we will be comparing the agents trajectory to see if it contains the information we were looking for in the answer.
+    const defaultSystemPrompt = `You are an expert evaluator that confidently returns YES or NO based on if the original goal was achieved. You have access to  ${screenshot ? "a screenshot" : "the agents reasoning and actions throughout the task"} that you can use to evaluate the tasks completion. Provide detailed reasoning for your answer.
```
filip-michalsky, line 78: can we double-check the logic? this is saying if there is a screenshot, use it, if not use the reasoning trace, but in the types below only agentReasoning param is optional while screenshot is not optional? This should be both if both available no?

evals/tasks/agent/webvoyager.ts
```diff
@@ -60,8 +60,9 @@ export const webvoyager: EvalFunction = async ({
 
     const evaluator = new Evaluator(stagehand);
     const evalResult = await evaluator.ask({
-      question: `Did the agent successfully complete this task: "${params.ques}"? Look at all the screenshots showing the progression of the task to verify if it was completed successfully.`,
+      question: `Did the agent successfully complete this task: "${params.ques}"?`,
       screenshot: screenshots,
+      agentReasoning: agentResult.message,
```
filip-michalsky: can we double click on this? not all models have reasoning. is `.message` always there? what if `agentResult` is None?

filip-michalsky:
when my two comments are addressed, should be g2g! runs like a charm locally

filip-michalsky (approved):

evals/tasks/agent/webvoyager.ts
tkattkat: if agent result is none, the task most likely failed

evals/evaluator.ts
tkattkat: not necessarily no, if the reasoning is passed in, it gets put in the normal LLM call rather than the system prompt

filip-michalsky: ok cool

evals/tasks/agent/webvoyager.ts
filip-michalsky: right but you should still at least do a safer getter

PR #1070, opened 2025-09-12
add playwright screenshot option for browserbase env

derekmeegan:
# why

Currently, using playwright screenshot command is not available when the execution environment is Stagehand. A customer has indicated they would prefer to use Playwright's native screenshot command instead of CDP when using Browserbase as CDP screenshot causes unexpected behavior for their target site.

# what changed

- added a StagehandScreenshotOptions type with useCDP argument added
- extended page type to accept custom stagehand screeenshot options
- update screenshot proxy to default useCDP to true if the env is browserbase and use playwright screenshot if false
- added eval for screenshot with and without cdp

# test plan
- tested and confirmed functionality with eval and external example script (not committed)

### DIFF
--- a/lib/StagehandPage.ts
+++ b/lib/StagehandPage.ts
@@ -421,7 +421,9 @@ ${scriptContent} \
           // Handle screenshots with CDP
           if (prop === "screenshot") {
             return async (options: StagehandScreenshotOptions = {}) => {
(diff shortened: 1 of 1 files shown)


miguelg719:
changeset and gtg @derekmeegan

PR #1079, opened 2025-09-16
patch api agent config

tkattkat:
# why

When a completely empty stagehand config is passed, an error involving integrations is thrown 

# what changed

fallback to empty agent config when options is not present 

# test plan

tested locally


miguelg719 (approved):
need a changeset

Issue #190, opened 2024-11-20
Allow users to specify where they want to store the cache instead of having it hardcoded


kamath:

its-jz:
Addressed not creating a new directory when enableCaching is set to false here #191

PR #1091, opened 2025-09-22
Allow configuring Stagehand's cache directory

MagellaX:
Let folks point the cache at a writable path via the new cacheDir option or STAGEHAND_CACHE_DIR, document it, and cover the behavior with regression tests. 
Covers #190


MagellaX:
@seanmcguire12 all good!

seanmcguire12:
closing this one as we've taken a different approach to caching in stagehand v3

Issue #263, opened 2024-12-03
Get off of `tsup` and use `esbuild` instead


kamath:
We're using an old and unmaintained build tool. ESBuild or other build tools are more prevalent and well maintained, and we should switch to that instead. Would love suggestions from the community!

molvqingtai:
The tsup author is using tsdown.

PR #1095, opened 2025-09-24
 Replace tsup with esbuild and precompile deterministic Playwright suite

MagellaX:
Swapped out tsup for esbuild so the Stagehand build stays fast and maintained, and taught Playwright to run against precompiled JS instead of scrambling at runtime. We now bundle the library and CLI with tiny esbuild scripts, compile the deterministic suite ahead of time, rewrite its imports, and keep the helper that Playwright was missing. The net effect: pnpm run build and the local e2e suite run clean again, no more __name explosions, and the build pipeline feels a lot healthier.
covers #263

### DIFF
--- a/scripts/build-e2e.mjs
+++ b/scripts/build-e2e.mjs
@@ -40,8 +40,10 @@ function patchImportMeta(filePath) {
     return;
   }
 
-  content = content.replace("var import_meta = {};", IMPORT_META_DECL);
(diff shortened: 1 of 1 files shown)


MagellaX:
Hey, i think u can merge this @seanmcguire12

pirate:
I'm picking up the torch here and finishing off this work in https://github.com/browserbase/stagehand/pull/1632

PR #1101, opened 2025-10-01
Update cerebras provider

kevint-cerebras:
# why
The Cerebras integration was out of date when it came to models and also nonfunctional during testing.

# what changed
Updated the Cerebras provider file to work with OpenAI format + added most recently released Cerebras models.

# test plan
I used the form_filling_sensible file and the 2048 tests to run through all the Cerebras models in the PR


kevint-cerebras:
@greptileai

PR #1114, opened 2025-10-08
set default viewport when running on browserbase

seanmcguire12:
# why
- need to set default viewport when running on browserbase. previously, we only defined the default inside the exported `StagehandConfig` 
# what changed
- set default viewport to 1288 * 711 when running on browserbase
# test plan
- tested locally,
- regression evals

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -45,6 +45,7 @@ import { StagehandAgentHandler } from "./handlers/stagehandAgentHandler";
 import { StagehandLogger } from "./logger";
 import { connectToMCPServer } from "./mcp/connection";
 import { resolveTools } from "./mcp/utils";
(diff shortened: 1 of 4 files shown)


lib/index.ts
```diff
@@ -642,7 +646,9 @@ export class Stagehand {
 
     this.domSettleTimeoutMs = domSettleTimeoutMs ?? 30_000;
     this.headless = localBrowserLaunchOptions?.headless ?? false;
-    this.browserbaseSessionCreateParams = browserbaseSessionCreateParams;
+    this.browserbaseSessionCreateParams = applyDefaultBrowserSettingsViewport(
+      browserbaseSessionCreateParams,
+    );
```
miguelg719: if we have it on getBrowser do we need it here

miguelg719 (approved):

PR #1123, opened 2025-10-10
add pageurl & timestamp to agent actions

tkattkat:
# why

currently, we have no sense of what url an action was taken on and what time it was taken when using agent 

this is useful to have, because in the dashboard it will allow us to filter the agents actions by url, and display timestamps of the individual actions  

# what changed

- added pageUrl to every action 
- added timestamp to every action

the url is grabbed prior to the action being taken. This is because if we do it after the action is taken, there is a chance the action could have caused a navigation, which would result in the incorrect url for the action 

# test plan

tested locally

### DIFF
--- a/.changeset/salty-moles-cross.md
+++ b/.changeset/salty-moles-cross.md
@@ -1,5 +1,0 @@
----
-"@browserbasehq/stagehand": patch
----
-
-add pageurl & timestamp to agent actions


.changeset/upset-ideas-fall.md
miguelg719, line 1: we should keep a single changeset

tkattkat: lmaooo my bad i read and outdated comment saying i was missing one and thought i forgot to add one

PR #1129, opened 2025-10-13
Support Gemini 2.5 model family with Google GenAI client

mcndt:
# why

I want to use the latest Gemini 2.5 models.

# what changed

I updated the list of available models and mapping to model provider to include the following models:

- Gemini 2.5 Flash
- Gemini 2.5 Flash (September 2025 Preview)
- Gemini 2.5 Flash Lite
- Gemini 2.5 Flash Lite (September 2025 Preview)
- Gemini 2.5 Pro

# test plan

Works out of the box with the Google GenAI client.

PR #1130, opened 2025-10-13
Support full Google GenAI client options (support Vertex AI as alternative to Gemini API key)

mcndt:
# why

1. In my organisation, we use Vertex AI for authenticated LLM calls using Google Cloud service accounts instead of the Gemini API keys.
2. I also need control over the Vertex AI location being used for data residency guarantees.

# what changed

I updated the typing of `ClientOptions` to support the full Google GenAI client constructor options, just like the Anthropic and OpenAI client options are already passed verbatim.

Because the Google GenAI client options don't have a `baseUrl` key, I also added explicit typing for the client options used for the ai-sdk model provider.

# test plan

As this is only a typing change, I foresaw no immediate need for unit tests. Authentication with VertexAI is already tested as part of the GenAI client package.

### DIFF
--- a/.changeset/evil-clubs-attend.md
+++ b/.changeset/evil-clubs-attend.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+Add support for google vertex provider
--- a/.changeset/four-knives-ask.md
+++ b/.changeset/four-knives-ask.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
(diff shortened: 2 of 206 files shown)


mcndt:
If in the future you want to only support the AI SDK (basing this off  https://github.com/browserbase/stagehand/pull/1129#issuecomment-3400295962), this is also possible, but typing will need to be expanded to allow passing other options than simply an API key and base URL:

Docs:
- https://ai-sdk.dev/providers/ai-sdk-providers/google-vertex

PR #1136, opened 2025-10-15
fix apikey loading agent evals

miguelg719:
# why

# what changed

# test plan


packages/evals/taskConfig.ts
```diff
@@ -109,7 +109,8 @@ const DEFAULT_AGENT_MODELS = process.env.EVAL_AGENT_MODELS
   : [
       "computer-use-preview-2025-03-11",
       "claude-sonnet-4-20250514",
-      "anthropic/claude-sonnet-4-20250514",
+      "gemini-2.5-computer-use-preview-10-2025",
+      // "anthropic/claude-sonnet-4-20250514",
```
miguelg719, line 113: will add back, following on another pr to differentiate cua/non-cua

PR #1139, opened 2025-10-16
update evals

miguelg719:
# why
Make it easier to parse/filter/group evals

# what changed
Evals tagged with more granular metadata and error parsing

# test plan

### DIFF
--- a/evals/evals.config.json
+++ b/evals/evals.config.json
@@ -561,12 +561,6 @@
       "categories": [	
         "external_agent_benchmarks"	
       ]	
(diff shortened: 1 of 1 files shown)


evals/evaluator.ts
```diff
@@ -343,6 +334,7 @@ export class Evaluator {
           name: "EvaluationResult",
           schema: EvaluationSchema,
         },
+        // temperature: 1,
```
miguelg719:
```suggestion
```

PR #1142, opened 2025-10-16
Fixing broken links

guergabo:
# why

Broken links. 

# what changed

Fixed broken links + added broken links checker for the future. 

# test plan

This PR.

### DIFF
--- a/docs/basics/observe.mdx
+++ b/docs/basics/observe.mdx
@@ -257,8 +257,8 @@ Extract structured data from observed elements
 Monitor and debug observation performance  
 </Card>
 
-<Card title="Best Practices" icon="star" href="/best-practices/caching">
-Advanced patterns and optimization techniques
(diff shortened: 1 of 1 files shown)


docs/basics/observe.mdx
```diff
@@ -257,7 +257,7 @@ Extract structured data from observed elements
 Monitor and debug observation performance  
 </Card>
 
-<Card title="Best Practices" icon="star" href="/best-practices/best-practices">
+<Card title="Best Practices" icon="star" href="/best-practices/caching">
```
sophie2chance2: I agree that this should go to caching, but lets change the language from Best practices to better reflect that this is going to caching docs

guergabo: Updated to "Optimizations" here since it seems that's the general focus.

PR #1143, opened 2025-10-17
fix: external aisdk client logging

seanmcguire12:
# why
- there was previously no logger parameter in the external aisdk client
# what changed
- added logger param
# test plan
- evals, 
- install from alpha & smoke test

### DIFF
--- a/examples/external_clients/aisdk.ts
+++ b/examples/external_clients/aisdk.ts
@@ -364,40 +364,6 @@ export class AISdkClient extends LLMClient {
       },
     } as T;
 
-    if (this.enableCaching) {
-      this.logger?.({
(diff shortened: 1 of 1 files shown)


examples/external_clients/aisdk.ts
```diff
@@ -82,45 +177,249 @@ export class AISdkClient extends LLMClient {
...
+            type: "object",
+          },
+          response: {
+            value: JSON.stringify(result),
+            type: "object",
+          },
+        },
+      });
+      this.cache.set(cacheOptions, result, options.requestId);
+    }
```
miguelg719:
```suggestion
```

miguelg719 (approved):

PR #1148, opened 2025-10-17
Update CerebrasClient to use ai-sdk instead of OpenAI

clayostroff:
# why
Cerebras integration was out of date for models and not buildable. AI-SDK ships a first‑party Cerebras provider that's now used.

# what changed
* Replace lib/llm/CerebrasClient.ts with an AI‑SDK‑based implementation.
* Update lib/llm/LLMProvider.ts to instantiate the Cerebras AI‑SDK client for Cerebras models.
* Keep old and add new Cerebras models added in types/model.ts.

# test plan
I tested using the form_filling_sensible file and the 2048 tests to test all the Cerebras models.

# notes
Message & tool mapping mirrors the existing AISdkClient example for consistent behavior. No changes to other providers.


JerryWu1234:
test

PR #1153, opened 2025-10-20
ai sdk v5 migration in v3

tkattkat:
# why

we need to upgrade from ai sdk v4 to v5

# what changed

migrated from v4 to v5

# test plan

### DIFF
--- a/packages/evals/package.json
+++ b/packages/evals/package.json
@@ -17,7 +17,6 @@
     "@browserbasehq/orca": "workspace:*",
     "ai": "^5.0.0",
     "@ai-sdk/provider": "^2.0.0",
-    "braintrust": "^0.4.7",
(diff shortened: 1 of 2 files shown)


packages/core/lib/v3/llm/aisdk.ts
```diff
@@ -255,7 +273,10 @@ export class AISdkClient extends LLMClient {
       id: `chatcmpl_${Date.now()}_${Math.random().toString(36).substr(2, 9)}`,
       object: "chat.completion",
       created: Math.floor(Date.now() / 1000),
-      model: this.model.modelId,
+      model:
+        typeof this.model === "string"
+          ? this.model
+          : (this.model as LanguageModelWithId).modelId || "unknown",
```
tkattkat: sus

PR #1157, opened 2025-10-22
Zodv4

tkattkat:
# why

migrate from zod v3 to v4

# what changed

# test plan

### DIFF
--- a/packages/core/examples/agent-custom-tools.ts
+++ b/packages/core/examples/agent-custom-tools.ts
@@ -1,7 +1,7 @@
 /**
  * This example shows how to pass custom tools to stagehand agent (both CUA and non-CUA)
  */
-import { z } from "zod/v3";
+import { z } from "zod";
 import { tool } from "ai";
 import { Stagehand } from "../lib/v3";
 import chalk from "chalk";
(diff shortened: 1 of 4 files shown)


packages/core/examples/custom_client_langchain.ts
```diff
@@ -3,15 +3,17 @@
...
  * You will need to reference the Langchain Client in /external_clients/langchain.ts
  */
-import { z } from "zod/v3";
+import { z } from "zod";
 import { Stagehand } from "../lib/v3";
 import { LangchainClient } from "./external_clients/langchain";
 import { ChatOpenAI } from "@langchain/openai";
 
 async function example() {
+  // @ts-expect-error Type instantiation is excessively deep and possibly infinite
```
miguelg719, line 12: does this fail without this?

packages/core/examples/v3/targetedExtract.ts
miguelg719, line 1: revert?

packages/core/examples/custom_client_langchain.ts
tkattkat: Yea

packages/core/examples/v3/targetedExtract.ts
tkattkat: oops was testing and accidentally commited, reverting

PR #1158, opened 2025-10-22
add missing fields to agent result

tkattkat:
# why

we are missing pageurl, and timestamp in all agent results, and the playwright arguments from stagehand agent results 

# what changed

added pageurl, and timestamp in all agent results, and the playwright arguments to stagehand agent results

# test plan

tested locally

### DIFF
--- a/packages/core/lib/v3/agent/utils/actionMapping.ts
+++ b/packages/core/lib/v3/agent/utils/actionMapping.ts
@@ -1,11 +1,5 @@
 import { AgentAction } from "../../types/public/agent";
-
-interface ActionMappingOptions {
-  toolCallName: string;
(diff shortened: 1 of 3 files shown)


packages/core/lib/v3/agent/utils/actionMapping.ts
```diff
@@ -0,0 +1,106 @@
+import { AgentAction } from "../../types/public/agent";
+
+export interface ActionMappingOptions {
+  toolCallName: string;
+  toolResult: unknown;
+  args: Record<string, unknown>;
```
seanmcguire12: can you put this in either a new or existing private types file?

PR #1167, opened 2025-10-24
patch cua model handling

tkattkat:
# why
when a model is defined in stagehand cofnig, but not the agent instantiation, an error is thrown 

# what changed

now, if no model is passed in agent instantiation, we use the one from stagehand config. 

# test plan

### DIFF
--- a/packages/core/lib/v3/v3.ts
+++ b/packages/core/lib/v3/v3.ts
@@ -1275,7 +1275,7 @@ export class V3 {
 
       if (!isCua) {
         throw new Error(
-          "To use our computer use agent, please provide a CUA model in the agent constructor or stagehand config. Try one of our supported CUA models: " +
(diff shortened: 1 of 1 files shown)


packages/core/lib/v3/v3.ts
```diff
@@ -1266,14 +1266,16 @@ export class V3 {
...
+        modelName: this.modelName,
+        ...this.modelClientOptions,
+      };
+
+      const { modelName, isCua, clientOptions } = resolveModel(modelToUse);
 
       if (!isCua) {
         throw new Error(
-          "Model is not a CUA model. Try one of the following: " +
+          "To use our computer use agent, please provide a CUA model in the agent constructor or stagehand config. Try one of our supported CUA models: " +
```
seanmcguire12: to use 'the' computer use agent

seanmcguire12 (approved):

PR #1173, opened 2025-10-25
update viewport on newly created pages

miguelg719:
test with anthropic/openai models on the `cua-example.ts` and see them work on this branch vs. unable to click on v3


miguelg719:
@greptileai

PR #1212, opened 2025-10-31
[fix] Stagehand Agent param is instructions instead of systemPrompt

bennycortese:
Stagehand agent no longer directly takes in instructions. It takes in a systemPrompt instead, I noticed this while running npx create-browser-app and seeing the error appear

Fixed this only in the v3 documentation here, but if we are breaking backwards compatibility (I assume probably not), then the following v2 doc files should also be updated:

agent.mdx
agent-fallbacks.mdx
computer-use.mdx
mcp-integrations.mdx
ai-rules.mdx
quickstart.mdx

### DIFF
--- a/packages/core/examples/mcp.ts
+++ b/packages/core/examples/mcp.ts
@@ -31,7 +31,7 @@
 //       provider: "anthropic",
 //       // For Anthropic, use claude-sonnet-4-20250514 or claude-3-7-sonnet-latest
 //       model: "claude-sonnet-4-20250514",
-//       systemPrompt: `You are a helpful assistant that can use a web browser.
+//       instructions: `You are a helpful assistant that can use a web browser.
 //       You are currently on the following page: ${page.url()}.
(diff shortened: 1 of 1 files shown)


packages/core/examples/mcp.ts
```diff
@@ -31,7 +31,7 @@
 //       provider: "anthropic",
 //       // For Anthropic, use claude-sonnet-4-20250514 or claude-3-7-sonnet-latest
 //       model: "claude-sonnet-4-20250514",
-//       instructions: `You are a helpful assistant that can use a web browser.
+//       systemPrompt: `You are a helpful assistant that can use a web browser.
```
Kylejeong2: our MCP server is not updated yet but will be soon

miguelg719:
```suggestion
//       instructions: `You are a helpful assistant that can use a web browser.
```

PR #1213, opened 2025-10-31
[fix] stagehand.agent isn't using the correct model defaults

bennycortese:
So when I ran stagehand.agent out of the box, I kept getting the model set to gpt-4.1-mini by default. This seems to be because the Vercel AI SDK is used as the default for logging, and the model  isn't ever getting set to openai/gpt-4.1-mini. This should fix both issues and hopefully fix default model configs for all providers

### DIFF
--- a/.changeset/whole-goats-clap.md
+++ b/.changeset/whole-goats-clap.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": minor
+---
+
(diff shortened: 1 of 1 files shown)


stefanbinoj:
IMO it's hardcoded (for some reason) [here](https://github.com/browserbase/stagehand/blob/main/packages/core/lib/v3/v3.ts#L68)

PR #1229, opened 2025-11-03
adjust extract tool + agent response

tkattkat:
# why

- A new selector arg was added in v3 migration, resulting in a higher failure rate with agent 
- fallback to z.any was replaced with throwing an error in v3 migration, resulting in a higher failure rate 

# what changed

- removed selector arg 
- changed the schema eval logic back to its original behavior of falling back to z.any 
- exposed extract result in the agent response so it can be easily parsed, and utilized 
# test plan

### DIFF
--- a/packages/core/lib/v3/agent/tools/v3-extract.ts
+++ b/packages/core/lib/v3/agent/tools/v3-extract.ts
@@ -25,8 +25,7 @@ export const createExtractTool = (
   logger?: (message: LogLine) => void,
 ) =>
   tool({
-    description:
-     `Extract structured data from the current page based on a provided schema.
+    description: `Extract structured data from the current page based on a provided schema.
     
     USAGE GUIDELINES:
     - Keep schemas MINIMAL - only include fields essential for the task


packages/core/lib/v3/agent/utils/actionMapping.ts
```diff
@@ -87,14 +87,24 @@ function mapFillFormToolResult(
...
     taskCompleted:
       toolCallName === "close" ? (args?.taskComplete as boolean) : false,
     ...args,
   };
+
+  // Spread the output from the tool result if it exists, exclude ariaTree tool result as it is very large and unnecessary
+  // todo : add better typing for every tool to avoid type casting
+  if (toolCallName !== "ariaTree" && toolResult) {
+    const { output } = toolResult as { output: unknown };
+    Object.assign(action, output);
```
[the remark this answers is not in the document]
tkattkat: ur wrong output always exists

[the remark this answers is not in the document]
tkattkat: it will always be present brother

packages/core/lib/v3/agent/tools/v3-extract.ts
```diff
@@ -28,24 +25,38 @@ export const createExtractTool = (
...
+       instruction: "extract all product names and prices"
+       schema: "z.object({ products: z.array(z.object({ name: z.string(), price: z.number() })) })"`,
     inputSchema: z.object({
-      instruction: z.string().optional(),
+      instruction: z.string(),
       schema: z
         .string()
         .optional()
         .describe("Zod schema as code, e.g. z.object({ title: z.string() })"),
-      selector: z.string().optional(),
```
miguelg719, line 39: what's the source of increased failures if selector is optional? is the model hallucinating selectors (invalid/inaccurate)?

packages/core/lib/v3/agent/tools/v3-extract.ts
```diff
@@ -28,24 +25,38 @@ export const createExtractTool = (
...
+    description: `Extract structured data from the current page based on a provided schema.
+    
+    USAGE GUIDELINES:
+    - Keep schemas MINIMAL - only include fields essential for the task
+    - IMPORANT: only use this if explicitly asked for structured output. In most scenarios, you should use the aria tree tool over this. 
+    - If you need to extract a link, make sure the type defintion follows the format of z.string().url()
+    EXAMPLES:
+    1. Extract a single value:
+       instruction: "extract the product price"
+       schema: "z.object({ price: z.number()})"
```
miguelg719, line 37: do schemas still need to be wrapped in z.obejct for parsing on v3?

packages/core/lib/v3/agent/tools/v3-extract.ts
tkattkat: the specific issue we saw was from the evaluateZodSchema being changed to throw an error instead of return z.any. When testing though I did notice selectors being hallucinated as well so I removed as it was not there prior

packages/core/lib/v3/agent/tools/v3-extract.ts
tkattkat: They do not have to be no. doing so without will just return it not as an object

PR #1235, opened 2025-11-05
make navigation methods return a `Response` object

seanmcguire12:
# why
- to mirror playwright's `page.goto()` return type
# what changed
this PR makes page navigation functions (`page.goto()`, `page.goBack()`, `page.goForward()`) return a stagehand native `Response`, similar to playwright's `Response`

`Response` implements the following functions:
- `response.url()`
  - returns the URL from the associated navigation
- `response.status()`
  - returns the http status code
- `response.statusText()`
  - returns the status text from the response
- `response.ok()`
  - whether the response was a 2xx
- `response.frame()`
  - returns the `Frame` object that the originally triggered the request
- `response.body()`
  - returns the raw response body from Chrome DevTools Protocol
- `response.text()`
  - decodes & returns the response body as UTF-8 text
- `response.json()`
  - parses the response body as JSON and throws if parsing fails
- `response.finished()`
- `response.fromServiceWorker()`
- `response.securityDetails()`
  - returns security details metadata when provided by the browser
- `response.serverAddr()`
  - returns the associated server address for the navigation if available (IP & port)
- `response.allHeaders()`
  - returns all headers including those surfaced via `Network.responseReceivedExtraInfo` (e.g. set-cookie, raw TLS headers); values are reported exactly as Chrome provides them, without case normalization
- `response.headers()`
  - returns the immediate response headers with keys normalized to lowercase and duplicate entries collapsed into a comma-separated string; fast, case-insensitive lookups
- `response.headerValue(name: string)`
  - convenience wrapper that returns the joined string for a specific header (or null if absent); equivalent to response.headers()[name.toLowerCase()] but respects extra-info headers when available
- `response.headerValues(name: string)`
  - returns an array containing each header value individually, preserving duplicates (ideal for set-cookie and other repeatable fields)
- `response.headersArray()`
  - returns the ordered list of `{ name, value }` pairs with original casing preserved; best when you need the
  verbatim wire representation or order-sensitive processing
- `response.applyExtraInfo()`
  - internal helper. updates the request with new info if any becomes available via a `Network.ResponseReceivedExtraInfoEvent`
- `response.markFinished()`
  - internal helper for resolving the `finished()` promise. invoked by the navigation tracker when Chrome emits a
  `Network.loadingFinished` or `Network.loadingFailed` event

adds an internal `NavigationResponseTracker` class which handles creation of the `Response`, registering and disposing of network related event listeners. 
- `NavigationResponseTracker` does the following: 
  - subscribes to `Network.responseReceived`, `Network.responseReceivedExtraInfo`, `Network.loadingFinished`, and `Network.loadingFailed` on the main session as soon as a navigation command starts
  - filters events down to the current navigation by checking the command id, main-frame id, resource type (Document), and the loader id Chrome returns from `Page.navigate`
  - buffers response payloads (and extra-info headers) until it can confidently match the main-document response, then instantiates the corresponding `Response`
  - handles scheme-only navigations (data:, about:) by short-circuiting and resolving with null
  - keeps the `Response` object up to date: when Chrome later sends extra headers (like set-cookie) they are fed back into the same `Response`, and when Chrome says the navigation succeeded or failed it settles the `response.finished()` promise so callers get the correct result
  - gives `page.goto()` a single promise (`navigationCompleted()`) that resolves to the right `Response` (or null) once the matching document response shows up, while LifecycleWatcher continues handling the usual “wait for load/domcontentloaded/networkidle” timing
  - detaches all listeners via `dispose()` once navigation completes, preventing leaks across subsequent navigations
# test plan
- added unit tests
- evals

### DIFF
--- a/packages/core/lib/v3/understudy/response.ts
+++ b/packages/core/lib/v3/understudy/response.ts
@@ -1,5 +1,5 @@
 /**
- * StagehandResponse
+ * Response
  * -----------------
  *
  * This module implements a Playwright-inspired response wrapper that exposes
--- a/packages/core/lib/v3/understudy/navigationResponseTracker.ts
+++ b/packages/core/lib/v3/understudy/navigationResponseTracker.ts
@@ -7,7 +7,7 @@
  * related APIs. The tracker listens for `Network.responseReceived` events that
  * correspond to the targeted document navigation, handles loader-id churn that
  * arises from redirects or preloading, and enriches the resulting
- * `StagehandResponse` with extra header information. It also observes
+ * `Response` with extra header information. It also observes
  * `Network.loadingFinished` / `Network.loadingFailed` to fulfil the
  * `response.finished()` contract exposed to consumers.
  */
@@ -210,13 +210,13 @@ export class NavigationResponseTracker {
     if (!event || !event.requestId) return;
     // Only the tracked document request should toggle the response state
     if (event.requestId !== this.selectedRequestId) return;
-    // Surface Chrome's failure text through StagehandResponse.finished()
+    // Surface Chrome's failure text through response.finished()
     const errorText = event.errorText || "Navigation request failed";
     this.selectedResponse?.markFinished(new Error(errorText));
   }
 
   /**
-   * Create the `StagehandResponse` wrapper for the chosen document response and
+   * Create the `Response` wrapper for the chosen document response and
    * resolve awaiting consumers. Subsequent events flesh out the header/body
    * helpers and mark the request as finished.
    */


seanmcguire12:
@greptileai

seanmcguire12:
@greptileai

packages/docs/v3/references/response.mdx
```diff
@@ -0,0 +1,250 @@
...
+
+<CardGroup cols={1}>
+<Card title="Navigation" icon="compass" href="/v3/references/page">
+  See how pages expose StagehandResponse objects from navigation methods
+</Card>
+</CardGroup>
+
+## Overview
+
+`StagehandResponse` mirrors Playwright’s [Response](https://playwright.dev/docs/api/class-response) interface and is returned from Stagehand navigation helpers such as `page.goto()`, `page.reload()`, `page.goBack()`, and `page.goForward()`. It provides a convenient way to inspect the HTTP metadata associated with a navigation, retrieve the response body on demand, and monitor when the underlying request finishes.
```
miguelg719: double check `page.reload()`

packages/core/lib/v3/understudy/response.ts
```diff
@@ -0,0 +1,349 @@
+/**
+ * StagehandResponse
```
miguelg719: response

PR #1241, opened 2025-11-07
Return response on page.goto api mode

miguelg719:
# Why

  When using Stagehand in API mode (connected to a remote browser via the Stagehand API), the
  `page.goto()` method was not returning a `Response` object, unlike the local CDP implementation. This
  meant users couldn't access important navigation response information like status codes, headers, and
  URLs when using API mode.

  # What Changed

  - Modified `StagehandAPIClient.goto()` to return `SerializableResponse | null` instead of `void`
  - Added `SerializableResponse` interface in `lib/v3/types/private/api.ts` to represent Response data
  that can be transmitted over the API
  - Created `Response.fromSerializable()` static method to reconstruct `Response` objects from serialized
   API data
  - Updated `Page.goto()` to handle the API response and create proper `Response` objects when in API
  mode
  - Added `isSerializableResponse()` type guard to validate API responses

# Test plan
- [x] tested with `page-goto-response.ts`

### DIFF
--- a/packages/core/lib/v3/api.ts
+++ b/packages/core/lib/v3/api.ts
@@ -20,7 +20,6 @@ import {
   AgentResult,
   ExtractResult,
   LogLine,
-  Response as StagehandNavigateResponse,
   StagehandMetrics,
   StagehandAPIError,
   StagehandAPIUnauthorizedError,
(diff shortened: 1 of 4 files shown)


seanmcguire12:
@greptileai

Issue #1204, opened 2025-10-30
DeepSeek API Compatibility Issue: JSON Format Requirement and Schema Mismatch


mgshao:
## Problem Description

When using DeepSeek API (`deepseek/deepseek-chat`) with Stagehand, there are two related issues:

1. **Missing "json" keyword error**: When using DeepSeek API, if the prompt doesn't explicitly contain the word "json", the API returns an error: `Prompt must contain the word 'json' in some form to use 'response_format' of type 'json_object'`

2. **Schema mismatch error**: When adding "json" to the prompt to satisfy DeepSeek's requirement, the model returns a JSON structure that doesn't match Stagehand's expected schema. DeepSeek returns a format like:
   ```json
   {
     "action": "fill",
     "element": {"id": "...", "role": "...", "name": "..."},
     "argument": "Cota",
     "twoStep": false
   }
   ```
   But Stagehand expects:
   ```json
   {
     "elementId": "string (required)",
     "description": "string (required)",
     "method": "string (required)",
     "arguments": ["array (required)"]
   }
   ```

This creates a catch-22 situation: without "json" in the prompt, DeepSeek API rejects the request; with "json" in the prompt, the response format doesn't match Stagehand's schema.

## Environment

- **Stagehand version**: `@browserbasehq/stagehand@^3.0.0`
- **Node.js version**: v22.2.0
- **DeepSeek API endpoint**: `https://api.deepseek.com/v1`
- **Model**: `deepseek/deepseek-chat`

## Code to Reproduce

```typescript
import "dotenv/config";
import { Stagehand } from "@browserbasehq/stagehand";
import { z } from "zod/v3";

async function main() {
  const stagehand = new Stagehand({
    env: "LOCAL",
    model: "deepseek/deepseek-chat"
  });

  await stagehand.init();
  const page = stagehand.context.pages()[0];

  await page.goto("https://www.baidu.com");

  await stagehand.act("fill the search input with the text 'hello'");
  await stagehand.act("click the search button");

  // Extract structured data
  const title = await stagehand.extract("extract the page title", z.string());

  console.log(title);
  await stagehand.close();
}

main().catch((err) => {
  console.error(err);
  process.exit(1);
});
```

### Error 1: Missing "json" keyword
```
APICallError [AI_APICallError]: Prompt must contain the word 'json' in some form to use 'response_format' of type 'json_object'.
    at ...
  url: 'https://api.deepseek.com/v1/chat/completions',
  requestBodyValues: {
[...]

PR #1242, opened 2025-11-07
Add DeepseekAIClient implementation and integrate with LLMProvider

JerryWu1234:
# what changed
- Added an optional `manual` parameter to the `V3Evaluator` constructor and propagated it during initialization to enable explicit manual-mode configuration.
- Updated `LLMProvider` to accept the `manual` flag, revising model-provider resolution and client selection logic; registered `deepseek/deepseek-reasoner` and pruned legacy Deepseek entries.
- Ensured manual mode bypasses automatic inference so that model parsing follows the caller’s explicit configuration.

#1204 related to ai bug https://github.com/vercel/ai/issues/7913
- The vercel/ai issue #7913 reports inference errors when models are auto-resolved; this change exposes a manual override to avoid incorrect mappings and to validate fixes quickly.

# test plan
- Initialize `V3Evaluator`/`LLMProvider` locally with `manual=true`; confirm instantiation succeeds and automatic mapping remains disabled.
- Run the default flow with `manual=false` or omitted to verify legacy behavior is unchanged.
- Exercise `deepseek/deepseek-reasoner` end to end and confirm the provider selects the correct client and returns responses.

# why
- Allows precise control of the model provider when diagnosing vercel/ai #7913, preventing auto-mapping from leading to inference failures.
- Provides a safe manual switch for experimental or self-hosted models, reducing regression risk.

# what changed
- `packages/core/lib/v3Evaluator.ts`: constructor accepts `manual`, defaults to previous behavior, forwards the option to `LLMProvider`.
- `packages/core/lib/v3/llm/LLMProvider.ts`: constructor plus `getModelProvider`/`getClient` honor manual mode; updated model registry for the new Deepseek entry.
- `packages/core/lib/v3/types/public/model.ts`: added `manual?: boolean` to `V3Options`; appended `deepseek/deepseek-reasoner` to `AvailableModel`.

# test plan
- Start Stagehand examples or integration scripts with `manual=true/false`, inspect `LLMProvider` logs to verify the proper branch runs.
- Supply a model name that should only resolve under manual mode and confirm it is not overwritten by auto-mapping.
- Run existing unit/integration suites (if available) to ensure backward compatibility.

### DIFF
--- a/packages/core/lib/v3/v3.ts
+++ b/packages/core/lib/v3/v3.ts
@@ -136,7 +136,7 @@ export class V3 {
   }
   private _onCdpClosed = (why: string) => {
     // Single place to react to the transport closing
-    this._immediateShutdown(`CDP transport closed: ${why}`).catch(() => {});
+    this._immediateShutdown(`CDP transport closed: ${why}`).catch(() => { });
   };
   public readonly experimental: boolean = false;
   public readonly logInferenceToFile: boolean = false;
@@ -173,6 +173,7 @@ export class V3 {
 
   constructor(opts: V3Options) {
     V3._installProcessGuards();
+
     this.externalLogger = opts.logger;
     this.verbose = opts.verbose ?? 1;
     this.instanceId =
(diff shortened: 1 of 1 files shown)


shrey150:
hey @JerryWu1234 , to avoid supporting an additional LLM client, we'll move towards the approach outlined in #1455

PR #1253, opened 2025-11-10
fix: page not found when connecting to existing browser

seanmcguire12:
# why
- when connecting to an existing browser, stagehand was sometimes unable to update its object model based on the current state of the browser
- for example, if there was an existing browser with 5 pages open, and stagehand connects to it, logging `stagehand.context.pages[].length` should show `5`, but this was not always the case.
# what changed
- changed the order of operations in the connection logic such that `enableAutoAttach` is called after event listeners are registered
- also added `waitForInitialTopLevelTargets()` with a default timeout of 3 seconds to ensure that stagehand has enough time for targets to be registered
# test plan
- added a unit test which 
  - creates a browserbase session, 
  - opens 5 pages, 
  - connects to the session with a new stagehand instance
  - confirms that the new stagehand instance correctly reports that there are 5 pages

### DIFF
--- a/packages/core/lib/v3/understudy/context.ts
+++ b/packages/core/lib/v3/understudy/context.ts
@@ -120,7 +120,7 @@ export class V3Context {
       v3Logger({
         category: "ctx",
         message: "Timed out waiting for existing top-level targets to attach",
-        level: 1,
+        level: 2,
         auxiliary: {
           remainingTargets: {
             value: JSON.stringify(Array.from(pending)),


seanmcguire12:
@greptileai

miguelg719 (approved):

packages/core/lib/v3/understudy/context.ts
```diff
@@ -101,6 +100,37 @@ export class V3Context {
...
+        }
+      }
+      if (!pending.size) return;
+      await new Promise((r) => setTimeout(r, 25));
+    }
+    if (pending.size) {
+      v3Logger({
+        category: "ctx",
+        message: "Timed out waiting for existing top-level targets to attach",
+        level: 1,
```
miguelg719:
```suggestion
        level: 2,
```

PR #1261, opened 2025-11-13
Update Mintlify docs to use dynamic redirects

shrey150:
## What changed
Mintlify dynamic redirects weren't working at the time of V3's launch, meaning we statically saved them.

## Why
Now, we've updated them accordingly to follow best practices.

## Test plan
Ran `mintlify dev` locally and ensured all paths redirect correctly

### DIFF
--- a/packages/docs/docs.json
+++ b/packages/docs/docs.json
@@ -294,6 +294,18 @@
       "destination": "/v3/integrations/mcp/:slug*"
     },
     {
+      "source": "/integrations/crew-ai/:slug*",
+      "destination": "/v3/integrations/crew-ai/:slug*"
+    },
+    {
+      "source": "/integrations/langchain/:slug*",
+      "destination": "/v3/integrations/langchain/:slug*"
+    },
+    {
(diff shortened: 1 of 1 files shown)


packages/docs/docs.json
```diff
@@ -274,184 +274,44 @@
...
-      "destination": "/v3/first-steps/ai-rules"
+      "source": "/best-practices/:slug*",
+      "destination": "/v3/best-practices/:slug*"
     },
     {
-      "source": "/basics/act",
-      "destination": "/v3/basics/act"
+      "source": "/integrations/mcp/:slug*",
+      "destination": "/v3/integrations/mcp/:slug*"
     },
```
derekmeegan, line 295: can you check if we cxan have 1 rule that is /integrations/:slug* and that will suffice for these multiple rules?

shrey150: just tested, it 404s without the explicit subpath slug

derekmeegan: frick

derekmeegan:
LGTM

PR #1264, opened 2025-11-13
feat: adding gpt 5.1 to stagehand

Kylejeong2:
# why

we want to benchmark/use the sota models in stagehand

# what changed

added gpt 5.1 to available models, setting reasoning to low for gpt 5.1 specifically since it cannot be set to minimal. 

# test plan

### DIFF
--- a/.changeset/bright-meals-add.md
+++ b/.changeset/bright-meals-add.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+feat: adding gpt 5.1 to stagehand


packages/core/lib/v3/llm/LLMProvider.ts
```diff
@@ -73,6 +73,7 @@ const modelToProviderMap: { [key in AvailableModel]: ModelProvider } = {
   "gpt-4o-mini": "openai",
   "gpt-4o-2024-08-06": "openai",
   "gpt-4.5-preview": "openai",
+  "gpt-5.1-2025-11-13": "openai",
```
miguelg719: not needed here

packages/core/lib/v3/types/public/model.ts
```diff
@@ -39,6 +39,7 @@ export type AvailableModel =
   | "gpt-4o-mini"
   | "gpt-4o-2024-08-06"
   | "gpt-4.5-preview"
+  | "gpt-5.1-2025-11-13"
```
miguelg719: same not needed

packages/core/lib/v3/llm/aisdk.ts
```diff
@@ -138,7 +139,7 @@ export class AISdkClient extends LLMClient {
             ? {
                 openai: {
                   textVerbosity: "low", // Making these the default for gpt-5 for now
-                  reasoningEffort: "minimal",
+                  reasoningEffort: isGPT51 ? "low" : "minimal",
```
miguelg719, line 142: any reason to make 5.1 low vs 5 minimal? is it faster?

packages/core/lib/v3/llm/LLMProvider.ts
Kylejeong2: is this for like the openai client? can remove since we're trying to move away from this

packages/core/lib/v3/llm/aisdk.ts
Kylejeong2: its a weird syntax change, you have to set to low because the model doesn't accept minimal as a param

packages/core/lib/v3/types/public/model.ts
miguelg719:
```suggestion
```

packages/evals/taskConfig.ts
```diff
@@ -32,6 +32,7 @@ const ALL_EVAL_MODELS = [
   "gpt-4o-mini",
   "gpt-4o",
   "gpt-4.5-preview",
+  "gpt-5.1-2025-11-13",
```
miguelg719:
```suggestion
```

packages/core/lib/v3/llm/aisdk.ts
Kylejeong2: @miguelg719 

can also apparently set the reasoning to none 

https://platform.openai.com/docs/guides/latest-model

PR #1267, opened 2025-11-14
fix: `setInputFiles()` not working on Browserbase

seanmcguire12:
# why
- file uploads were failing on Browserbase. this is because `setInputFiles()` was not serializing the file payload
# what changed
changed business logic of `setInputFiles()` to: 
- call `normalizeInputFiles()`, which returns a serializable/base64 encoded file
- check if the browser is remote
- if remote, `assignFilesViaPayloadInjection()` is called, which then calls `assignFilePayloadsToInputElement()` (browser-side JS) and reconstructs the base64 back into a file, and directly assigns the reconstructed files to the `input` element
- if not remote (local), then the files are written to temp dirs, and those dirs are passed into `DOM.setInputFiles` (CDP command)
# test plan
- added additional tests for various file input shapes
- regression evals

### DIFF
--- a/packages/core/lib/v3/understudy/fileUploadUtils.ts
+++ b/packages/core/lib/v3/understudy/fileUploadUtils.ts
@@ -5,8 +5,8 @@ import { StagehandInvalidArgumentError } from "../types/public/sdkErrors";
 import {
   SetInputFilesArgument,
   SetInputFilePayload,
-  NormalizedFilePayload,
 } from "../types/public/locator";
+import { NormalizedFilePayload } from "../types/private/locator";
 
 const DEFAULT_MIME_TYPE = "application/octet-stream";
 


packages/core/lib/v3/types/public/locator.ts
```diff
@@ -0,0 +1,23 @@
...
+  lastModified?: number;
+}
+
+export type SetInputFilesArgument =
+  | string
+  | string[]
+  | SetInputFilePayload
+  | SetInputFilePayload[];
+
+export interface NormalizedFilePayload {
```
seanmcguire12: this doesn't need to be public

seanmcguire12:
@greptileai

tkattkat (approved):

PR #1273, opened 2025-11-16
fix: don't reload OOPIFs

seanmcguire12:
# why
- we needed to reload OOPIFs so that our shadow-root piercing JS was able to intercept the `attachShadow()` call
- this is wasteful, and can cause unpredictable behaviour
# what changed
- swapped iframe reload logic for new logic that clones, and replaces shadow root nodes
- this triggers the patched `attachShadow()` function
# test plan
- existing unit tests & evals should sufficiently cover shadow root piercing


seanmcguire12:
@greptileai

madcio-rgb (commented):
T

PR #1282, opened 2025-11-18
add support for zod v4

tkattkat:
Why 

Adds support for zod 4 while maintaining backwards compatibility with zod 3 

What changed 

- Introduced a zodCompat layer so Stagehand speaks both Zod v3 and v4 natively. Every place that used to import zod/v3, zod-to-json-schema, or poke at schema internals now routes through the same helpers (StagehandZodSchema, InferStagehandSchema, toJsonSchema, etc.). That keeps existing apps on v3 working while letting us upgrade to the v4 ecosystem.
- Swapped all examples, eval tasks, tests, and LLM clients over to standard zod imports, updated dependencies to accept 3.25.76 || 4.1.8, and dropped zod-to-json-schema. Tool plumbing (Anthropic/Google CUA) and custom clients now produce the right JSON Schema regardless of which Zod flavor the user passes in.
- Tightened typings so structured LLM calls return { data, usage } directly no more as casts throughout inference or the samples.

Manual testing
To make sure nothing regressed, I ran:
Custom OpenAI client example (schema conversion + structured output).
Standard extract (including URL injection path).
Google computer-use agent with a custom tool.
Anthropic computer-use agent with a custom tool.
All behaved as before

### DIFF
--- a/.changeset/add-browserbase-url-accessors.md
+++ b/.changeset/add-browserbase-url-accessors.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+Add Browserbase session URL and debug URL accessors
--- a/.changeset/nice-socks-bow.md
+++ b/.changeset/nice-socks-bow.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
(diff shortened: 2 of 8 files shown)


seanmcguire12:
@greptileai

seanmcguire12:
@greptileai

PR #1290, opened 2025-11-19
Add public API contract tests

monadoid:
# why
We need to create stubs for the client sdk, and this allows us to write tests so that we can confirm that our stub signatures have full parity.

# what changed
Added full contract tests which check just the public facing API *types*
https://vitest.dev/guide/testing-types

Also made it so that all types from `/public/types` are explicitly re-exported through index.ts, so that we can have a single place to review all public exports. 

These tests might be a bit more in-depth then we ultimately want - we are checking function signatures for stuff like the vercel AI client, which might be more brittle than we want currently, but we will likely be removing some of these public exports anyways with the canonical work.

# test plan
You can run `pnpm --filter @browserbasehq/stagehand test:vitest`

### DIFF
--- a/packages/core/lib/v3/index.ts
+++ b/packages/core/lib/v3/index.ts
@@ -1,69 +1,8 @@
 export { V3 } from "./v3";
 export { V3 as Stagehand } from "./v3";
 
+export * from "./types/public";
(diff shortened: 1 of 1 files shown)


monadoid:
@greptileai

seanmcguire12 (approved):

PR #1292, opened 2025-11-20
Add stubs

monadoid:
# why
We want stubs for our thin TS sdk so we can start carving out the network boundary and see what RPC events we'll want to implement.

# what changed
Added a /sdk package and created stubs for all public exports.
# test plan
I updated our contract tests to test both the core and the sdk, so that we can prove that we have full parity.


monadoid:
@greptileai

PR #1293, opened 2025-11-20
[CI] install stable chrome

seanmcguire12:
# why
- chrome launcher was installing canary, which broke some unit tests
# what changed
- added install step for a stable chrome version in `ci.yml`
- also lowered worker count for `shadow-frame.spec.ts`

### DIFF
--- a/packages/core/lib/v3/tests/xpath-for-location-deep.spec.ts
+++ b/packages/core/lib/v3/tests/xpath-for-location-deep.spec.ts
@@ -23,7 +23,7 @@ test.describe("resolveNodeForLocationDeep", () => {
 
     await page.goto(
       "https://browserbase.github.io/stagehand-eval-sites/sites/iframe-hn/",
-      {waitUntil: "networkidle"}
+      { waitUntil: "networkidle" },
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -202,14 +202,29 @@ jobs:
           node-version: "20"
           cache: "pnpm"
 
+      - name: Install stable Chromium
+        run: |
+          set -euo pipefail
+          CHROME_VERSION=$(curl -s https://googlechromelabs.github.io/chrome-for-testing/last-known-good-versions.json | jq -r '.channels.Stable.version')
+          DOWNLOAD_URL="https://storage.googleapis.com/chrome-for-testing-public/${CHROME_VERSION}/linux64/chrome-linux64.zip"
+          INSTALL_DIR="${RUNNER_TEMP}/chrome-stable"
```
pirate, line 210: can save a bit of time on all tests by caching this dir in github actions cache but not critical for now (using OS, arch, and chrome version in cache key)

pirate (approved):

PR #1300, opened 2025-11-21
[feat]: add `context.addInitScript()`

seanmcguire12:
# why
- adds a public facing API to inject scripts into new pages before existing JS is run
# what changed
- added `context.addInitScripts()` which calls `Page.addScriptToEvaluateOnNewDocument` whenever we attach to a page target
# test plan
- added unit tests which check that the injected scripts are run before existing scripts


seanmcguire12:
@greptileai

PR #1308, opened 2025-11-25
[fix]: make `page.evaluate()` use main world

seanmcguire12:
# why
- currently, we add the init scripts to the main world
- this is problematic, because `page.evaluate()` uses the `v3-world`
- this means that you can't easily evaluate the script you add with `addInitScript()`
# what changed
- makes sure that `page.evaluate()` uses the main world
# test plan
- added a test which calls `page.evaluate()` on a script added via `context.addInitScript()`


seanmcguire12:
@greptielai

seanmcguire12:
@cubic-dev-ai review this PR

PR #1328, opened 2025-12-01
Add fastify + connectrpc for SDK clients

monadoid:
# why
We want a fastify server & connectrpc set up so that we can create a client/server split

# what changed
Added fastify server + connectrpc + buf and an example "ping" handler.

`pnpm dev` was added - this starts the fastify server.
`pnpm rpc:generate` was added - this runs the codegen when we change the schema (we'll want to have 
# test plan
- Added vitest tests to spin up a test server and test the basic "ping" handler works and will reject if invalid data is passed
- Added an `pnpm rpc:check` command which is run in CI, so if someone changes the proto but forgets to run `rpc:generate`, the mismatch will be caught.

---
## Summary by cubic
Sets up a Fastify server with ConnectRPC and Buf codegen, and adds a typed Ping RPC using google.protobuf.Timestamp. Adds validation via ConnectRPC and tests for RTT, latency, and clock offset.

## Why:
- Kick off a clean client/server split with a typed RPC layer.
- Provide a simple Ping endpoint for time sync.

## What:
- Fastify server with ConnectRPC and validate interceptor.
- Proto: StagehandPingService.Ping using Timestamp; TS generated via protoc-gen-es.
- Handler echoes client_send_time and sets server_send_time with timestampNow().
- Scripts: dev, rpc:deps, rpc:lint, rpc:generate.
- Ignore generated code in packages/core/gen for Prettier/ESLint.

## Test Plan:
- [x] Vitest spins a Fastify + ConnectRPC server on an ephemeral port; asserts client_send_time echo and server_send_time >= client time.
- [x] Invalid timestamp structure is rejected with Code.Internal via validate interceptor.
- [x] Manual: pnpm dev to start server; call Ping with any Connect client and verify RTT/latency/offset.

<sup>Written for commit 159b27fb8d00a5292c7324b07728c04146ac3bb0. Summary will update automatically on new commits.</sup>


packages/core/proto/stagehand/v1/ping.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+package stagehand.v1;
+import "buf/validate/validate.proto";
+
+message PingRequest {
+  string message = 1;
```
pirate: lets include client timestamp in the request and response as well so we can calculate transport layer latency by doing RTT/2

packages/core/proto/stagehand/v1/ping.proto
```diff
@@ -0,0 +1,19 @@
...
+}
+
+message PingResponse {
+   string message = 1 [
+     (buf.validate.field).string.min_len = 1,
+     (buf.validate.field).string.max_len = 250
+   ];
```
pirate: also include current server datetime in the response so we can order timestamped events properly based on server time

PR #1330, opened 2025-12-02
fix: `act`, `extract`, and `observe` not respecting `timeout` param

seanmcguire12:
# why
- async functions invoked by act, extract, and observe all continued to run even after the timeout was reached
# what changed
- this PR introduces a time remaining check mechanism which runs between each major IO operation inside each of the handlers
- this ensures that user defined timeout are actually respected inside of act, extract, and observe
# test plan
- added tests to confirm that internal async functions do not continue running after the timeout is reached

---
## Summary by cubic
Fixes act, extract, and observe to truly honor the timeout parameter with step-wise guards that abort early and return clear errors. Deterministic actions now use the same guard path in v3.

- **Bug Fixes**
  - Added createTimeoutGuard and specific ActTimeoutError, ExtractTimeoutError, and ObserveTimeoutError (exported).
  - Replaced Promise.race with per-step checks across snapshot capture, LLM inference, action execution, and self-heal retries.
  - Enforced per-step timeouts in ActHandler.takeDeterministicAction; metrics unchanged.
  - Wired v3 deterministic actions to pass a timeout guard; shadow DOM and unsupported actions behavior unchanged.

<sup>Written for commit d6bbfb826dbf2d3c3424e9310b82f59410fdfb2d. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/core/tests/timeout-handlers.test.ts
+++ b/packages/core/tests/timeout-handlers.test.ts
@@ -1,63 +1,45 @@
-// Vitest supplies the BDD helpers plus the mock factory we rely on in this file.
 import { beforeEach, describe, expect, it, vi } from "vitest";
-// Import the concrete handlers so we can exercise their logic in isolation.
 import { ActHandler } from "../lib/v3/handlers/actHandler";
 import { ExtractHandler } from "../lib/v3/handlers/extractHandler";
 import { ObserveHandler } from "../lib/v3/handlers/observeHandler";
-// Static Page type to satisfy handler constructor contracts without a real page.
 import type { Page } from "../lib/v3/understudy/page";
-// Pull in the client option type so we can build a minimal config object.
 import type { ClientOptions } from "../lib/v3/types/public/model";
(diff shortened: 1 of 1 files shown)


seanmcguire12:
@greptileai

packages/core/lib/v3/handlers/extractHandler.ts
```diff
@@ -106,146 +110,135 @@ export class ExtractHandler {
...
+      message: completed
+        ? "Extraction completed successfully"
+        : "Extraction incomplete after processing all data",
+      level: 1,
+      auxiliary: {
+        prompt_tokens: { value: String(prompt_tokens), type: "string" },
+        completion_tokens: { value: String(completion_tokens), type: "string" },
+        inference_time_ms: {
+          value: String(inference_time_ms),
+          type: "string",
```
miguelg719, line 209: we should move this to after injecting urls to prevent confusion on logs whenever users extract with urls

miguelg719 (approved):

PR #1359, opened 2025-12-04
Update model handling for agent evals

tkattkat:
# why

We need to update the model handling to account for v3 changes in agent evals cli 

# what changed

- adjusted handling for cua models to properly utilize model map
- added /cua suffix to model strings to easily differentiate if the model should be used for a cua, or non cua run. (this is mainly only relevant for anthropic models, as they can be used for cua, and non )

# test plan

---
## Summary by cubic
Update agent evals v3 to replace the /cua suffix with explicit CUA/non-CUA model lists and strengthen model/provider mapping and fallback. This clarifies model selection and ensures correct API keys and internal models are used.

- **New Features**
  - Split models into DEFAULT_AGENT_MODELS_NON_CUA and DEFAULT_AGENT_MODELS_CUA; added isModelCUA(); defaults use vendor-prefixed names.
  - Validate provider lookup using short model names and select a safe internal model for CUA when API keys are available.

- **Migration**
  - Remove the /cua suffix. Use vendor/model-name; CUA is determined by the CUA list.
  - If overriding models, set EVAL_AGENT_MODELS with vendor-prefixed names (no suffix).

<sup>Written for commit 9dca11fcb9ec1c491aba0a03f1c5817d7f5ba021. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/evals/initV3.ts
+++ b/packages/evals/initV3.ts
@@ -23,6 +23,7 @@ import {
 import dotenv from "dotenv";
 import { env } from "./env";
 import { EvalLogger } from "./logger";
+import { isModelCUA } from "./taskConfig";
 
 dotenv.config();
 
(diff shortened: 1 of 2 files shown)


miguelg719:
we should have a way to break DEFAULT_AGENT_MODELS into cua/non-cua models instead, rather than appending the /cua piece to the model string @claude

PR #1399, opened 2025-12-10
kill cua agent on stagehand.close

tkattkat:
# Why

CUA agent loops continued running after `stagehand.close()` was called, causing errors like "Cannot read properties of null (reading 'awaitActivePage')".

# What Changed

- Added `StagehandClosedError` class to detect when Stagehand session is closed
- `V3CuaAgentHandler` now checks if context is null before executing actions/screenshots and throws `StagehandClosedError`
- All CUA clients (Anthropic, Google, OpenAI, Microsoft) re-throw `StagehandClosedError` in catch blocks to stop the agent loop immediately instead of swallowing the error

---
## Summary by cubic
Stops CUA agent loops immediately after stagehand.close() to prevent null-context errors and stray actions. Adds a clear signal so clients and handlers halt cleanly.

- **Bug Fixes**
  - Introduced StagehandClosedError to indicate a closed session.
  - V3CuaAgentHandler checks context before screenshots/actions and throws StagehandClosedError.
  - Anthropic, Google, and OpenAI CUA clients rethrow StagehandClosedError to exit loops.

<sup>Written for commit d34c5cdbadd3ff91210f63c7421d87733ed9bacb. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/core/lib/v3/agent/MicrosoftCUAClient.ts
+++ b/packages/core/lib/v3/agent/MicrosoftCUAClient.ts
@@ -8,9 +8,7 @@ import {
 } from "../types/public/agent";
 import { ClientOptions } from "../types/public/model";
 import { AgentClient } from "./AgentClient";
-import {
-  AgentScreenshotProviderError,
-} from "../types/public/sdkErrors";
+import { AgentScreenshotProviderError } from "../types/public/sdkErrors";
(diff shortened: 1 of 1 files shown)


miguelg719:
@greptileai

PR #1409, opened 2025-12-12
[docs]: update `act` reference with preferred model name formatting

ziruihao:
# why
- update `act` reference to use `"provider/model-name"` formatting

### DIFF
--- a/packages/docs/v3/references/act.mdx
+++ b/packages/docs/v3/references/act.mdx
@@ -235,7 +235,7 @@ await stagehand.act("choose 'Peach' from the favorite color dropdown", {
 await stagehand.act("choose 'Peach' from the favorite color dropdown", {
   model: {
     modelName: "google/gemini-2.5-flash",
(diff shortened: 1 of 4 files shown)


packages/docs/v3/references/act.mdx
```diff
@@ -234,8 +234,8 @@ await stagehand.act("choose 'Peach' from the favorite color dropdown", {
 // Using object format with custom configuration
 await stagehand.act("choose 'Peach' from the favorite color dropdown", {
   model: {
-    modelName: "gemini-2.5-flash",
-    apiKey: process.env.GOOGLE_API_KEY,
+    modelName: "google/gemini-2.5-flash",
```
ziruihao, line 237: Assuming we want the `google/` prefix here, like the example above it.

seanmcguire12: yes nice catch

PR #1420, opened 2025-12-15
[STG-1053] [server] Use fastify-zod-openapi + zod v4 for openapi generation

monadoid:
# why
 - Our OpenAPI file was stale and hand-maintained; routes and shapes diverged from reality (e.g., /act and /start payloads).
  - Validation had been disabled; we lacked consistent request/response enforcement and clear error reporting.
  - We were on Zod v3 plus zod-to-json-schema
  - We need generated docs that match the live Zod schemas and a path to stricter, type-safe responses.
# what changed
- Upgraded to zod v4 and removed zod-to-json-schema; we now rely on Zod’s native schemas directly with fastify-zod-openapi.
  - Integrated fastify-zod-openapi + @fastify/swagger/UI
  - Added pnpm gen:openapi (scripts/gen-openapi.ts) to write openapi.v3.yaml from the live app.
  - Added Zod response schemas that mirror our actual envelope ({ success, data: ... }) for start/act/agentExecute/navigate/observe/extract/end so response validation is strict without breaking the API.
  - Improved error handling:
  - Schema formatter + error handler now surface Zod issues (422 for requests, 500 for response serialization) with structured logs

### Future stuff:
- Remove healthz and readiness endpoints
- Tighten up request / response schemas
# test plan
Ran eval suite - passing

---
## Summary by cubic
Generates an OpenAPI 3.1 spec from live Zod v4 schemas with strict request/response validation. /v1/sessions/start now accepts browser.type and can return connectUrl; cdpUrl is removed. Addresses STG-1053.

- **New Features**
  - Switched to Zod v4 + fastify-zod-openapi with Swagger UI; OpenAPI is generated from runtime schemas.
  - Added pnpm gen:openapi to write packages/server/openapi.v3.yaml.
  - Centralized API request/response schemas in the client and exported under Stagehand.Api; server routes use these directly.
  - /v1/sessions/start accepts browser.type ("local" | "browserbase") and returns { sessionId, available, connectUrl? }.
  - Standardized { success, data } responses; improved errors (422 on request validation, 500 on response serialization).
  - Health endpoints moved to /healthz and readiness to /readyz.
  - Added /v1/sessions/:id/replay stub.
  - Added SSE streaming via the x-stream-response header.

- **Migration**
  - Do not expect cdpUrl from /v1/sessions/start; use connectUrl when present. For Browserbase, pass browserbaseSessionID.
  - Rename localBrowserLaunchOptions.cdpUrl to connectUrl.
  - Expect the { success, data } envelope on all responses.
  - Use /healthz and /readyz instead of the old paths.
  - Run pnpm gen:openapi to refresh the spec.

<sup>Written for commit 2ae58ab477997cc279a94f1c7904dcba9ca40efe. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/server/src/lib/errorHandler.ts
+++ b/packages/server/src/lib/errorHandler.ts
@@ -7,8 +7,6 @@ import { StatusCodes } from "http-status-codes";
 
 import { error } from "./response.js";
 
-const INTERNAL_SERVER_ERROR_STATUS_CODE = 500;
-
 export class AppError extends Error {
   statusCode: number;
   isInternal: boolean;
@@ -32,7 +30,7 @@ export class AppError extends Error {
    */
   getClientMessage(): string {
     if (this.isInternal) {
-      return this.statusCode >= INTERNAL_SERVER_ERROR_STATUS_CODE
+      return this.statusCode >= StatusCodes.INTERNAL_SERVER_ERROR
         ? "An internal server error occurred"
         : "An error occurred while processing your request";
     }
--- a/packages/core/lib/v3/types/public/api.ts
+++ b/packages/core/lib/v3/types/public/api.ts
@@ -329,10 +329,7 @@ export const SessionStartRequestSchema = z
       example: 5000,
     }),
     verbose: z
-      .number()
-      .int()
-      .min(0)
-      .max(2)
+      .union([z.literal(0), z.literal(1), z.literal(2)])
       .optional()
       .meta({
         description: "Logging verbosity level (0=quiet, 1=normal, 2=debug)",


packages/core/lib/v3/types/private/api.ts
```diff
@@ -33,11 +33,15 @@ export interface StartSessionParams {
   > & { projectId?: string };
   selfHeal?: boolean;
   browserbaseSessionID?: string;
+  browser?: {
```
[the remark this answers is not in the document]
pirate: local should define all the params/headers that cloud requires, just marked as optional + a comment saying cloud-only

packages/core/lib/v3/v3.ts
```diff
@@ -863,11 +867,71 @@ export class V3 {
...
+            };
+            this.browserbaseSessionId = this.opts.browserbaseSessionID ?? "";
+            this.browserbaseSessionUrl = `https://www.browserbase.com/sessions/${this.browserbaseSessionId}`;
+            return;
+          }
+          // If we're in local-browser API mode and still no cdpUrl, fail fast
+          if ((this.opts.browser?.type ?? "browserbase") === "local") {
+            throw new StagehandInitError(
+              "API mode requested local browser, but server did not return a cdpUrl.",
+            );
```
pirate: why are these changes needed? wasn't the client previously handling this stuff correctly?

packages/core/lib/v3/types/public/options.ts
```diff
@@ -1,71 +1,36 @@
...
+  typeof LocalBrowserLaunchOptionsSchema
+>;
 
 /** Constructor options for V3 */
 export interface V3Options {
   env: V3Env;
   // Browserbase (required when env = "BROWSERBASE")
   apiKey?: string;
   projectId?: string;
+  browser?: {
```
seanmcguire12: this would be breaking, can we make it a separate PR?

packages/core/lib/v3/v3.ts
```diff
@@ -699,14 +699,14 @@ export class V3 {
           const lbo: LocalBrowserLaunchOptions =
             this.opts.localBrowserLaunchOptions ?? {};
 
-          // If a CDP URL is provided, attach instead of launching.
-          if (lbo.cdpUrl) {
+          // If a connect URL is provided, attach instead of launching.
+          if (lbo.connectUrl) {
```
seanmcguire12: also breaking, can we revert @pirate

packages/evals/args.ts
```diff
@@ -12,6 +12,7 @@ const parsedArgs: {
   evalName?: string;
   env?: string;
   api?: string;
+  browser?: { type?: "browserbase" | "local" };
```
seanmcguire12: separate PR also @pirate

packages/server/src/lib/SessionStore.ts
```diff
@@ -45,6 +48,9 @@ export interface CreateSessionParams {
   /** Local browser launch overrides when browserType is local */
   localBrowserLaunchOptions?: LocalBrowserLaunchOptions;
 
+  /** WebSocket URL for connecting to the browser (returned to client) */
+  connectUrl?: string;
+
   // Cloud-specific metadata fields
   /** Debug DOM mode */
   debugDom?: boolean;
```
seanmcguire12: debugDom is a legacy param feel free to rm

packages/server/src/routes/v1/sessions/:id/extract.ts
```diff
@@ -1,41 +1,15 @@
 import type { RouteHandlerMethod, RouteOptions } from "fastify";
 import { StatusCodes } from "http-status-codes";
-import { zodToJsonSchema } from "zod-to-json-schema";
-import { z } from "zod/v3";
+import type { ZodTypeAny } from "zod/v3";
```
seanmcguire12, line 3: should this be v4? looks like some imports are v3 & others are v4

packages/core/lib/v3/understudy/response.ts
```diff
@@ -367,7 +367,7 @@ export class Response {
       requestId: serialized.requestId,
       frameId: serialized.frameId,
       loaderId: serialized.loaderId,
-      response: serialized.response,
+      response: serialized.response as Protocol.Network.Response,
```
seanmcguire12, line 370: why are we casting this now?

packages/core/lib/v3/types/public/options.ts
```diff
@@ -1,57 +1,20 @@
-import Browserbase from "@browserbasehq/sdk";
+import { z } from "zod/v4";
```
tkattkat: I think we need to remove the /v4 or may break peoples stuff who use v3? - not confident on this

packages/core/lib/v3/types/public/options.ts
```diff
@@ -62,10 +25,7 @@ export interface V3Options {
   /**
    * Optional: fine-tune Browserbase session creation or resume an existing session.
    */
-  browserbaseSessionCreateParams?: Omit<
-    Browserbase.Sessions.SessionCreateParams,
-    "projectId"
-  > & { projectId?: string };
+  browserbaseSessionCreateParams?: BrowserbaseSessionCreateParams;
```
seanmcguire12, line 28: is it possible to keep the Omit type? it prevents ppl from having to define project ID twice

packages/core/lib/v3/understudy/response.ts
pirate: it had no type hint before, but we know what shape CDP will return so we may as well add the hint so anything downstreme of here knows the type.

packages/core/lib/v3/types/public/options.ts
pirate: it's already marked optional in the zod type `BrowserbaseSessionCreateParamsSchema`, we're no longer using the browserbase sdk type because we need zod in order to generate openapi schemas (the bb sdk only provides typescript interfaces not zod schemas)

packages/server/src/routes/v1/sessions/:id/extract.ts
pirate: we can change this in next pr, packages/server is only in this pr to support the openapi spec generation and isn't called by anything

packages/server/openapi.v3.yaml
```diff
@@ -1,237 +1,2076 @@
...
+        streamResponse:
+          description: Whether to stream the response via SSE
+          example: true
+          type: boolean
+    ExtractResult:
+      type: object
+      properties:
+        result:
+          description: Extracted data matching the requested schema
+          type: unknown
```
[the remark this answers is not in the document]
pirate: it's expected by stainless

packages/server/src/routes/v1/sessions/start.ts
```diff
@@ -105,27 +114,80 @@ const startRouteHandler: RouteHandler = withErrorHandling(
...
+          return error(reply, "Browserbase session missing connectUrl");
+        }
+      }
     }
 
     const sessionStore = getSessionStore();
+
+    // For local browsers without a connectUrl, get it from browser.connectUrl
+    if (browserType === "local") {
+      connectUrl = browser?.cdpUrl;
```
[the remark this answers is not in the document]
pirate: it's fine, will normalize in separate pr

packages/server/scripts/gen-openapi.ts
```diff
@@ -0,0 +1,217 @@
...
+      AgentUsage: Api.AgentUsageSchema,
+      AgentResultData: Api.AgentResultDataSchema,
+      AgentExecuteOptions: Api.AgentExecuteOptionsSchema,
+      AgentExecuteRequest: Api.AgentExecuteRequestSchema,
+      AgentExecuteResult: Api.AgentExecuteResultSchema,
+      AgentExecuteResponse: Api.AgentExecuteResponseSchema,
+      // Navigate
+      NavigateOptions: Api.NavigateOptionsSchema,
+      NavigateRequest: Api.NavigateRequestSchema,
+      NavigateResultData: Api.NavigateResultDataSchema,
```
miguelg719: Api.NavigateResultDataSchema is not defined in api types, does it need to be?

pirate: it's not actually needed, `undefined` / unknown is fine because it's the raw playwright/CDP response object, I'll just remove this line for clarity. fixed in 211b19e5

packages/core/lib/v3/api.ts
```diff
@@ -81,36 +149,44 @@ export class StagehandAPIClient {
...
     this.modelProvider = modelName?.includes("/")
       ? modelName.split("/")[0]
       : undefined;
 
     const region = browserbaseSessionCreateParams?.region;
     if (region && region !== "us-west-2") {
-      return { sessionId: browserbaseSessionID ?? null, available: false };
+      throw new StagehandAPIError(
+        `Unsupported Browserbase region requested: ${region}`,
+      );
```
miguelg719: this breaks the disabled api path for client - browser direct connection

pirate: good catch, fixed in [84868d0](https://github.com/browserbase/stagehand/pull/1420/commits/84868d00da9b24273b7ac3581940f024f07d886e)

pirate:
I got claude to do a deep-dive review to look for any unintentional changes introduced in `api.ts` or `v3.ts`, it caught a few things which I fixed, the final report is:

Screenshot 2025-12-18 at 6 12 21 PM

packages/server/src/lib/errorHandler.ts
```diff
@@ -7,6 +7,8 @@ import { StatusCodes } from "http-status-codes";
 
 import { error } from "./response.js";
 
+const INTERNAL_SERVER_ERROR_STATUS_CODE = 500;
```
miguelg719: not sure if we have the same ts / lints restrictions here, but we should. Would it be useful to use `http-status-codes` rather than declare these?

packages/core/lib/v3/types/public/options.ts
```diff
@@ -78,6 +38,9 @@ export interface V3Options {
   experimental?: boolean;
   verbose?: 0 | 1 | 2;
   selfHeal?: boolean;
+  // V2 compatibility fields - only included because the server imports this type and supports V2
+  waitForCaptchaSolves?: boolean;
```
seanmcguire12, line 42: waitForCaptchaSolves & actTimeoutMs are both v2 params

PR #1424, opened 2025-12-16
[fix]: handle CBOR stack limit errors

seanmcguire12:
# why
- when running on a site with deep nesting of elements/nodes, the `captureHybridSnapshot()` will fail and log this error:
`Error: -32000 Failed to convert response to JSON: CBOR: stack limit exceeded at position 60578`
- this is because `Dom.getDocument` fails 
# what changed
- added handling and retries to `Dom.getDocument` where we exponentially reduce the depth parameter on each retry
# test plan
- added an e2e test `nested-div.spec.ts` just to confirm the `captureHybridSnapshot()` fn does not throw on the test website
- added unit tests in `snapshot-cbor.test.ts` which use mock data to confirm that (1) retries are actually happening when depth is exceeded, (2) after a failed step, missing nodes are glued back into the tree, and (3) the `StagehandDomProcessError` is thrown when all depth attempts have failed

---
## Summary by cubic
Prevent CBOR “stack limit exceeded” errors in captureHybridSnapshot and DOM map building on deeply nested pages. Adapt DOM depth and hydrate missing nodes so snapshots complete reliably.

- **Bug Fixes**
  - Added adaptive retries for DOM.getDocument with reducing depth.
  - Hydrated truncated trees via DOM.describeNode with fallback depths; merged children/shadow roots/contentDocument.
  - Threw StagehandDomProcessError with a clear message when all attempts fail.
  - Tests: e2e to confirm no throw on a deeply nested site; unit tests for retries, hydration, and failure path.

<sup>Written for commit b11476a3067c6fdac00f16b7e930160700254113. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/core/lib/v3/understudy/a11y/snapshot.ts
+++ b/packages/core/lib/v3/understudy/a11y/snapshot.ts
@@ -506,19 +506,27 @@ async function absoluteXPathForBackendNode(
   }
 }
 
+// starting from infinite depth (-1), exponentially shrink down to 1
 const DOM_DEPTH_ATTEMPTS = [-1, 256, 128, 64, 32, 16, 8, 4, 2, 1];
 const DESCRIBE_DEPTH_ATTEMPTS = [-1, 64, 32, 16, 8, 4, 2, 1];
 
+/** Identify CDP failures caused by deep DOM trees blowing the CBOR encoder stack. */
 function isCborStackError(message: string): boolean {
   return message.includes("CBOR: stack limit exceeded");
 }
 
+/**
+ * Determine if CDP truncated a node's children when streaming the DOM tree.
+ * childNodeCount stays accurate even when `children` are omitted; we use this to
+ * decide whether DOM.describeNode must be re-run for that node.
+ */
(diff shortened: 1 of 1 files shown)


packages/core/lib/v3/understudy/a11y/snapshot.ts
```diff
@@ -503,6 +506,150 @@ async function absoluteXPathForBackendNode(
   }
 }
 
+const DOM_DEPTH_ATTEMPTS = [-1, 256, 128, 64, 32, 16, 8, 4, 2, 1];
+const DESCRIBE_DEPTH_ATTEMPTS = [-1, 64, 32, 16, 8, 4, 2, 1];
```
pirate, line 511: worth a comment explaining what these magic nums are

pirate (approved):
tenatively approve this PR, but it could use some more docstrings/comments explaining the logic flow / why these steps are necessary (in the code next to the logic, not just as PR comments).

also worth explaining how it interacts with iframes on different origins across a page, shadow doms, etc.

PR #1427, opened 2025-12-17
[chore]: rm inlined JS from `snapshot.ts`

seanmcguire12:
# why
  - `a11y/snapshot.ts` has large inline stringified scripts for scroll offsets, active-element lookup, and XPath building. stringified JS is not good practice as it is difficult to debug and is ignored by linting
  - there was also duplicate selector resolution logic inside `snapshot.ts` which already exists inside `locatorScripts` dir

# what changed
  - added `lib/v3/dom/a11yScripts` plus a new `genA11yScripts.ts` generator so those helpers are bundled once, exported as strings and injected as needed
  - added shared `buildA11yInvocation` & `buildLocatorInvocation` helpers that wrap the JS expressions with a guard that checks if the script already exists
  - changed `a11y/snapshot.ts` to consume the generated helpers for scroll offsets, deep-active resolution, and XPath serialization, and to call locator scripts for CSS/XPath queries instead of embedding custom JS strings.
# test plan
- existing tests 

---
## Summary by cubic
Replaced inline stringified JS in a11y/snapshot.ts with a generated a11y script bundle and small invocation helpers for cleaner, lintable code. Also routed CSS/XPath queries through existing locator scripts to remove duplicated logic.

- **Refactors**
  - Added lib/v3/dom/a11yScripts and genA11yScripts.ts to bundle and export a11y helpers as strings.
  - Introduced buildA11yInvocation and buildLocatorInvocation to inject bundles and call helpers safely.
  - Updated snapshot.ts to use generated helpers for scroll offsets, deep-active element, bounding rect, and absolute XPath.
  - Switched CSS/XPath resolution to locator scripts instead of custom inlined JS.
  - Trimmed ~180 lines of inline JS from snapshot.ts; improved debuggability and lint coverage.
  - Updated build-dom-scripts to include a11y script generation.

<sup>Written for commit a799be8d8d2aa5813111500bbbc841fe6a5daa77. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/core/lib/v3/dom/a11yScripts/index.ts
+++ b/packages/core/lib/v3/dom/a11yScripts/index.ts
@@ -46,64 +46,65 @@ export function resolveDeepActiveElement(): Element | null {
   }
 }
 
-const nodeToAbsoluteXPathImpl = (node: Node | null | undefined): string => {
-  try {
-    const sibIndex = (n: Node | null | undefined): number => {
-      if (!n || !n.parentNode) return 1;
-      let i = 1;
-      const targetKey = `${n.nodeType}:${(n.nodeName || "").toLowerCase()}`;
-      for (let p = n.previousSibling; p; p = p.previousSibling) {
-        const key = `${p.nodeType}:${(p.nodeName || "").toLowerCase()}`;
-        if (key === targetKey) i += 1;
-      }
-      return i;
-    };
-
-    const step = (n: Node | null | undefined): string => {
-      if (!n) return "";
-      if (n.nodeType === Node.DOCUMENT_NODE) return "";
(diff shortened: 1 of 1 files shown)


seanmcguire12:
@greptileai

packages/core/lib/v3/dom/genA11yScripts.ts
```diff
@@ -0,0 +1,75 @@
...
+
+async function main(): Promise<void> {
+  fs.mkdirSync(outDir, { recursive: true });
+
+  esbuild.buildSync({
+    entryPoints: [entry],
+    bundle: true,
+    format: "esm",
+    platform: "browser",
+    target: "es2020",
```
miguelg719, line 21: we should add a note in case we upgrade our builds to ES2022 to remember to update here

PR #1437, opened 2025-12-18
Update screenshot collector to use existing screenshots, handle events automagically & update evals to all use collector

tkattkat:
# why

- Currently, when using screenshot collector with hybrid mode, it takes a new screenshot on every step
- we have outdated evals which are flaky, and less reliably than screenshot collector based ones 
- screenshot collector currently requires manually handling the events in the file it is defined within

# what changed

- now, we just pass the screenshot taken by agent, directly into the collector when hybrid mode is used. 
- for dom mode, we still take on every step as a non vision model is much more likely to take its own screenshots 
- removed dead code from screenshot collector type 
- moved event handling for screenshot collector into the collector.
- removed timeout from all evals, so they utilize only event images 
- updated all non screenshot collector evals to utilize them, and fixed two evals which would throw false negatives ( apple_tv & nba trades) 
# test plan

---
## Summary by cubic
Use agent-provided screenshots in hybrid mode to stop duplicate captures and reduce eval overhead. The screenshot collector now auto-subscribes to agent events, and all evals were updated to use it.

- **Refactors**
  - Hybrid: forward screenshot tool output directly to the collector (no new page screenshot).
  - DOM (evals only): still capture after each tool execution.
  - Collector: automatically handles bus events; optional interval capture remains.
  - Evals: switched to event-driven collector, removed manual listeners, and pass screenshots to the evaluator.
  - Removed unused ScreenshotCapablePage type.

- **Bug Fixes**
  - Corrected event name to agent_screenshot_taken_event.

<sup>Written for commit d9d5e26d3dd29365c0a63316bf9cad806414d29c. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/core/lib/v3/handlers/v3AgentHandler.ts
+++ b/packages/core/lib/v3/handlers/v3AgentHandler.ts
@@ -168,6 +168,14 @@ export class V3AgentHandler {
                 : allReasoning || "Task completed successfully";
             }
           }
+          // Intercept screenshot tool output for evals (hybrid mode only)
+          if (this.mode === "hybrid" && toolCall.toolName === "screenshot") {
+            const screenshotOutput = event.toolResults?.[i]?.output;
+            this.v3.bus.emit(
+              "agent_screensot_taken_event",
+              Buffer.from(screenshotOutput.base64, "base64"),
+            );
+          }
           const mappedActions = mapToolResultToActions({
             toolCallName: toolCall.toolName,
             toolResult,
@@ -183,17 +191,9 @@ export class V3AgentHandler {
         }
         state.currentPageUrl = (await this.v3.context.awaitActivePage()).url();
 
-        // Capture screenshot after tool execution (only for evals)
-        if (process.env.EVALS === "true") {
-          try {
-            await this.captureAndEmitScreenshot();
-          } catch (e) {
-            this.logger({
-              category: "agent",
-              message: `Warning: Failed to capture screenshot: ${getErrorMessage(e)}`,
-              level: 1,
-            });
-          }
+        // Capture screenshot after tool execution (only for evals, DOM mode)
+        if (process.env.EVALS === "true" && this.mode === "dom") {
+          await this.captureAndEmitScreenshot();
         }
       }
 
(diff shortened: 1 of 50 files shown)


packages/core/lib/v3/handlers/v3AgentHandler.ts
```diff
@@ -168,6 +168,14 @@ export class V3AgentHandler {
                 : allReasoning || "Task completed successfully";
             }
           }
+          // Intercept screenshot tool output for evals (hybrid mode only)
+          if (this.mode === "hybrid" && toolCall.toolName === "screenshot") {
```
miguelg719: is this only meant for evals as well? If so, should it also include the process.env.EVALS check?

packages/core/lib/v3/handlers/v3AgentHandler.ts
```diff
@@ -168,6 +168,14 @@ export class V3AgentHandler {
...
+            const screenshotOutput = event.toolResults?.[i]?.output;
+            this.v3.bus.emit(
```
miguelg719: at this point we have two places where we emit the event, we should consider breaking the captureAndEmitScreenshot into emitScreenshot and move capturing logic outside/intercept

tkattkat: good call

packages/core/lib/v3/handlers/v3AgentHandler.ts
tkattkat: Yea we should

packages/core/lib/v3/handlers/v3AgentHandler.ts
```diff
@@ -489,13 +481,32 @@ export class V3AgentHandler {
...
+        if (options.toolName !== "screenshot" || !options.toolOutput?.base64) {
+          return;
+        }
+        screenshot = Buffer.from(options.toolOutput.base64, "base64");
+      } else {
+        const page = await this.v3.context.awaitActivePage();
+        screenshot = await page.screenshot({ fullPage: false });
+      }
+
+      this.v3.bus.emit("agent_screenshot_taken_event", screenshot);
```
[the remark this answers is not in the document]
tkattkat: this has been addressed no?

PR #1440, opened 2025-12-19
[Feature] exclude tools from agent

tkattkat:
# why

Allow users to exclude specific tools from agent 

# what changed

- Added `excludeTools` option to `agent.execute()` that accepts an array of tool names to exclude
- Excluded tools are filtered from both the toolset and system prompt
- Feature is experimental and not supported in CUA mode

# test plan

```typescript
  env: "LOCAL",
  experimental: true,
});
await stagehand.init();

const agent = stagehand.agent();
  instruction: "click the submit button",
  excludeTools: ["screenshot", "extract"],
});
```

---
## Summary by cubic
Adds an excludeTools option to agent.execute so you can disable specific tools per run. Excluded tools are removed from both the available toolset and the system prompt to reduce unwanted calls.

- **New Features**
  - agent.execute({ excludeTools: string[] }) to skip tools; experimental and not supported in CUA mode
  - createAgentTools now applies mode-based and explicit exclusions
  - System prompt builds the tools list dynamically and omits excluded tools
  - Validation enforces no excludeTools in CUA and requires experimental; handler passes the option to tools and prompt

<sup>Written for commit c041b45fc89bdd9016f9401800d97b4a839cda44. Summary will update automatically on new commits.</sup>


packages/core/lib/v3/agent/prompts/agentSystemPrompt.ts
```diff
@@ -15,6 +17,103 @@ export interface AgentSystemPromptOptions {
...
+    { name: "think", description: "Think about the task" },
+    { name: "extract", description: "Extract structured data" },
+    { name: "goto", description: "Navigate to a URL" },
+    { name: "wait", description: "Wait for a specified time" },
+    { name: "navback", description: "Navigate back in browser history" },
+    { name: "scroll", description: "Scroll the page x pixels up or down" },
+    { name: "close", description: "Mark the task as complete or failed" },
+  ];
+
+  const baseTools = isHybridMode ? hybridTools : domTools;
```
miguelg719, line 98: nice rearrangement, we could also centralize some of the commonly shared tools (goto, close..) to extend with the custom hybrid/dom ones

miguelg719 (commented):
lgtm we should add a couple unit tests. Also should there be tools that cannot be disabled (i.e close)?

DLillard0:
@tkattkat When will this feature be merged?

seanmcguire12 (approved):

pkiv (approved):

PR #1446, opened 2025-12-19
[test]: add tests for ax tree post-processing

seanmcguire12:
# why
- to add more coverage for the hybrid snapshot flow by testing lower level helpers
# what changed
added `tests/snapshot-a11y-tree-utils.test.ts` which tests:
- `decorateRoles()`
  - checks that scrollable nodes get labelled correctly
  - checks that `#document` nodes get left as 'generic'
  - checks that `html` nodes always get the scrollable label
- `buildHierarchicalTree()`
  - checks that structural nodes with no meaningful content are excluded
  - `combobox` & `select` roles are remapped to their HTML tag names
  - tag name substitution is applied to non-combobox structural nodes
  - nodes with negative IDs are skipped
- `isStructural()`
   - confirms "generic", "none", and "InlineTextBox" roles are flagged structural while real roles like "button" stay non-structural
- `removeRedundantStaticTextChildren()`
   - removes `StaticText` children when their combined text exactly matches the parent label, keeps them when the text differs, and returns the original array when the parent has no name
- `extractUrlFromAXNode()`
  - trims and returns the URL when a valid url property exists, but yields undefined when the property is missing or the value isn’t a string
# test plan
- this is it

---
## Summary by cubic
Add unit tests for AX tree post-processing helpers to increase coverage of the hybrid snapshot flow (STG-1095). Covers role decoration (scrollable labels, encoding fallback), hierarchical tree building (structural pruning, tag remapping, negative ID skip), structural role detection, redundant StaticText pruning, and URL extraction.

<sup>Written for commit 310fd7508234be2b711deb6fa12ae8c8475e24be. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/core/tests/snapshot-a11y-tree-utils.test.ts
+++ b/packages/core/tests/snapshot-a11y-tree-utils.test.ts
@@ -130,7 +130,6 @@ describe("buildHierarchicalTree", () => {
         encodedId: "root",
         parentId: undefined,
         childIds: ["child"],
-        children: [],
       },
       {
         role: "StaticText",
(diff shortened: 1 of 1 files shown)


seanmcguire12:
@greptileai

monadoid (approved):

PR #1451, opened 2025-12-20
[test]: add tests for ax tree fetching & object ID resolvers

seanmcguire12:
# why
- add more unit tests for the functions that `captureHybridSnapshot()` depends on
# what changed
added `snapshot-a11y-resolvers.test.ts` which tests the following functions:
- `a11yForFrame()`
  - checks that the unscoped path/no selector provided path returns the full AX outline and url map
  - checks that xpath scoping trims the tree to the resolved node, including the branch where frame scoping fails, and it retries without a `frameId`
  - checks that when selector resolution throws (e.g., CSS resolver rejects), the function falls back to the full tree and leaves `scopeApplied` false
- `resolveObjectIdForXPath()`
  - checks that it waits for the frame’s main world, evaluates the helper script, and returns the resulting object id
  - checks that that runtime exceptions (or missing contexts) cause it to return null
- `resolveObjectIdForCss()` 
  - checks that on `primaryExpr` evaluation success, the object id is returned
  - checks that `fallbackExpr` is used when `primaryExpr` fails 
  - checks that null is returned if both `primaryExpr` and `fallbackExpr` fail
- `tryScopedSnapshot()` 
  - checks that the CSS and xpath focus selectors both produce prefixed combined maps when scoping succeeds
  - checks that when `a11yForFrame()` returns `scopeApplied: false` the helper logs and returns null
  - checks that when focus resolution throws, the catch block logs and returns null
# test plan
- this is it

---
## Summary by cubic
Adds unit tests covering AX tree scoping, CSS/XPath object ID resolvers, DOM session builders, and CBOR fallback paths to harden captureHybridSnapshot and related helpers.

- **Refactors**
  - Exported tryScopedSnapshot for direct testing.
  - Replaced ad-hoc test session with shared MockCDPSession.

<sup>Written for commit 1117b16d01be65145386b5f5eb938199e2bdef56. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/core/tests/snapshot-a11y-resolvers.test.ts
+++ b/packages/core/tests/snapshot-a11y-resolvers.test.ts
@@ -1,0 +1,399 @@
+import type { Protocol } from "devtools-protocol";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { a11yForFrame } from "../lib/v3/understudy/a11y/snapshot/a11yTree";
+import type { AccessibilityTreeResult } from "../lib/v3/types/private";
+import * as focusSelectors from "../lib/v3/understudy/a11y/snapshot/focusSelectors";
+import { MockCDPSession } from "./helpers/mockCDPSession";
+import { executionContexts } from "../lib/v3/understudy/executionContextRegistry";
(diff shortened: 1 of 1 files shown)


seanmcguire12:
@greptileai

PR #1456, opened 2025-12-22
Invoke page.hover on agent move action

shrey150:
# why
- CUA agents couldn't hover over elements to reveal tooltips without clicking
- `move` action was implemented as a no-op, silently returning success without dispatching mouse events
- This was noticed when `highlightCursor` was set to true

# what changed
- Modified move case in `v3CuaAgentHandler.ts` to call `page.hover(x, y)` instead of just returning success
- This dispatches mouseMoved CDP events and triggers CSS `:hover` states

# test plan
- Created local test with hover-triggered tooltips
- Verified `hover_at` actions now trigger `mouseenter` events and display tooltips
- Confirmed clicks still work correctly

---
## Summary by cubic
Fix agent move to trigger hover effects, so tooltips and CSS :hover states appear without clicks. Also records hover steps with a selector during recording for replay.

- **Bug Fixes**
  - Invoke page.hover(x, y) in the move case of v3CuaAgentHandler when coordinates are provided.
  - Emits mouseMoved/mouseenter events; verified tooltips show and clicks still work.

- **New Features**
  - When recording, capture the hovered element’s XPath and log a hover action for replay.

<sup>Written for commit 66d93def6b975778b16ae1059ad35396dafb1c89. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/core/lib/v3/handlers/v3CuaAgentHandler.ts
+++ b/packages/core/lib/v3/handlers/v3CuaAgentHandler.ts
@@ -431,7 +431,25 @@ export class V3CuaAgentHandler {
       case "move": {
         const { x, y } = action;
         if (typeof x === "number" && typeof y === "number") {
-          await page.hover(x, y);
+          if (recording) {
+            const xpath = await page.hover(x, y, { returnXpath: true });
+            const normalized = ensureXPath(xpath);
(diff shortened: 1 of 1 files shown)


shrey150:
@greptileai

PR #1472, opened 2025-12-24
[fix]: refresh agent cache on action failure

seanmcguire12:
# why
- when the agent is replaying cached actions and an xpath (from the cache) does not resolve, the agent will heal and reprocess the page
- however, the cached entry is not getting updated with the "healed" selector
# what changed
- added logic so agent cache replays detect when deterministic actions self-heal and immediately rewrite the cache entry with the updated selectors/timestamps, preventing perpetual failures on corrupted entries
# test plan
- added `agent-cache-self-heal.spec.ts` end-to-end test that runs an agent with a real cacheDir, corrupts a cached selector, reruns the workflow, and asserts the cache file now contains the healed XPath

---
## Summary by cubic
Fixes agent cache not refreshing after self-heal, so healed selectors are saved and future replays don’t keep failing. Addresses Linear STG-1106.

- **Bug Fixes**
  - Detect updated actions during replay (act/fillForm) and update the cache entry when actions change.
  - Persist refreshed steps and timestamp via refreshAgentCacheEntry.
  - Add e2e test and CI job to verify a corrupted selector is healed and the cache file is rewritten.

<sup>Written for commit 9634d84e9a81c38670326f85f0353d95649da7da. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/core/lib/v3/tests/agent-cache-self-heal.spec.ts
+++ b/packages/core/lib/v3/tests/agent-cache-self-heal.spec.ts
@@ -1,5 +1,4 @@
 import { test, expect } from "@playwright/test";
-import type { TestInfo } from "@playwright/test";
 import fs from "fs/promises";
 import path from "path";
 import { V3 } from "../v3";
@@ -7,11 +6,8 @@ import { v3TestConfig } from "./v3.config";
 import type {
   AgentReplayActStep,
   AgentReplayFillFormStep,
-  AgentReplayStep,
   CachedAgentEntry,
 } from "../types/private/cache";
-
-const LOG_PREFIX = "[agent-cache-self-heal]";
 
 test.describe("Agent cache self-heal (e2e)", () => {
   let v3: V3;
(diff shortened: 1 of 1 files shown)


seanmcguire12:
@greptileai

packages/core/lib/v3/cache/AgentCache.ts
```diff
@@ -495,14 +496,20 @@ export class AgentCache {
...
     const handler = this.getActHandler();
     if (!ctx || !handler) return null;
     try {
+      const updatedSteps: AgentReplayStep[] = [];
+      let stepsChanged = false;
       for (const step of entry.steps ?? []) {
-        await this.executeAgentReplayStep(step, ctx, handler);
+        const replayedStep =
+          (await this.executeAgentReplayStep(step, ctx, handler)) ?? step;
+        stepsChanged ||= replayedStep !== step;
```
miguelg719, line 511: first time I see this shorthand

miguelg719 (approved):

PR #1483, opened 2025-12-29
Screenshot after actions

tkattkat:
# Why

Currently, we only provide screenshots to the agent when it explicitly calls the `screenshot` tool.

This is wasteful on both **tokens** and **latency** as it requires the LLM to make an additional tool call to get a new screenshot in situations where we can reliably assume it will need one — like after clicking, typing, scrolling, or waiting for content to load.

# What Changed

### Screenshot Feedback After Vision Actions
- `click`, `type`, `dragAndDrop`, `fillFormVision`, and `scroll` tools now automatically return a screenshot alongside the tool result
- `wait` tool returns a screenshot in **hybrid mode** only (since it's the mode that relies on visual feedback)
-  added screenshotHandler which abstracts handling delay + screenshot to a shared util function 
-  updated action handler to parse base64's from stored actions 

### Standardized Post Action Delay
- Removed Google specific delay logic
- Added a universal **500ms delay** before capturing screenshots, giving the page time to settle after actions ( some of the tools override this with their own delays, typically with 0 due to not needing to wait after these actions 

### Message Compression Updates
- All vision action tool screenshots now share the same "2 most recent" compression pool as the `screenshot` tool
- Older screenshots are stripped from message history while preserving text results

# Test Plan

- Tested locally with hybrid mode agent
- Verified screenshots are returned after each vision action
- Confirmed message compression correctly keeps only the 2 most recent screenshots across all vision tools

---
## Summary by cubic
Automatically returns a screenshot after vision actions so the agent gets instant visual feedback without extra screenshot tool calls, reducing tokens and latency.

- **New Features**
  - click, type, dragAndDrop, scroll, and fillFormVision now return a screenshot; wait returns one in hybrid mode.
  - Message compression keeps only the 2 most recent vision items (screenshots or vision-action outputs); older images are dropped.
  - Tools emit media via toModelOutput so the model receives image + text together.

- **Refactors**
  - Added waitAndCaptureScreenshot with a default 500ms post-action delay; removed Google-specific delays.
  - Excluded screenshotBase64 from recorded actions; updated action/message mapping to strip large fields.
  - Introduced typed tool results and wired wait(v3, mode) in createAgentTools for hybrid-only screenshot behavior.

<sup>Written for commit 93faeb889090fad6d649ea4780d15da2ba0dc22f. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/core/lib/v3/agent/tools/click.ts
+++ b/packages/core/lib/v3/agent/tools/click.ts
@@ -2,7 +2,10 @@ import { tool } from "ai";
 import { z } from "zod";
 import type { V3 } from "../../v3";
 import type { Action } from "../../types/public/methods";
-import type { ClickToolResult, ModelOutputContentItem } from "../../types/public/agent";
+import type {
+  ClickToolResult,
+  ModelOutputContentItem,
+} from "../../types/public/agent";
 import { processCoordinates } from "../utils/coordinateNormalization";
 import { ensureXPath } from "../utils/xpath";
 import { waitAndCaptureScreenshot } from "../utils/screenshotHandler";
--- a/packages/core/lib/v3/agent/tools/dragAndDrop.ts
+++ b/packages/core/lib/v3/agent/tools/dragAndDrop.ts
@@ -2,7 +2,10 @@ import { tool } from "ai";
 import { z } from "zod";
 import type { V3 } from "../../v3";
 import type { Action } from "../../types/public/methods";
-import type { DragAndDropToolResult, ModelOutputContentItem } from "../../types/public/agent";
+import type {
+  DragAndDropToolResult,
+  ModelOutputContentItem,
+} from "../../types/public/agent";
 import { processCoordinates } from "../utils/coordinateNormalization";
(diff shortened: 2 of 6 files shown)


packages/core/lib/v3/agent/utils/screenshotHandler.ts
```diff
@@ -0,0 +1,31 @@
...
+    await page.waitForTimeout(delayMs);
+  }
+
+  try {
+    const buffer = await page.screenshot({ fullPage: false });
+    return buffer.toString("base64");
+  } catch {
+    return undefined;
+  }
+}
```
pirate, line 31: do we need this helper? I'd just inline it and make page.waitForTimeout a no-op when delay is 0

pirate (approved):

packages/core/lib/v3/agent/utils/screenshotHandler.ts
tkattkat: I originally inlined it, but it started bothering me after using it in 6 places with a try catch for each so decided to make it a helper

PR #1487, opened 2025-12-31
Chromie/add page snapshot method

chromiebot:
# why

# what changed

# test plan

---
## Summary by cubic
Adds a public Page.snapshot() to capture a hybrid DOM + Accessibility snapshot. Returns a combined tree with XPath and URL maps, plus per-frame data; includes tests covering scoping and frame payloads.

- **New Features**
  - New Page.snapshot(options) using captureHybridSnapshot.
  - Options: focusSelector (CSS or XPath), pierceShadow (default true), experimental.
  - Exported types: HybridSnapshot, SnapshotOptions, PerFrameSnapshot.

<sup>Written for commit de6bd5f7e022a9ccebcb50d193ffd42e490fe6d1. Summary will update on new commits.</sup>

### DIFF
--- a/packages/docs/v3/references/page.mdx
+++ b/packages/docs/v3/references/page.mdx
@@ -392,89 +392,23 @@ await page.screenshot(options?: ScreenshotOptions): Promise<Buffer>
 
 ### snapshot()
 
-Capture a hybrid DOM + Accessibility snapshot of the page. This method exposes Stagehand's internal page representation, which combines DOM structure with accessibility tree information across all frames.
+Capture a hybrid DOM + Accessibility snapshot of the page.
 
 ```typescript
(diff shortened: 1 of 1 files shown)


miguelg719:
@chromiebot can we add docs to the relevant section for page.snapshot on the v3 reference docs

miguelg719:
@chromiebot simplify the docs changes for page.snapshot to match the overall level of detail in the v3 page reference docs

seanmcguire12 (commented):
@miguelg719 forgot about this one but we merged [this](https://github.com/browserbase/stagehand/pull/1518) the other day. the only difference here is that I decided not to expose args/params in the `page.snapshot()` function yet. the idea is to get signal from users on what kind of parameters they might want (which may be different from the parameters exposed by `captureHybridSnapshot()`) before we lock in an API contract

PR #1498, opened 2026-01-05
[fix]: use correct model on cache replay failure

seanmcguire12:
# why
- when cached act & agent runs fail and need to rerun inference (self heal), the default model was getting used every time. it should have been using the original model that was defined by the user. 
# what changed
- added logic to pass the the resolved `LLMClient` from `Stagehand.act()` & `Stagehand.agent()` into `ActCache.tryReplay` & `AgentCache.tryReplay(AsStream)` so cache hits reuse the same client as their original runs and only fall back to the default when no override is provided
# test plan
- added unit tests which confirm that act & agent caches use the provided override client and that non LLM steps (like goto) replay successfully without requiring one

---
## Summary by cubic
Fixes a bug where cache replay used the default model after an action failed; agent and act replays now use the requested model when inference is rerun. Addresses Linear STG-930.

- **Bug Fixes**
  - Pass the resolved LLM client (from options.model) into ActCache and AgentCache replay paths, including streaming.
  - Use the effective client when re-executing act and fillForm steps during replay.
  - Keep normal cache hits unchanged; only reruns use the specified model.

<sup>Written for commit 615c9976e0edbaa9885a65404efb0a807a1c6ff3. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/mean-melons-repeat.md
+++ b/.changeset/mean-melons-repeat.md
@@ -2,4 +2,4 @@
 "@browserbasehq/stagehand": patch
 ---
 
-fix: replaying cached actions (for agent & act) uses default model when action fails and rerunning inference is needed
+fix: replaying cached actions (for agent & act) now uses the originally defined model, (instead of default model) when action fails and rerunning inference is needed


seanmcguire12:
@greptileai

PR #1499, opened 2026-01-05
[chore]: fix bid postings eval

seanmcguire12:
# why
- this eval went stale (website changed) & was causing regression evals to fail in CI
# what changed
- updated the eval to use a cloned site
- updated success condition to check for exact value, since we no longer need to tolerate a range (we are now using a cloned site that wont change)
- also updated formatting for an unrelated changeset
# test plan
- this is it


seanmcguire12:
@greptileai

seanmcguire12:
@greptileai

tkattkat (approved):

PR #1505, opened 2026-01-06
Update close tool + add output to agent result

tkattkat:
# Why

Previously, agent executions could end without a structured final response - either by hitting `maxSteps` or the LLM breaking out of its loop without calling the `close` tool. This made it difficult to:
1. Reliably determine if a task was completed successfully
2. Extract structured data from the agent's execution

# What Changed

### Ensured Close Tool is Always Called
- Added `handleCloseToolCall` utility that forces a `close` tool call via a separate `generateText` call when the main agent loop ends without explicitly closing
- Integrated via new `ensureClosed` private method in `v3AgentHandler.ts`
- Works for both `execute()` and `stream()` modes
- Triggers when `maxSteps` is reached or the LLM stops ( completes its task) 
### Added Output Schema Support (Experimental)
- Users can now pass a Zod schema to `agent.execute({ output: z.object({...}) })` to return structured data at the end of execution
- The schema dynamically extends the close tool's input schema
- Extracted data is returned in `result.output`
- Added validation:
  - **CUA mode**: Throws `StagehandInvalidArgumentError` (not supported)
  - **Non-CUA without `experimental: true`**: Throws `ExperimentalNotConfiguredError`

### Example Usage
```typescript
  instruction: "search for a shampoo on amazon and click into one of the results",
  maxSteps: 20,
  output: z.object({
    productName: z.string().describe("The name of the shampoo product"),
    price: z.string().describe("The price of the product"),
    rating: z.string().describe("The star rating of the product"),
  }),
});

console.log(result.output);
// { productName: "...", price: "$12.99", rating: "4.5 out of 5 stars" }
```

# Test Plan

- [x] Verify close tool is called when agent naturally completes (no change in behavior)
- [x] Verify close tool is forced when `maxSteps` is reached
- [x] Verify `output` schema extracts data correctly in `execute()` mode
- [x] Verify `output` schema extracts data correctly in `stream()` mode
- [x] Verify `output` throws `StagehandInvalidArgumentError` when used with CUA mode
- [x] Verify `output` throws `ExperimentalNotConfiguredError` when used without `experimental: true`

---
## Summary by cubic
Ensures every agent run ends with a structured final response and adds optional structured output via a Zod schema. Improves reliability by always setting completion status and final reasoning.

- **New Features**
  - Always triggers a "close" tool call at the end of a run (LLM stops or maxSteps), for both execute() and stream().
  - Optional output schema: pass output: z.object({...}) to return typed data in result.output.
  - Validation: output schema is not supported in CUA (throws StagehandInvalidArgumentError). In non-CUA, requires experimental: true (throws ExperimentalNotConfiguredError otherwise).
  - Removed "close" from the main tool list and system prompt; closing is handled automatically post-run.

<sup>Written for commit dfb703ad41af9dae4a7923a7adfbe4b340c2d6ff. Summary will update on new commits.</sup>

### DIFF
--- a/packages/core/lib/v3/handlers/v3AgentHandler.ts
+++ b/packages/core/lib/v3/handlers/v3AgentHandler.ts
@@ -507,13 +507,6 @@ export class V3AgentHandler {
       );
     }
 
-    // Combine input messages with response messages for full conversation history
-    const responseMessages = result.response?.messages || [];
-    const fullMessages: ModelMessage[] = [
-      ...inputMessages,
-      ...responseMessages,
-    ];
-
     return {
       success: state.completed,
       message: state.finalMessage || "Task execution completed",
@@ -529,7 +522,7 @@ export class V3AgentHandler {
             inference_time_ms: inferenceTimeMs,
           }
         : undefined,
-      messages: fullMessages,
+      messages: inputMessages,
     };
   }
 


packages/core/lib/v3/agent/prompts/agentSystemPrompt.ts
```diff
@@ -67,7 +67,6 @@ function buildToolsSection(
     { name: "wait", description: "Wait for a specified time" },
     { name: "navback", description: "Navigate back in browser history" },
     { name: "scroll", description: "Scroll the page x pixels up or down" },
-    { name: "close", description: "Mark the task as complete or failed" },
```
pirate, line 70:
```suggestion
    { name: "close", description: "Mark the task as complete or failed" },  // TODO: consider renaming this tool to "done"
```

packages/core/lib/v3/agent/utils/handleCloseToolCall.ts
```diff
@@ -0,0 +1,128 @@
+import { generateText, ModelMessage, LanguageModel, ToolSet } from "ai";
+import { z } from "zod";
+import { tool } from "ai";
+import { LogLine } from "../../types/public/logs";
+import { StagehandZodObject } from "../../zodCompat";
+interface CloseResult {
+  reasoning: string;
+  taskComplete: boolean;
```
pirate, line 8: I think `success: true | false` may be better

packages/core/lib/v3/tests/agent-hybrid-mode.spec.ts
```diff
@@ -252,7 +249,7 @@ test.describe("Stagehand agent hybrid mode", () => {
 
       await agent.execute({
         instruction:
-          "Use the ariaTree to understand the page, then use close tool with taskComplete: true",
+          "Use the ariaTree to understand the page, then describe what you see briefly and mark the task as complete.",
```
pirate:
```suggestion
          "Use the ariaTree to understand the page, then provide the final requested output or a summary of the page.",
```

packages/core/lib/v3/agent/utils/handleCloseToolCall.ts
```diff
@@ -0,0 +1,128 @@
...
+  messages: ModelMessage[];
+  output?: Record<string, unknown>;
```
pirate, line 10: I recommend making output required, LLMs are really good at inferring what the ideal output should be for a task.

e.g. if user is researching something often it nails it and puts the exact data they were looking for in output.

packages/core/lib/v3/tests/agent-hybrid-mode.spec.ts
pirate: honestly less guidance is better, the models are smart and will do well as long as the fields are named well.

pirate (approved):

PR #1514, opened 2026-01-08
rename close tool in agent to done

tkattkat:
# why
the name "close" as a tool to end an agent run is confusing, we should instead call it "done" 

# what changed

updated all usage of the "close" tool to be "done"

# test plan

-tested locally 
-updated tests 

---
## Summary by cubic
Renamed the agent’s “close” tool to “done” for clearer task completion. Core logic, cache, types, and tests updated; behavior remains the same.

- **Refactors**
  - Replaced handleCloseToolCall with handleDoneToolCall and enforced a final “done” call.
  - Switched handler checks, action mapping, and cache replay from “close” to “done”.
  - Updated public agent types/docs to list “done” instead of “close”.
  - Adjusted tests and instructions to remove “close” references.

- **Migration**
  - Change any toolName === "close" checks to "done".
  - Update allowed/excluded tool lists to use "done".
  - Update instruction text or UI copy that tells the agent to “close” the task; use “done”.
  - If using outputSchema, read final data from the “done” tool call.

<sup>Written for commit 205d48c2352c56ec10a0dfb7c17cbdff75685fc0. Summary will update on new commits.</sup>

### DIFF
--- a/packages/core/lib/v3/agent/tools/index.ts
+++ b/packages/core/lib/v3/agent/tools/index.ts
@@ -86,7 +86,6 @@ export function createAgentTools(v3: V3, options?: V3AgentToolOptions) {
     ariaTree: ariaTreeTool(v3),
     click: clickTool(v3, provider),
     clickAndHold: clickAndHoldTool(v3, provider),
-    //done: doneTool(),
     dragAndDrop: dragAndDropTool(v3, provider),
     extract: extractTool(v3, executionModel, options?.logger),
     fillForm: fillFormTool(v3, executionModel),


packages/core/lib/v3/agent/tools/index.ts
```diff
@@ -86,7 +86,7 @@ export function createAgentTools(v3: V3, options?: V3AgentToolOptions) {
     ariaTree: ariaTreeTool(v3),
     click: clickTool(v3, provider),
     clickAndHold: clickAndHoldTool(v3, provider),
-    //close: closeTool(),
+    //done: doneTool(),
```
miguelg719: remove?

miguelg719 (approved):

PR #1516, opened 2026-01-09
[feat]: add support for local caching of agent when using api

seanmcguire12:
# why

# what changed

# test plan

---
## Summary by cubic
Adds local agent caching when running via the API. We stream the agent’s cache entry from the server and persist it to the local cacheDir for future cache hits. Fulfills STG-894.

- **New Features**
  - V3 auto-enables API cache streaming when cacheDir is set; successful runs persist the received entry locally.
  - StagehandAPIClient: new enableAgentCacheHeader option, request header x-stagehand-agent-cache-enabled, and consumeAgentCacheEntry() to retrieve streamed entries.
  - AgentCache: supports recording without storage, capturing the latest entry in memory, and persisting a serialized entry to disk.
  - Agent executions mark cache-eligible calls and validate cacheKey before persisting to prevent mismatches.

<sup>Written for commit 469aae841eaf46b4052cf4b54de6c5a782a03c6e. Summary will update on new commits.</sup>

### DIFF
--- a/packages/core/lib/v3/v3.ts
+++ b/packages/core/lib/v3/v3.ts
@@ -39,6 +39,7 @@ import {
   AgentReplayStep,
   InitState,
   AgentCacheContext,
+  SerializedAgentCacheEntry,
 } from "./types/private";
 import {
   AgentConfig,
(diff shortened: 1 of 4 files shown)


packages/core/lib/v3/v3.ts
```diff
@@ -440,6 +444,40 @@ export class V3 {
...
+  public setAgentCacheCapture(enabled: boolean): boolean {
+    return this.agentCache.setCaptureEntries(enabled);
+  }
+
+  public consumeCapturedAgentCacheEntry(
+    cacheKey?: string,
+  ): SerializedAgentCacheEntry | null {
+    return this.agentCache.consumeCapturedEntry(cacheKey);
+  }
+
```
tkattkat, line 456: add comment so we do not accidentally remove this in future

PR #1521, opened 2026-01-10
[fix]: add support for local caching of agent when using api (2)

seanmcguire12:
# why

# what changed

# test plan

---
## Summary by cubic
Adds local agent caching when executing via the API. When a cacheContext is provided, agent cache entries returned by the server are saved to local cache for replay and faster repeats.

- **New Features**
  - Extended AgentExecute (types + OpenAPI) with shouldCache and optional cacheEntry in the response.
  - API client captures cacheEntry from SSE and exposes consumeLatestAgentCacheEntry().
  - V3 persists remote cache entries to local CacheStorage when using API mode.
  - Added in-memory CacheStorage and a server agent cache handle to buffer and transfer cache entries.

<sup>Written for commit cda468f37f2b18b928eb057a400d0428a39b3e49. Summary will update on new commits.</sup>


packages/server/openapi.v3.yaml
```diff
@@ -76,6 +76,32 @@ components:
         id: $response.body#/data/sessionId
       description: End the session and release resources
   schemas:
+    AgentCacheContext:
```
monadoid: This should just be a boolean - something like `useCaching`. `AgentCacheContext` and `AgentCacheContextOutput` makes us stray even further from the REST/resource model. The main issue is that it leaks internals & makes future changes harder because clients now depend on cache mechanics.

Long-term we want an interface that’s basically just GET/POST resources — like `GET /products/{productId}` returns a product, not “a product + instructions for how the warehouse finds it”.

Rule of thumb: clients should describe *what they want*, not *how we optimize it internally*.

packages/core/lib/v3/cache/AgentCache.ts
```diff
@@ -368,6 +373,59 @@ export class AgentCache {
...
+    const entry = cloneForCache(payload.entry);
+    const { error, path } = await this.storage.writeJson(
+      `agent-${payload.cacheKey}.json`,
+      entry,
+    );
+    if (error && path) {
+      this.logger({
+        category: "cache",
+        message: "failed to import remote agent cache entry",
+        level: 1,
```
seanmcguire12: this should be 0

packages/server/openapi.v3.yaml
```diff
@@ -706,6 +719,11 @@ components:
           description: Whether to stream the response via SSE
           example: true
           type: boolean
+        shouldCache:
```
monadoid, line 728: `useCache`? `shouldCache` sounds to me like "client would prefer if caching was used here, but it might not get used" - probably better to use more standard language. `enableCache` maybe?

packages/server/openapi.v3.yaml
```diff
@@ -917,6 +937,20 @@ components:
         - success
         - data
       additionalProperties: false
+    AgentCacheEntryOutput:
+      type: object
+      properties:
+        cacheKey:
```
monadoid, line 955: Awesome, this is looking better! I think that we could go one step further and instead of calling it a cacheKey, calling it something that makes it sound like less of an implementation detail and more user-friendly  - `planKey`? `routineKey`?  In my opinion, this should be a first class citizen of stagehand, so we might want to look at api routes

seanmcguire12:
@greptileai

miguelg719 (approved):

miguelg719 (approved):

PR #1542, opened 2026-01-14
Fix (v3): global verbosity check in v3Logger console fallback

### LINKED ISSUE #1490: Bug : v3Logger fallback bypasses verbose level filtering, causing debug logs even with verbose: 0
## Bug Description

When using Stagehand V3 with `verbose: 0`, debug-level logs like `DEBUG: xpath main-world` are still printed to the console. The `v3Logger` function has a fallback path that bypasses verbose level filtering.

## Steps to Reproduce

```typescript
import { Stagehand } from "@browserbasehq/stagehand";

  env: "LOCAL",
  verbose: 0,  // Errors only
  disablePino: true,
});

await stagehand.init();
const page = stagehand.context.pages()[0];
await page.goto("https://example.com");
// Use any locator/xpath operation
await page.locator("//div").first();
```

## Expected Behavior

With `verbose: 0`, only error-level logs should be printed. Debug logs should be suppressed.

## Actual Behavior

Debug logs are printed:
```
[2026-01-02T13:16:37.661Z] DEBUG: xpath main-world
    frameId: F6611B8D017F9801E6E83A956066C58D
    xp: /html/body[1]/div[1]/div[3]/main[1]/div[1]/div[2]/form[1]/div[1]/input[2]
    ctxId: 3
```

## Root Cause Analysis

In `lib/v3/logger.ts`, the `v3Logger` function has two code paths:

1. **Instance logger path** (lines ~500-510): Calls the registered instance logger which respects verbose level
2. **Fallback path** (lines ~514-540): Directly logs to console **without checking verbose level**

The fallback path logs to console regardless of configured verbosity:
```typescript
if (lvl === 2) {
  (console.debug ?? console.log)(output);  // No verbose check!
}
```

Meanwhile, `StagehandLogger.log()` correctly checks: `if (logLine.level > this.verbose) return;`

## Suggested Fix

The fallback path in `v3Logger` should either:
1. Not log at all when no instance logger is registered (let it be silent)
2. Or check against a global/default verbose level before logging

## Environment

- Stagehand version: latest (V3)
- Node.js version: 22
- OS: Alpine Linux (Docker) / macOS

kr1shna-exe:
## Problem

When using Stagehand V3 with `verbose: 0`, debug-level logs are still printed if `v3Logger` falls back to console logging.The fallback path bypasses verbosity filtering, resulting in unexpected debug output.

## Solution

Applied a default verbosity threshold check to the fallback logging path to ensure debug messages are suppressed unless explicitly allowed.

## Changes

- Added a default fallback verbosity check in `v3Logger`
- Prevented debug-level logs from being printed when verbosity is effectively disabled

Fixes #1490

---
## Summary by cubic
Stops debug logs from printing when verbose: 0 by adding a fallback verbosity threshold (1) to the v3Logger console fallback. Fixes #1490.

<sup>Written for commit 36b82cc7ad3e7c2147b17fac4cc44230d006c193. Summary will update on new commits.</sup>

### DIFF
--- a/packages/core/lib/v3/logger.ts
+++ b/packages/core/lib/v3/logger.ts
@@ -31,6 +31,7 @@ import { AsyncLocalStorage } from "node:async_hooks";
 // Per-instance routing using AsyncLocalStorage
 const logContext = new AsyncLocalStorage<string>();
 const instanceLoggers = new Map<string, (line: LogLine) => void>();
+const FALLBACK_VERBOSITY = 1;
 
 export function bindInstanceLogger(
   instanceId: string,
@@ -76,6 +77,10 @@ export function v3Logger(line: LogLine): void {
   const lvl = line.level ?? 1;
   const levelStr = lvl === 0 ? "ERROR" : lvl === 2 ? "DEBUG" : "INFO";
   let output = `[${ts}] ${levelStr}: ${line.message}`;
+
+  if (lvl > FALLBACK_VERBOSITY) {
+    return;
+  }
 
   if (line.auxiliary) {
     for (const [key, { value, type }] of Object.entries(line.auxiliary)) {


seanmcguire12:
closing this one as we opted to just remove the noisy log instead. relevant PR [here](https://github.com/browserbase/stagehand/pull/1686)

Issue #1524, opened 2026-01-11
Feature Request: Expose Gemini 3 thinking/reasoning via includeThoughts in providerOptions


hoverlover:
## Summary

  When using Gemini 3 models with `stream: true`, reasoning/thinking output is never available in the `onChunk` callback or `reasoningText` field. This is because Stagehand hardcodes `providerOptions` for Gemini 3 and doesn't include `thinkingConfig: { includeThoughts: true }`.

  ## Current Behavior

  With `stream: true` and a Gemini 3 model, the `onChunk` callback only receives tool-related chunks:
  - `tool-input-start`
  - `tool-input-delta`
  - `tool-call`
  - `tool-result`

  No `text-delta` or `reasoning-delta` chunks are ever emitted. In `onStepFinish`, `reasoning` is always an empty array and `reasoningText` is always `undefined`.

  ## Expected Behavior

  Gemini 3's thinking/reasoning should be accessible via streaming when enabled. According to [Google's documentation](https://ai.google.dev/gemini-api/docs/thinking), this requires setting `includeThoughts: true` in the thinking config.

  ## Root Cause

  In `dist/index.js`, Stagehand hardcodes `providerOptions` for Gemini 3 models:

  ```js
  providerOptions: wrappedModel.modelId.includes("gemini-3") ? {
    google: {
      mediaResolution: "MEDIA_RESOLUTION_HIGH"
    }
  } : void 0
```

  This appears in two places (for streaming and non-streaming execution). The `thinkingConfig` with `includeThoughts: true` is not included.

  Proposed Solution

  Allow users to pass custom `providerOptions` through `AgentModelConfig`, or add `thinkingConfig` support:

  Option A: Extend `AgentModelConfig`

```js
  const agent = stagehand.agent({
    model: {
      modelName: "google/gemini-3-flash-preview",
      providerOptions: {
        google: {
          thinkingConfig: {
            includeThoughts: true,
            thinkingLevel: "medium"
          }
        }
      }
    },
    stream: true,
    mode: "hybrid",
  });
```

  Option B: Add dedicated thinking config

```js
  const agent = stagehand.agent({
    model: { modelName: "google/gemini-3-flash-preview" },
    stream: true,
    thinking: {
      includeThoughts: true,
      thinkingLevel: "medium"  // or thinkingBudget for Gemini 2.5
    }
  });
```

  Option C: Merge user-provided `providerOptions` with defaults

[...]

PR #1547, opened 2026-01-14
Customize thinking in agent

tkattkat:
## Provider Options for Agent Thinking/Reasoning

Addresses #1524

Added `providerOptions` configuration for `agent.execute()` that allows users to pass provider-specific thinking/reasoning options directly.

### Usage

Pass provider-specific options directly to enable thinking capabilities:

**Google Gemini:**
```typescript
  instruction: "Solve this complex problem",
  providerOptions: {
    google: {
      thinkingConfig: {
        includeThoughts: true,
        thinkingBudget: 10000
      }
    }
  }
});
```

**Anthropic Claude:**
```typescript
  instruction: "Solve this complex problem",
  providerOptions: {
    anthropic: {
      thinking: {
        type: "enabled",
        budgetTokens: 10000  // required, min 1024, max 64000
      }
    }
  }
});
```

**OpenAI:**
```typescript
  instruction: "Solve this complex problem",
  providerOptions: {
    openai: {
      reasoningSummary: "detailed",  // "auto" | "detailed" | "concise"
      reasoningEffort: "high"        // "low" | "medium" | "high"
    }
  }
});
```

### Available Options Per Provider

| Provider | Allowed Options |
|----------|-----------------|
| Google | `thinkingConfig: { includeThoughts?, thinkingBudget? }` |
| Anthropic | `thinking: { type: "enabled", budgetTokens } \| { type: "disabled" }` |
| OpenAI | `reasoningSummary?`, `reasoningEffort?` |

### Type Safety

The `providerOptions` field uses TypeScript `Pick` types to only allow thinking-related options:

- `GoogleThinkingOptions` - only `thinkingConfig`
- `AnthropicThinkingOptions` - only `thinking`
- `OpenAIThinkingOptions` - only `reasoningSummary` and `reasoningEffort`

Attempting to pass other provider options (e.g., `temperature`, `mediaResolution`) will result in a compile-time error.

### Notes

- **Experimental**: Requires `experimental: true` on Stagehand init
- **Not supported in CUA mode**: Throws `StagehandInvalidArgumentError` if used with `mode: "cua"`
- **Anthropic**: `budgetTokens` is required when `type: "enabled"` (min 1024, max 64000)
- **Full type safety**

### DIFF
--- a/packages/core/lib/v3/handlers/v3AgentHandler.ts
+++ b/packages/core/lib/v3/handlers/v3AgentHandler.ts
@@ -106,6 +106,16 @@ export class V3AgentHandler {
       return { openai: userOptions.openai };
     }
 
+    // Pass through xAI reasoning options directly
+    if (userOptions?.xai) {
+      return { xai: userOptions.xai };
+    }
+
+    // Pass through Groq reasoning options directly
+    if (userOptions?.groq) {
+      return { groq: userOptions.groq };
+    }
+
     return undefined;
   }
 
--- a/packages/core/lib/v3/types/private/agent.ts
+++ b/packages/core/lib/v3/types/private/agent.ts
@@ -1,6 +1,8 @@
 import type { AnthropicProviderOptions } from "@ai-sdk/anthropic";
 import type { GoogleGenerativeAIProviderOptions } from "@ai-sdk/google";
+import type { GroqProviderOptions } from "@ai-sdk/groq";
 import type { OpenAIResponsesProviderOptions } from "@ai-sdk/openai";
+import type { XaiProviderOptions } from "@ai-sdk/xai";
 
 export interface ActionMappingOptions {
   toolCallName: string;
@@ -19,9 +21,13 @@ export interface AgentProviderOptions {
   google?: GoogleGenerativeAIProviderOptions;
   anthropic?: AnthropicProviderOptions;
   openai?: OpenAIResponsesProviderOptions;
+  xai?: XaiProviderOptions;
+  groq?: GroqProviderOptions;
   [key: string]:
     | GoogleGenerativeAIProviderOptions
     | AnthropicProviderOptions
     | OpenAIResponsesProviderOptions
+    | XaiProviderOptions
+    | GroqProviderOptions
     | undefined;
 }
--- a/packages/core/lib/v3/types/public/agent.ts
+++ b/packages/core/lib/v3/types/public/agent.ts
@@ -14,7 +14,9 @@ import {
 } from "ai";
 import type { AnthropicProviderOptions } from "@ai-sdk/anthropic";
 import type { GoogleGenerativeAIProviderOptions } from "@ai-sdk/google";
+import type { GroqProviderOptions } from "@ai-sdk/groq";
 import type { OpenAIResponsesProviderOptions } from "@ai-sdk/openai";
+import type { XaiProviderOptions } from "@ai-sdk/xai";
 import { LogLine } from "./logs";
 import { ClientOptions } from "./model";
 import { StagehandZodObject } from "../../zodCompat";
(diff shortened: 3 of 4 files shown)


packages/core/lib/v3/handlers/v3AgentHandler.ts
```diff
@@ -143,6 +259,24 @@ export class V3AgentHandler {
...
+    if (event.reasoningText && event.reasoningText.length > 0) {
+      return event.reasoningText;
+    }
+    // Fallback: regular text output
+    if (event.text && event.text.length > 0) {
+      return event.text;
+    }
+
+    return null;
+  }
```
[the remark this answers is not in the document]
tkattkat: this is intended brother

packages/core/lib/v3/handlers/v3AgentHandler.ts
```diff
@@ -69,6 +76,115 @@ export class V3AgentHandler {
...
+    thinkingConfig?: ThinkingConfig,
+  ): { options: AgentProviderOptions | undefined; suppressWarnings: boolean } {
+    const isGoogle = modelId.includes("gemini");
+    const isAnthropic = modelId.includes("claude");
+    const isOpenAI =
+      modelId.includes("gpt-") ||
+      modelId.includes("o1") ||
+      modelId.includes("o3") ||
+      modelId.includes("o4");
+    const isGemini3 = modelId.includes("gemini-3");
```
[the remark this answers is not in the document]
tkattkat: intentional

pirate (commented):
I would recommend just passing through the config params to the LLM providers without trying to normalize the names/levels. as there are more LLM providers over time with more drift between them, we dont want to create a neverending stream of normalization work for ourselves. better to just document the recommended configs, but pass through 100% of the provider config options without trying to interpret them.

packages/core/lib/v3/handlers/v3AgentHandler.ts
```diff
@@ -69,6 +72,45 @@ export class V3AgentHandler {
...
+    }
+
+    // Pass through Anthropic thinking options directly
+    if (userOptions?.anthropic) {
+      return { anthropic: userOptions.anthropic };
+    }
+
+    // Pass through OpenAI reasoning options directly
+    if (userOptions?.openai) {
+      return { openai: userOptions.openai };
```
pirate, line 106: is there any way we can do these generically based on provider name without having to hardcode specific providers here?

e.g. just `return userOptions`?

packages/core/lib/v3/handlers/v3AgentHandler.ts
```diff
@@ -156,21 +216,24 @@ export class V3AgentHandler {
         level: 2,
       });
 
+      // Capture reasoning from various sources (Google thinkingConfig, Anthropic thinking, OpenAI reasoning)
+      // The AI SDK provides reasoningText and reasoning array for models with thinking enabled
+      const stepReasoning = this.extractReasoningFromStep(event);
+      if (stepReasoning) {
+        state.collectedReasoning.push(stepReasoning);
+        this.logger({
```
pirate, line 232: can you check if flowLogger is capturing reasoning correctly as well?

packages/core/lib/v3/types/private/agent.ts
```diff
@@ -1,6 +1,27 @@
...
   reasoning?: string;
 }
+
+/**
+ * Internal provider options type used by the AI SDK.
+ * This is the full provider options type used internally after merging
+ * user's thinking options with internal defaults (e.g., mediaResolution for Gemini 3).
+ * @internal
+ */
+export interface AgentProviderOptions {
```
pirate, line 20: can we add all of the ones aisdk supports? ollama, deepseek, vertex, bedrock, etc.

even if they aren't supported now I'd rather not have to add them one by one as they become supported.

tkattkat: yea

akeimach:
Closing this as part of a cleanup of old open PRs after the v4 release.

PR #1553, opened 2026-01-15
x-sent-at removal, x-stainless-any fix, optional frame id, ModelConfigString fix

monadoid:
# why
Fixes for having openapi spec pass stainless checks.
# what changed

# test plan

### DIFF
--- a/packages/server/openapi.v3.yaml
+++ b/packages/server/openapi.v3.yaml
@@ -142,11 +142,11 @@ components:
           type: boolean
       additionalProperties: false
     ModelConfigString:
-      description:
-        Model name string with provider prefix (e.g., 'openai/gpt-5-nano',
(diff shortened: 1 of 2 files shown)


packages/server/openapi.v3.yaml
```diff
@@ -161,9 +161,9 @@ components:
             - google
             - microsoft
         modelName:
-          description: Model name string (e.g., 'openai/gpt-5-nano',
-            'anthropic/claude-4.5-opus')
-          example: openai/gpt-5-nano
+          description: Model name string without prefix (e.g., 'gpt-5-nano',
```
pirate: Nah keep prefix recommended in all possible shapes, API can handle it.

pirate (approved):

PR #1560, opened 2026-01-16
Adjust coordinate handling

tkattkat:
# Why

Google's computer use models return coordinates in a normalized 0-1000 range, which must be converted to actual pixel coordinates for browser interactions. Previously, this normalization used hardcoded viewport dimensions that didn't account for:

1. **User-configured viewports** - Users can specify custom viewport sizes in `localBrowserLaunchOptions` or `browserbaseSessionCreateParams`
2. **advancedStealth mode**

This caused coordinate misalignment issues where clicks would land in the wrong location, especially when using non-default viewport sizes or advancedStealth mode.

---

# What Changed

### Core Changes

**`v3.ts`**
- Added `isAdvancedStealth` getter to check if advancedStealth is enabled in Browserbase settings
- Added `configuredViewport` getter to retrieve viewport dimensions from user config (with 1288x711 fallback)

**`coordinateNormalization.ts`**
- Now accepts `v3` instance to access configuration
- Uses `v3.configuredViewport` for normal mode (respects user's viewport settings)
- Uses hardcoded `1288x711` for advancedStealth mode (matches what stealth produces)

**`v3CuaAgentHandler.ts`**
- Updated `updateClientViewport()` to set correct dimensions for Google CUA:
  - advancedStealth → 1288x711
  - normal → configured viewport
- Removed unnecessary `getPNGDimensions` function and `setScreenshotSize` calls

**`GoogleCUAClient.ts`**
- Simplified `normalizeCoordinates()` to use just `currentViewport` (no more screenshot size tracking)
- Removed `actualScreenshotSize` property and `setScreenshotSize` method

### Cleanup

**`OpenAICUAClient.ts` & `MicrosoftCUAClient.ts`**
- Removed `actualScreenshotSize` property and `setScreenshotSize` method
- Reverted coordinate scaling to use original viewport-based logic

### Tool Updates

Updated all coordinate-using tools to pass `v3` to `processCoordinates()`:
- `click.ts`
- `type.ts`
- `scroll.ts`
- `clickAndHold.ts`
- `dragAndDrop.ts`
- `fillFormVision.ts`

---

# Test Plan

### Manual Testing

- [ ] **Local with default viewport**: Run agent task, verify clicks land correctly
- [ ] **Local with custom viewport** (e.g., 1920x1080): Verify coordinate normalization uses configured size
- [ ] **Browserbase without advancedStealth**: Verify uses configured viewport
- [ ] **Browserbase with advancedStealth**: Verify uses 1288x711 for normalization

---
## Summary by cubic
Fixes misaligned clicks by converting Google CUA’s 0–1000 coordinates to the correct viewport: user-configured in normal mode and fixed 1288×711 in advancedStealth. Simplifies the path from model coordinates to browser actions for consistent interactions.

- **Bug Fixes**
  - Use the configured viewport (local or Browserbase) for normalization.
  - Use 1288×711 when advancedStealth is enabled to match stealth behavior.
  - Apply the correct viewport to Google CUA so clicks, scrolls, and drags land accurately.

- **Refactors**
  - Removed screenshot-size tracking and PNG parsing; clients now scale from currentViewport only.
  - Added v3.isAdvancedStealth and v3.configuredViewport; coordinate utils now consume v3.
  - All coordinate-using tools pass v3 to processCoordinates; handler sets Google CUA viewport based on mode.

<sup>Written for commit 18fcf17cb6563e95c15bf248cbf6905869685d5b. Summary will update on new commits.</sup>

### DIFF
--- a/packages/core/lib/v3/handlers/v3AgentHandler.ts
+++ b/packages/core/lib/v3/handlers/v3AgentHandler.ts
@@ -155,11 +155,6 @@ export class V3AgentHandler {
         message: `Step finished: ${event.finishReason}`,
         level: 2,
       });
-      this.logger({
-        category: "agent",
-        message: `reasoning: ${event.text}`,
-        level: 1,
-      });
 
       if (event.toolCalls && event.toolCalls.length > 0) {
         for (let i = 0; i < event.toolCalls.length; i++) {
@@ -169,6 +164,11 @@ export class V3AgentHandler {
 
           if (event.text && event.text.length > 0) {
             state.collectedReasoning.push(event.text);
+            this.logger({
+              category: "agent",
+              message: `reasoning: ${event.text}`,
+              level: 1,
+            });
           }
 
           if (toolCall.toolName === "close") {


packages/core/lib/v3/v3.ts
```diff
@@ -172,6 +172,42 @@ export class V3 {
...
     return this.state.kind === "BROWSERBASE";
   }
+
+  /**
+   * Returns true if advancedStealth is enabled in Browserbase settings.
+   */
+  public get isAdvancedStealth(): boolean {
+    return (
+      this.opts.browserbaseSessionCreateParams?.browserSettings
+        ?.advancedStealth === true
```
miguelg719, line 182: this might not return an accurate value whenever the session is created with the bb sdk and the session id is passed into stagehand (just fyi not blocking)

miguelg719 (approved):

packages/core/lib/v3/v3.ts
tkattkat: ah shit good point

PR #1562, opened 2026-01-16
feat(cache): add custom tool caching and fix replay navigation timing

### LINKED ISSUE #1558: Agent cache does not record or replay custom tool calls
## Problem

When using agent caching with custom tools (via the `tools` option in `agent()`), custom tool executions are not recorded in the cache and are skipped during replay. This causes agent replay to fail when custom tools perform essential actions like form filling with credentials.

## Reproduction

1. Create an agent with custom tools (e.g., for filling login credentials)
2. Enable caching via `cacheDir` option
3. Execute the agent - custom tools work correctly
4. Re-run the same instruction (cache hit) - custom tool calls are skipped

## Root Cause

I traced through the source code and found two issues:

### 1. Custom tools are not recorded during execution

In `v3CuaAgentHandler.ts`, the `custom_tool` case just returns success without recording:

```typescript
case "custom_tool": {
  // Custom tools are handled by the agent client directly
  return { success: true };
}
```

Compare this to other action types like `goto`, `scroll`, `wait` which all call `this.v3.recordAgentReplayStep()`.

### 2. Custom tools are not replayed

In `AgentCache.ts`, the `executeAgentReplayStep` method's switch statement doesn't handle `custom_tool`:

```typescript
switch (step.type) {
  case "act": ...
  case "fillForm": ...
  case "goto": ...
  // ... other cases ...
  default:
    this.logger({ message: `agent cache skipping step type: ${step.type}` });
    return step;  // Custom tools fall through here and are skipped
}
```

### 1. Add new type for custom tool steps

In `types/private/cache.ts`:

```typescript
export interface AgentReplayCustomToolStep {
  type: "custom_tool";
  name: string;
  arguments: Record<string, unknown>;
}

export type AgentReplayStep =
  | AgentReplayActStep
  // ... existing types ...
  | AgentReplayCustomToolStep
  | { type: string; [key: string]: unknown };
```

### 2. Record custom tool executions

In `v3CuaAgentHandler.ts` or where the action is processed after tool execution:

```typescript
case "custom_tool": {
  if (recording) {
    this.v3.recordAgentReplayStep({
      type: "custom_tool",
      name: action.name as string,
      arguments: action.arguments as Record<string, unknown>,
    });
  }
  return { success: true };
}
```

### 3. Add replay support for custom tools

This requires passing the `tools` object to `AgentCache` so it can re-execute tools during replay:

```typescript
private async replayAgentCustomToolStep(
  step: AgentReplayCustomToolStep,
  tools: ToolSet,
): Promise<void> {
  const tool = tools[step.name];
  if (tool) {
    await tool.execute(step.arguments, { toolCallId: `replay_${Date.now()}`, messages: [] });
  }
}
```

## Use Case

I'm building a system that uses custom tools for filling forms with runtime credentials. Without custom tool caching, the cache records button clicks but skips the credential filling, causing login failures during replay.

## Willingness to Contribute

I'm happy to submit a PR implementing this fix if the approach looks reasonable to the maintainers.

### LINKED ISSUE #1561: Agent cache replay executes steps too fast, causing navigation failures
## Description

When replaying cached agent executions, steps are executed immediately after each other without waiting for page navigation to complete. This causes subsequent steps to run on the wrong page, leading to failed workflows.

## Steps to Reproduce

1. Record an agent workflow that includes navigation (e.g., login → click link → search)
2. Run the workflow again to trigger cache replay
3. Observe that steps after navigation-triggering actions fail or execute on the wrong page

## Expected Behavior

After an action triggers page navigation (detected by URL change), replay should wait for the page to fully load before executing the next step.

## Actual Behavior

Replay proceeds immediately to the next step without waiting for navigation, causing:
- Actions to execute on the previous page
- XPath selectors to fail or match wrong elements
- Workflow failures despite having valid cached data

## Root Cause

In `AgentCache.ts`, the replay methods (`replayAgentActStep`, `replayAgentFillFormStep`, `replayAgentKeysStep`) execute cached actions but don't wait for navigation to complete before continuing to the next step.

## Proposed Fix

After executing each action during replay:
1. Capture URL before action
2. Execute the action
3. Check URL after action
4. If URL changed, call `page.waitForLoadState("load")` before continuing

This ensures the page is fully loaded before the next step executes.

## Related

This issue was discovered while testing the fix for #1558 (custom tool caching). Both issues affect agent cache replay reliability.

hoverlover:
## Summary

This PR adds two improvements to the agent cache replay system:

### 1. Custom Tool Caching (fixes #1558)

Custom tools (like form-filling helpers) were not being recorded or replayed from cache, causing workflows to fail on cache replay.

**Changes:**
- Add `AgentReplayCustomToolStep` type for caching custom tool invocations
- Wrap custom tools with recording logic via `wrapToolsForRecording()`
- Record custom tool calls in both AISDK (hybrid/dom) and CUA agent modes
- Replay custom tools by re-executing with cached arguments
- Thread `tools` parameter through `tryReplay` and related methods

### 2. Navigation Waiting (fixes #1561)

Cache replay was executing steps too fast without waiting for page navigation to complete, causing subsequent steps to run on the wrong page.

**Changes:**
- Detect URL changes after actions during replay
- Wait for page load (`waitForLoadState`) before continuing to next step
- Apply to `replayAgentActStep`, `replayAgentFillFormStep`, `replayAgentKeysStep`
- Prevents race conditions where steps execute before navigation completes

## Test Plan

Tested with a real-world workflow:
- [x] Login flow with custom `fillUsername` and `fillPassword` tools
- [x] Search flow with custom `fillSearch` tool
- [x] Multi-page navigation (login → home → search → product page)
- [x] Verified 0 token usage on cache replay (no LLM calls)
- [x] Verified consistent results across 3+ replay runs

## Files Changed

- `packages/core/lib/v3/types/private/cache.ts` - Add `AgentReplayCustomToolStep` type
- `packages/core/lib/v3/v3.ts` - Add `wrapToolsForRecording()`, update tool threading
- `packages/core/lib/v3/handlers/v3CuaAgentHandler.ts` - Add custom tool recording for CUA mode
- `packages/core/lib/v3/cache/AgentCache.ts` - Add custom tool replay, add navigation waiting

---
## Summary by cubic
Add custom tool caching to agent replay and wait for navigation between steps. This fixes tool-based workflows and prevents steps from running on the wrong page.

- **New Features**
  - Record and replay custom tool calls as custom_tool steps.
  - Wrap tools to auto-record; replay re-executes with cached args.
  - Pass tools through tryReplay and stream replay.

- **Bug Fixes**
  - Detect URL changes during replay and wait for page load.
  - Applied to act, fillForm, and keys (e.g., Enter submits forms).
  - Prevents steps executing on the wrong page.
  - Clean up recording state if stream creation fails.

<sup>Written for commit 36d32749ba33e103aad6067b0bd991f9020c71e7. Summary will update on new commits.</sup>

### DIFF
--- a/packages/core/lib/v3/cache/AgentCache.ts
+++ b/packages/core/lib/v3/cache/AgentCache.ts
@@ -1,8 +1,10 @@
 import { createHash } from "crypto";
+import type { ToolSet } from "ai";
 import type { ActHandler } from "../handlers/actHandler";
 import type { LLMClient } from "../llm/LLMClient";
 import type {
   AgentReplayActStep,
+  AgentReplayCustomToolStep,
   AgentReplayFillFormStep,
   AgentReplayGotoStep,
   AgentReplayKeysStep,
@@ -157,6 +159,7 @@ export class AgentCache {
   async tryReplay(
     context: AgentCacheContext,
     llmClientOverride?: LLMClient,
+    tools?: ToolSet,
   ): Promise<AgentResult | null> {
     if (!this.enabled) return null;
 
@@ -192,7 +195,12 @@ export class AgentCache {
       },
     });
 
-    return await this.replayAgentCacheEntry(context, entry, llmClientOverride);
+    return await this.replayAgentCacheEntry(
+      context,
+      entry,
+      llmClientOverride,
+      tools,
+    );
   }
 
   /**
@@ -214,8 +222,9 @@ export class AgentCache {
   async tryReplayAsStream(
     context: AgentCacheContext,
     llmClientOverride?: LLMClient,
+    tools?: ToolSet,
   ): Promise<AgentStreamResult | null> {
-    const result = await this.tryReplay(context, llmClientOverride);
+    const result = await this.tryReplay(context, llmClientOverride, tools);
     if (!result) return null;
     return this.createCachedStreamResult(result);
   }
@@ -503,6 +512,7 @@ export class AgentCache {
     context: AgentCacheContext,
     entry: CachedAgentEntry,
     llmClientOverride?: LLMClient,
+    tools?: ToolSet,
   ): Promise<AgentResult | null> {
     const ctx = this.getContext();
     const handler = this.getActHandler();
@@ -518,6 +528,7 @@ export class AgentCache {
             ctx,
             handler,
             effectiveClient,
+            tools,
           )) ?? step;
         stepsChanged ||= replayedStep !== step;
         updatedSteps.push(replayedStep);
@@ -557,6 +568,7 @@ export class AgentCache {
     ctx: V3Context,
     handler: ActHandler,
     llmClient: LLMClient,
+    tools?: ToolSet,
   ): Promise<AgentReplayStep> {
     switch (step.type) {
       case "act":
@@ -588,6 +600,12 @@ export class AgentCache {
       case "keys":
         await this.replayAgentKeysStep(step as AgentReplayKeysStep, ctx);
         return step;
+      case "custom_tool":
+        await this.replayAgentCustomToolStep(
+          step as AgentReplayCustomToolStep,
+          tools,
+        );
+        return step;
       case "close":
       case "extract":
       case "screenshot":
@@ -614,6 +632,9 @@ export class AgentCache {
       const page = await ctx.awaitActivePage();
       const updatedActions: Action[] = [];
       for (const action of actions) {
+        // Capture URL before action to detect navigation
+        const urlBefore = page.url();
+
         const result = await handler.takeDeterministicAction(
           action,
           page,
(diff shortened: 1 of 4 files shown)


hoverlover:
## Response to Review Comments

### Re: Using `wrappedTools` instead of `tools` in `tryReplay()` calls

After analysis, I believe the current code is correct and should **not** be changed to use `wrappedTools`. Here's why:

**During replay**, `tryReplay()` executes custom tools with their cached arguments via `replayAgentCustomToolStep()`, which calls `tool.execute()`. 

If we passed `wrappedTools`:
- The recording wrapper would intercept these replayed tool calls
- Each replayed call would be recorded as a new step
- This would cause duplicate entries if the cache is later updated

**The current flow is correct:**
1. **Cache miss (initial execution)**: Handler uses `wrappedTools` → tool invocations are recorded → steps stored in cache
2. **Cache hit (replay)**: `tryReplay()` uses unwrapped `tools` → cached tools execute without recording → no duplicates

I've added inline comments at all three locations to clarify this intentional design decision (commit 36d32749).

---

### Fixed: Recording state cleanup (P2)

Added try-catch around `handler.stream()` to call `discardAgentReplayRecording()` on error, preventing the cache from being stuck in recording mode if stream creation fails (commit 63a6d526).

rastc:
Any update on this?

PR #1578, opened 2026-01-20
End endpoint cleanup

monadoid:
# why
`stagehand-api` was updated so that no request body should be send when calling the `/end` endpoint. 

# what changed
Thus, we can simplify our `stagehand-server` code so that no request body should be sent as well, cleaning up our stainless sdks.

Also made it so that when we want to release a new github release of `stagehand-server`, we add in a minor or major changeset - no longer having to bump the `package.json` version manually. 
# test plan
Added a minor changeset for `stagehand-server` and will confirm new github release.

---
## Summary by cubic
Removed the empty JSON body requirement for POST /v1/sessions/:id/end. The endpoint now accepts no request body, and the release workflow uses Changesets instead of manual version bumps.

- **Refactors**
  - Removed the SessionEnd request body from server route and OpenAPI; SDK types now allow no body.
  - Validate that the request body is empty; allow 0-byte JSON and return 400 for non-empty or non-object bodies. Tests assert 200 with no body or empty JSON.
  - Updated the GitHub release workflow to detect server releases via Changesets (no package.json bump needed).
  - Added port to browser launch options in the OpenAPI spec.

- **Migration**
  - Call POST /v1/sessions/:id/end with no body; sending Content-Type: application/json with an empty body is allowed.

<sup>Written for commit 975ccb6da02a782078590424a44640f1a5b01d1c. Summary will update on new commits.</sup>

### DIFF
--- a/packages/server/src/routes/v1/sessions/_id/end.ts
+++ b/packages/server/src/routes/v1/sessions/_id/end.ts
@@ -14,6 +14,27 @@ const endRouteHandler: RouteHandlerMethod = withErrorHandling(
       return error(reply, "Unauthorized", StatusCodes.UNAUTHORIZED);
     }
 
+    // This endpoint intentionally has no request body. Reject unexpected bodies to
+    // catch misconfigured clients, while still allowing empty JSON bodies.
+    const body = (request as { body?: unknown }).body;
+    if (body != null) {
+      if (typeof body !== "object" || Buffer.isBuffer(body)) {
+        return error(
+          reply,
+          "Request body must be empty",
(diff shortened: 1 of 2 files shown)


pirate (changes requested):
This will break on local because it doesn't include my fix to actually allow empty body.

https://github.com/browserbase/core/pull/6885/changes#diff-4927448eb53ff99f86d3473fdf1d2e5e61a3bee9831b7adf3c0d0198a7661fc4

pirate (approved):

PR #1611, opened 2026-01-26
Using 'mode' enum instead of old 'cua' boolean

monadoid:
# why
Discord user reported that the `mode` property isn't typed in the Python SDK, which was due to the OpenAPI spec being out of date 
# what changed
Replaced `cua` boolean field with `mode` enum in zod objects + regenerated openapi spec.
# test plan

---
## Summary by cubic
Updated agent config to use a mode enum (dom, hybrid, cua) instead of the old cua boolean, and regenerated the OpenAPI spec. Kept cua for backward compatibility but deprecated; mode takes precedence if both are set.

- **Migration**
  - Use agentConfig.mode; cua is deprecated but still accepted. If both are set, mode overrides cua.
  - Valid values: dom, hybrid, cua.

<sup>Written for commit 9efc0db26760b9ca3d168cb189147f34b32dd8e0. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1611">Review in cubic</a></sup>

### DIFF
--- a/packages/server/test/integration/v3/agentExecute.test.ts
+++ b/packages/server/test/integration/v3/agentExecute.test.ts
@@ -1033,4 +1033,32 @@ describe("POST /v1/sessions/:id/agentExecute - validation errors (V3)", () => {
       ctx,
     );
   });
+
+  it("should return 400 for invalid agentConfig.mode", async () => {
+    const url = getBaseUrl();
+
+    const ctx = await fetchWithContext<{
+      success?: boolean;
+      error?: string;
+      message?: string;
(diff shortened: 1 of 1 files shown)


.changeset/shy-clouds-type.md
```diff
@@ -0,0 +1,6 @@
+---
+"@browserbasehq/stagehand-server": minor
```
shrey150, line 2: Just confirming -- is this `minor` b/c we're updating the API contract? If so, makes sense to me

monadoid: Yup!

monadoid:
@cubic-dev-ai

Issue #1585, opened 2026-01-21
screenshot mask locators only masking first found item


theoephraim:
## Environment Information

**Stagehand:**

- Language/SDK: TypeScript
- Stagehand version: 3.0.7

## Issue Description

It seem that when using the `mask` option on `page.screenshot`, that only the first matching element is masked.

If this is intended behaviour, it certainly was not expected. There does not seem to be any way to explicitly opt into masking all matching elements.

### Steps to Reproduce

1. use `page.screenshot` with an overly broad locator (e.g., `mask: page.locator('input')`)
2. call page.screenshot on a page with multiple matching elements
3. save the screenshot to a file to observe
4. note that only the first input is masked

### Minimal Reproduction Code

I will try to set up a minimal repro and edit this soon.

seanmcguire12:
hey @theoephraim this is definitely a bug, thanks for flagging! working on a fix now

seanmcguire12:
going to close this one @theoephraim since #1612 was merged? should be available on the alpha release now

PR #1612, opened 2026-01-26
[fix]: `mask` param in `page.screenshot()` only applied to first node

seanmcguire12:
# why
- fixes issue #1585 
- the `mask` param should be applied to all nodes that the locator resolves to (unless `.nth()` is specified) 
- currently, it is only being applied to the first node
# what changed
- renamed `resolveMaskRect()` to `resolveMaskRects()`, and updated behaviour to resolve all bounding rects instead of just 1
- added `locator.resolveNodesForMask()` which is called by `resolveMaskRects()` and resolves all nodes & object IDs so they can be iterated through, and have masks applied to them
# test plan
- extended tests in `page-screenshot.spec.ts` to cover the case where multiple elements match the `mask` locator

### DIFF
--- a/packages/core/lib/v3/understudy/frameLocator.ts
+++ b/packages/core/lib/v3/understudy/frameLocator.ts
@@ -155,7 +155,7 @@ class LocatorDelegate {
     const value = Number(index);
     if (!Number.isFinite(value) || value < 0) {
       throw new StagehandInvalidArgumentError(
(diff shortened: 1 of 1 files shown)


seanmcguire12:
@greptileai

seanmcguire12:
@cubic-dev-ai

seanmcguire12:
@cubic-dev-ai

Issue #1616, opened 2026-01-27
mask selector not working within an open `dialog`


theoephraim:
## Environment Information

**Stagehand:**

Language/SDK: TypeScript
Stagehand version: 3.0.7

## Issue Description

Masking selectors are not working within an open `dialog` component. As in, there is a dialog that has been opened with `.showModal()`.
It does not matter if the selector targets the input exactly using an ID.
Seems to be something related to how the browser moves this into the `#top-layer` pseudo element...

## Repro
```
import { $ } from 'bun';
import * as fs from 'fs/promises';
import { Stagehand } from '@browserbasehq/stagehand';

const stagehand = new Stagehand({
  env: 'LOCAL',
});
await stagehand.init();

let page = stagehand.context.pages()[0];
if (!page) throw new Error('Expected browser page');

const sensitiveSelectors = [
  '#dialog-input'
];

// Define the HTML we'll serve
const html = `
<!DOCTYPE html>
<html><head><title>Agent Mask Test</title>
<script>
document.addEventListener("DOMContentLoaded", () => {
  document.getElementById('dialog').showModal();
});
</script>

</head><body>
  <h2>Testing Masked Input</h2>
  <dialog id="dialog">
    <input id="dialog-input" readonly value="input in dialog" />
  </dialog>
</body></html>
`;

const port = 6789;
const localhostUrl = `http://127.0.0.1:${port}/`;

const server = Bun.serve({
  port,
  fetch(req) {
    return new Response(html, {
      headers: { "Content-Type": "text/html" }
    });
  }
});


console.log(`Test server running at ${localhostUrl}`);

const oldScreenshot = page.screenshot;
page.screenshot = async function (options?: Parameters<typeof page.screenshot>[0]) {
  console.log('screenshot');
  options ??= {};
  options.mask = sensitiveSelectors.map((s) => page.locator(s));
  const screenshot = await oldScreenshot.call(this, options);
  await fs.writeFile('screenshot.png', screenshot);

  return screenshot;
};

await page.goto(localhostUrl);
await page.screenshot();

await stagehand.close();
await server.stop();
```

seanmcguire12:
fix merged in #1624

PR #1624, opened 2026-01-27
[fix]: screenshot mask not working on `<dialog>` elements

seanmcguire12:
# why
- top‑layer UI (like `<dialog>` and popovers) is painted above the normal document stacking context, so a mask appended to `document.body` can never cover it. in order to mask dialog content, the overlay must be inserted _inside_ the top‑layer container itself
- the screenshot masking logic for elements inside `<dialog>` nodes (and other top‑layer elements) didn’t work because the overlay divs were always appended to the document root, which renders _underneath_ the dialog
- this PR addresses #1616 
# what changed
- updated logic in `resolveMaskRect()` to detect top‑layer roots (dialog[open], :popover-open), compute rects relative to that root, and return a `rootToken` so the caller can target the correct container
- updated `applyMaskOverlays()` to insert mask overlays inside the top‑layer root when a `rootToken` is present, and to restore any temporary position changes on cleanup. `resolveMaskRectForObject()` now passes a `maskToken` and preserves the `rootToken` output
# test plan
- added another test in `page-screenshot.spec.ts` to check whether masks are correctly injected into the dialog node

---
## Summary by cubic
Fixes screenshot masking for elements inside dialog and other top‑layer UIs by inserting overlays into the top‑layer root instead of the document root. Masks now correctly cover dialogs and are fully cleaned up after the screenshot.

- Bug Fixes
  - Detect top‑layer roots (dialog[open], :popover-open) in resolveMaskRect, compute rects relative to that root, and return a rootToken.
  - Insert mask overlays inside the detected top‑layer container; temporarily adjust position when needed and restore on cleanup.
  - Thread a maskToken/rootToken through masking calls to group overlays and handle multi-frame cases safely.
  - Add a test to confirm masks are injected into dialog elements.

<sup>Written for commit 4781745cf466238cecc04f9ebc093ece2d8e1126. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1624">Review in cubic</a></sup>

### DIFF
--- a/.changeset/fifty-gifts-win.md
+++ b/.changeset/fifty-gifts-win.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+fix issue where screenshot masks were not being applied to dialog elements


theoephraim:
One note, although not sure if relevant - I think you can have a `dialog[open]` without having opened it via `.showModal()`, in which case I dont think it's in the top layer.

seanmcguire12:
@theoephraim good call out! i did test locally on dialogs opened via both `show()` & `showModal()`. correct me if im wrong here but i believe these are the only two ways to open a dialog

theoephraim:
Just if someone (or a framework) wrote html that had a dialog and used the open property directly. Your code might assume it’s in the top layer, but calling out that you may need to verify that assumption.

seanmcguire12:
@theoephraim that makes sense! probably could have been more clear in my above comment, but was trying to explain that this PR fixes the mask issue for both cases. ie, after this is merged, the mask will work on dialogs regardless of whether the dialog was opened via `show()` or `showModal()`

PR #1632, opened 2026-01-28
[STG-1232] Speed up PR Github Actions checks, add code coverage, fix e2e bb tests running not running in bb mode, remove tsup, generate unified re-used ESM and CJS builds with sourcemaps and types, switch CI to new Evals CLI

pirate:
# why

- unit tests, e2e integration tests, and evals were taking 10~15 minutes per pr
- finding the specific failing eval or test was tedious and required reading tons of log output
- we weren't leveraging pnpm or turbo caches or github action's native chromium install
- tests were flaky because CDP would randomly fail to connect when github actions runner was overloaded
- e2e bb tests were not actually running against bb, so it was previously just double-running e2e:local and reporting it as bb
- tsup has been deprecated for a while and should be replaced by esbuild

# what changed

- parallelized all CI tests and broke out evals and tests into individual matrix jobs
- created unified node+pnpm+turbo setup action that all other actions call
- verified vanilla chromium is able to launch inside all the e2e local jobs before running any of our own code to catch resource contention flakyness
- add proper esm and csj builds with sourcemaps and types for aligned coverage
- remove tsup dependency
- switch to new evals cli
- add coverage reporting
- fix some flaky tests needed to get everything passing reliably

Screenshot 2026-01-26 at 1 04 04 PM

---
## Summary by cubic
Speeds up CI and makes tests stable by running Vitest/Playwright/Evals on the built ESM dist with cached artifacts, verified runner Chromium, weighted Browserbase routing, and merged coverage. Publishes a unified ESM+CJS SDK with sourcemaps/types and SEA binaries from ESM; fixes Browserbase e2e, switches to the new Evals CLI, and adds flaky‑test reporting, addressing Linear STG‑1232.

- **New Features**
  - CI: split pnpm/Turbo caches; composite actions (setup/caches, Chromium launch check, region selection, CTRF/V8 uploads); reuse artifacts; use Actions Chromium via CHROME_PATH with --no-sandbox; unique CTRF reports; V8 coverage normalization/merge with cancellation safety; push triggers enabled.
  - Tests: run from dist/esm; env reporter; true Browserbase runs with weighted/randomized regions; hard timeouts for LLM and session create/CDP connect; faster multi‑click dispatch; higher timeouts and targeted skips; cleanup helpers; fixes for init‑script, wait‑for‑selector, and screenshot reliability.
  - Build/SDK/SEA: export map (ESM import + CJS require) with sourcemaps/types; SEA binaries built from ESM with sourcemaps and plain logs in CI; removed tsup and dotenv; AISdk/CustomOpenAI moved to lib and exported; getAISDKLanguageModel added; replay metrics now include method/parameters/result/timestamps/cost.

- **Migration**
  - Set STAGEHAND_BROWSER_TARGET=local|browserbase, STAGEHAND_ENV (replaces TEST_ENV), CHROME_PATH; tune LLM_MAX_MS, BROWSERBASE_SESSION_CREATE_MAX_MS, BROWSERBASE_CDP_CONNECT_MAX_MS; optionally set BROWSERBASE_REGION_DISTRIBUTION.
  - Remove dotenv; set env vars directly and import @browserbasehq/stagehand via the export map (avoid dist paths).

<sup>Written for commit 5c76fecf6b9016f4ad5273d4143c3fe747b5d06f. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1632">Review in cubic</a></sup>

---
## Summary by cubic
Ships an ESM‑first SDK with a CJS entry, unified sourcemaps, and types. CI runs Vitest/Playwright and the new Evals CLI against the built ESM dist using the runner’s Chromium (launch‑verified), with region‑weighted Browserbase routing, normalized V8 coverage, unique CTRF artifacts, and more stable tests (timeouts, Browserbase skips, understudy click/API fixes).

Speed up CI by running Vitest and Playwright tests in parallel matrices, add V8/CTRF coverage, and stabilize Browserbase/LLM flows with hard timeouts and better cleanup. Server integration tests now use the runner’s Chromium for reliability and are folded into the main CI.

- **Refactors**
  - Parallelized core unit tests via discover-core-tests/core-unit-tests (Vitest matrix, fail-fast).
  - Parallelized E2E via discover-e2e-tests (Playwright local + Browserbase) gated on core changes; weighted Browserbase region selection; verify Chromium/CDP before tests; local uses runner Chromium via CHROME_PATH and --no-sandbox; Playwright can emit JUnit when CTRF_JUNIT_PATH is set.
  - Unified Lint & Build with Turbo caching and shared artifacts; tests/evals download and reuse; prepare script skips build in CI; SEA builds support sourcemaps.
  - Enforced configurable timeouts and parallelism via env (LLM_MAX_MS, BROWSERBASE_CDP_CONNECT_MAX_MS, BROWSERBASE_SESSION_CREATE_MAX_MS, LOCAL_SESSION_LIMIT_PER_E2E_TEST, BROWSERBASE_SESSION_LIMIT_PER_E2E_TEST, EVAL_*); added LLM and Browserbase hard timeouts; improved click event dispatch to reduce remote flakiness.
  - Explicitly close Browserbase sessions in tests and evals to free capacity.
  - Added composite actions: setup-node-pnpm-turbo, verify-chromium-launch, select-browserbase-region, publish-ctrf-report (JUnit→CTRF and evals), upload-v8-coverage.
  - Renamed TEST_ENV to STAGEHAND_ENV across E2E; added an env reporter for CI logs; removed the standalone stagehand-server tests workflow and ran server tests in the unified CI with Node’s test runner and GitHub Actions Chromium.

- **New Features**
  - Build/Exports: ESM in dist/esm and CJS in dist/cjs via export map; types from ESM; SEA binaries built from dist with sourcemaps; server SEA job targets CJS; new build/test/coverage scripts; Node loader maps package imports to dist/esm; ESM‑safe pathing (fileURLToPath); ESLint/Prettier tweaks; dotenv removed.
  - CI: Composite actions (setup-node-pnpm-turbo, verify-chromium-launch, select-browserbase-region, upload-ctrf-report, upload-v8-coverage); uses runner Chromium via CHROME_PATH with --no-sandbox and launch checks; weighted Browserbase region selection; unique CTRF artifacts; V8 coverage normalized/merged with debug logs and graceful cancellation; Playwright/Vitest discover tests in dist/esm; push/PR triggers enabled.
  - Tests & API: Vitest/Playwright run on dist/esm; new env reporter and closeV3 helper; tighter multi‑click CDP dispatch; higher timeouts and targeted skips (e.g., console on Browserbase) reduce flakiness; local launches honor CHROME_PATH; replay metrics now include parameters/result/timestamps/cost with parser fixes; AISdk/CustomOpenAI clients moved to lib/v3/external_clients and exported; env‑driven hard timeouts for LLM ops and Browserbase session create/CDP connect; understudy API usage fixes.

- **Migration**
  - Set STAGEHAND_BROWSER_TARGET=local|browserbase and CHROME_PATH; tune LLM_MAX_MS, BROWSERBASE_SESSION_CREATE_MAX_MS, and BROWSERBASE_CDP_CONNECT_MAX_MS; optionally set BROWSERBASE_REGION_DISTRIBUTION.
  - dotenv support is removed; set env vars explicitly in dev and CI.
  - Import @browserbasehq/stagehand via the export map; do not reference dist paths or prior tsup outputs.

<sup>Written for commit 8119145426a4e6e5cb614dee5b3f2d8fe41a65e9. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1632">Review in cubic</a></sup>

### DIFF
--- a/.changeset/fair-cups-protect.md
+++ b/.changeset/fair-cups-protect.md
@@ -1,0 +1,12 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+Remove automatic `.env` loading via `dotenv`.
+
+If your app relies on `.env` files, install `dotenv` and load it explicitly in your code:
+
+```ts
+import dotenv from "dotenv";
+dotenv.config({ path: ".env" });
+```


.github/workflows/ci.yml
```diff
@@ -55,848 +103,753 @@ jobs:
...
+              capture { print }
+              /^Evaluation summary written to / { capture=0 }
+            ' "$log_file"
+          )"
+
+          if [ -n "$summary_block" ]; then
+            {
+              echo "summary_text<<EOF"
+              echo "$summary_block"
+              echo "EOF"
```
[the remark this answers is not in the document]
pirate: nah its fine

packages/core/lib/v3/tests/page-addInitScript.spec.ts
```diff
@@ -39,20 +39,22 @@ test.describe("page.addInitScript", () => {
...
-      }
-      if (document.readyState === "loading") {
-        document.addEventListener("DOMContentLoaded", markScope, {
-          once: true,
-        });
-      } else {
-        markScope();
-      }
-    });
+    await first.addInitScript(`
```
pirate, line 42: tsx produces different output from func.toString() vs normal esm, to make sure it's stable in all builds (esm/cjs/tsx) I switched this to a string.

packages/core/lib/v3/types/public/api.ts
```diff
@@ -805,31 +805,38 @@ export const TokenUsageSchema = z
...
     timeMs: z.number().optional(),
+    cost: z.number().optional(),
   })
   .meta({ id: "TokenUsage" });
 
 /** Action entry in replay metrics */
 export const ReplayActionSchema = z
   .object({
-    method: z.string().optional(),
+    method: z.string(),
```
pirate, line 816: these were missing from the original types we had, the tests caught it so I added the missing params

packages/core/lib/v3/types/public/index.ts
```diff
@@ -9,7 +9,5 @@ export * from "./model";
 export * from "./options";
 export * from "./page";
 export * from "./sdkErrors";
-// Exporting the example AISdkClient for backwards compatibility
-// Note added for revisiting this scaffold for an improved version based on llm/aisdk.ts
-export { AISdkClient } from "../../../../examples/external_clients/aisdk";
-export { CustomOpenAIClient } from "../../../../examples/external_clients/customOpenAI";
+export { AISdkClient } from "../../external_clients/aisdk";
+export { CustomOpenAIClient } from "../../external_clients/customOpenAI";
```
pirate, line 13: these files need to be built as part of the esm/cjs build so they should not live under `examples/` (which is not part of our published dist)

packages/core/lib/v3/understudy/context.ts
```diff
@@ -95,16 +96,44 @@ export class V3Context {
...
-    const ctx = new V3Context(
-      conn,
-      opts?.env ?? "LOCAL",
-      opts?.apiClient ?? null,
-      opts?.localBrowserLaunchOptions ?? null,
-    );
-    await ctx.bootstrap();
-    await ctx.waitForFirstTopLevelPage(5000);
-    return ctx;
+    const connectTask = async () => {
```
pirate, line 99: this was previously unbounded and could hang forever, I applied a max limit that we can tune for CI

packages/core/lib/v3/understudy/locator.ts
```diff
@@ -401,32 +401,41 @@ export class Locator {
...
-        button: "none",
-      } as Protocol.Input.DispatchMouseEventRequest);
-
-      // Dispatch mouse pressed and released events for the given click count
-      for (let i = 1; i <= clickCount; i++) {
-        await session.send<never>("Input.dispatchMouseEvent", {
-          type: "mousePressed",
+      // Dispatch input (from the same session) without waiting between events.
+      // This keeps multi-clicks within tight thresholds on remote browsers.
+      const dispatches: Array<Promise<unknown>> = [];
```
pirate, line 406: this was a real bug that was causing flaky tests, previously network/CI CPU jitter could lead to unbounded delay between the mousedown and mouseup, causing double clicks to become two single clicks.

I fixed it by generating the click events and promise.race()'ing so they execute nearly back-to-back.

packages/core/lib/v3/understudy/page.ts
```diff
@@ -1389,34 +1389,43 @@ export class Page {
...
-      x,
-      y,
-      button: "none",
-    } as Protocol.Input.DispatchMouseEventRequest);
-
-    // Dispatch mouse pressed and released events for the given click count
-    for (let i = 1; i <= clickCount; i++) {
-      await this.mainSession.send<never>("Input.dispatchMouseEvent", {
-        type: "mousePressed",
+    const dispatches: Array<Promise<unknown>> = [];
```
pirate, line 1395: same bugfix here, mousedown and up need to be sent closer together to guarantee it registers as a doubleclick

packages/core/lib/v3/api.ts
```diff
@@ -480,8 +480,20 @@ export class StagehandAPIClient {
         if (tokenUsage) {
           const inputTokens = tokenUsage.inputTokens || 0;
           const outputTokens = tokenUsage.outputTokens || 0;
-          const reasoningTokens = tokenUsage.reasoningTokens || 0;
-          const cachedInputTokens = tokenUsage.cachedInputTokens || 0;
+          const reasoningTokens =
```
pirate, line 483: just a typefix for the optional values that are not always present

packages/core/lib/v3/v3.ts
```diff
@@ -122,7 +121,6 @@ function resolveModelConfiguration(
 
   return { modelName: DEFAULT_MODEL_NAME };
 }
-dotenv.config({ path: ".env" });
```
pirate, line 125: libraries should never auto-load .env files!
we should only read from our own env that we are passed.

packages/core/lib/inference.ts
```diff
@@ -74,8 +81,8 @@ export async function extract<T extends StagehandZodObject>({
   }
 
   const extractStartTime = Date.now();
-  const extractionResponse =
-    await llmClient.createChatCompletion<ExtractionResponse>({
+  const extractionResponse = await withLlmTimeout(
```
pirate, line 84: previously LLM operations were potentially unbounded, leading to super long test runtimes when there is upstream lag, we should always set an upper bound on LLM inference time in CI so we can differentiate lag in LLM providers vs our own code.

packages/docs/package.json
```diff
@@ -13,8 +14,7 @@
   "license": "ISC",
   "dependencies": {
     "mintlify": "^4.2.47",
-    "tsx": "^4.19.2",
-    "zod": "^3.25.0"
+    "zod": "^4.2.1"
```
pirate, line 17: (I think bumping zod here is ok, this is just the docs package, not our main core package, correct me if I'm wrong)

packages/server/src/lib/InMemorySessionStore.ts
```diff
@@ -181,12 +203,35 @@ export class InMemorySessionStore implements SessionStore {
 
     // Create V3 instance (lazy initialization)
     const options = this.buildV3Options(node.params, ctx, node.loggerRef);
+    const shouldRetryInit = node.params.browserType === "local";
```
pirate: The browser can take a few seconds to start, this is expected in CI and slower local machines. Instead of hard erroring and forcing the user to retry in a loop, we can simply try to connect a few times within the server package before giving up.

packages/core/scripts/resolve-stagehand-dist.js
seanmcguire12, line 1: should be able to do without the custom resolver. Node supports [this](https://nodejs.org/api/packages.html#self-referencing-a-package-using-its-name) which already resolves import "@browserbasehq/stagehand" through the exports field in package.json

packages/core/scripts/register-stagehand-dist.js
seanmcguire12, line 1: same as comment on `packages/core/scripts/resolve-stagehand-dist.js`

packages/core/vitest.config.ts
```diff
@@ -1,6 +1,86 @@
...
+  const rel = path.relative(coreRoot, resolvedPath);
+  const candidate = path.join(distEsmRoot, rel);
+  return resolveDistFile(candidate);
+};
+
+const resolveStagehandSpecifier = (source: string): string | null => {
+  if (source === "@browserbasehq/stagehand") {
+    return resolveDistFile(path.join(distEsmRoot, "index.js"));
+  }
+  if (!source.startsWith("@browserbasehq/stagehand/")) return null;
```
seanmcguire12: duplicates logic from `resolve-stagehand-dist.js`.

packages/core/scripts/test-core.ts
```diff
@@ -0,0 +1,156 @@
...
+
+const registerPath = path.join(
+  repoRoot,
+  "packages",
+  "core",
+  "scripts",
+  "register-stagehand-dist.js",
+);
+const baseNodeOptions = `--enable-source-maps --experimental-specifier-resolution=node --import ${registerPath}`;
+const nodeOptions = [process.env.NODE_OPTIONS, baseNodeOptions]
```
seanmcguire12, line 72: should be removable if we rely on Node's self-referencing through `package.json` exports. This would also let you drop `--experimental-specifier-resolution=node`

packages/core/scripts/test-e2e.ts
```diff
@@ -0,0 +1,213 @@
...
+);
+const baseNodeOptions = [
+  "--enable-source-maps",
+  "--experimental-specifier-resolution=node",
+  `--import ${registerPath}`,
+].join(" ");
+const nodeOptions = [process.env.NODE_OPTIONS, baseNodeOptions]
+  .filter(Boolean)
+  .join(" ");
+
```
seanmcguire12: same as comment on `test-core.ts`: injecting the custom loader via NODE_OPTIONS here should be unnecessary

packages/core/lib/v3/tests/v3.playwright.config.ts
```diff
@@ -1,14 +1,111 @@
-import { defineConfig } from "@playwright/test";
+import { defineConfig, type ReporterDescription } from "@playwright/test";
+import fs from "node:fs";
+import path from "node:path";
+
+const resolveRepoRoot = (startDir: string): string => {
```
seanmcguire12, line 5: `resolveRepoRoot` is the third duplicate of the same `findRepoRoot` logic (also in `resolve-stagehand-dist.js` and `test-utils.ts`). if we still need this after removing the custom loader, can we have it exist in a single module that we can import?

packages/evals/index.eval.ts
```diff
@@ -41,14 +43,14 @@ import {
 import { AISdkClientWrapped } from "./lib/AISdkClientWrapped";
 import { getAISDKLanguageModel } from "@browserbasehq/stagehand/lib/v3/llm/LLMProvider";
```
monadoid: Evals still deep‑import `@browserbasehq/stagehand/lib/v3/*` (LLMProvider/LLMClient/zodCompat/public types). With the new exports map, those subpaths are blocked, so running evals now throws ERR_PACKAGE_PATH_NOT_EXPORTED.

The cleanest solution is to export these from stagehand if we want to use them in other packages.

packages/core/scripts/build-cjs.ts
```diff
@@ -0,0 +1,75 @@
...
+  const result = spawnSync("pnpm", args, { stdio: "inherit", cwd: repoRoot });
+  if (result.status !== 0) {
+    process.exit(result.status ?? 1);
+  }
+};
+
+fs.rmSync(cjsDist, { recursive: true, force: true });
+fs.mkdirSync(cjsDist, { recursive: true });
+
+run([
```
monadoid, line 29: We bundle only index.ts for CJS but I think we might need to copy supervisor.js as well for CJS to work?

packages/evals/scripts/build-esm.ts
```diff
@@ -0,0 +1,142 @@
...
+]);
+
+const run = (args: string[]) => {
+  const result = spawnSync("pnpm", args, { stdio: "inherit", cwd: repoRoot });
+  if (result.status !== 0) {
+    process.exit(result.status ?? 1);
+  }
+};
+
+const resolveRuntimeSpecifier = (
```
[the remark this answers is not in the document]
pirate: no, we're getting rid of it all in a fast-follow pr that manually adds .js to all our imports

packages/core/scripts/build-cjs.ts
pirate: good catch, fixed

packages/core/package.json
```diff
@@ -40,7 +52,6 @@
   "license": "MIT",
   "peerDependencies": {
     "deepmerge": "^4.3.1",
-    "dotenv": "^16.4.5",
```
monadoid, line 43: We probably either want to keep this, or we need to update the docs to tell users that they need to install dotenv themselves - keeping it seems like the best UX to me?

pirate: I will update the docs + release notes, libraries should never be pulling in dotenv automatically, thats for users to do before running our code, otherwise `.env` files can force-fill values that we want to leave empty for certain CI runs. I removed all the dotenv usage in stagehand so it doesn't make sense to keep it as a peer dep imo.

pirate: added changeset and docs note ✅

PR #1642, opened 2026-01-30
[fix]: ctx `addInitScript` on popup pages

### LINKED ISSUE #1628: context.addInitScript is not applied on new pages
## Environment Information

Please provide the following information to help us reproduce and resolve your issue:

**Stagehand:**

- Language/SDK: TypeScript
- Stagehand version: 3.0.8

## Issue Description

Using `context.addInitScript` I expected that the script would be injected on every new page that is opened.
The docs say:
```
Automatically applies the same script to any pages created after calling context.addInitScript()
```

It appears this is not the case. Does not matter if the page was opened via clicking within an existing tab or via `newPage()`

## Minimal repro
```
import { setTimeout as delay } from 'node:timers/promises';
import { Stagehand } from '@browserbasehq/stagehand';

const stagehand = new Stagehand({ env: 'LOCAL' });
await stagehand.init();

let page = stagehand.context.pages()[0];
if (!page) throw new Error('Expected browser page');

// Define the HTML we'll serve
const html = `
<!DOCTYPE html>
<html><head><title>Init script test</title>

</head><body id="body">
  <h2>Testing contxt.addInitScript</h2>

  <p><b>Page background should be red!</b></p>
  <p>
  <a href="/" target="_blank">Open another tab</a>
  </p>
</body></html>
`;

const port = 6789;
const localhostUrl = `http://127.0.0.1:${port}/`;

const server = Bun.serve({
  port,
  fetch(req) {
    return new Response(html, { headers: { "Content-Type": "text/html" }});
  }
});

console.log(`Test server running at ${localhostUrl}`);

// use context.addInitScript - it should run on _every_ page
await stagehand.context.addInitScript(`
(() => {
  document.addEventListener('DOMContentLoaded', () => {
    document.getElementById('body').style.backgroundColor = 'red';
   });
})();  
`);

// navigate in the initial page
await page.goto(localhostUrl);
await delay(1000);
// open a new page via clicking
await page.locator('a').click();
await delay(1000);
// open another via newPage()
await stagehand.context.newPage(localhostUrl);
await delay(1000);

await stagehand.close();
await server.stop();
```

seanmcguire12:
# why
- fixes #1628 
- scripts injected via`context.addInitScript()` were not reliably applied to pages that were created via a popup (eg, clicking a link that opens a new page), or, when a new page was created via `context.newPage(url)`. 
- when a target was created directly at a URL, navigation often started before attaching to the target & installing the init script. this means that the script would either run too late or not run at all. 
# what changed
- enabled `waitForDebuggerOnStart` in `Target.setAutoAttach` in `cdp.ts`, so that new targets pause until we explicitly resume them
- in `understudy/context.ts`:
  - `onAttachedToTarget()` now installs context init scripts on the **session** before resuming, then resumes with `Runtime.runIfWaitingForDebugger`. this guarantees scripts are registered before the first real document executes.
  - worker targets are resumed immediately. we still are not attaching to them, but they need to be resumed since we are now telling everything to wait for the debugger.
  - added `seedOnly` to `applyInitScriptsToPage()` so the `page` object records scripts without re‑installing them
  - `newPage()` first creates targets at `about:blank` and then navigates after attach (fire‑and‑forget). this avoids the race where the first document loads before init scripts are installed
-  in`understudy/page.ts`, i added `seedInitScript()` so the `page` can track scripts that were already installed on the session
# test plan
- added additional tests to `tests/context-addInitScript.spec.ts` which explicitly covers `context.newPage(url)` and popup pages to ensure init scripts run on the initial document

---
## Summary by cubic
Fixes a race where context.addInitScript() didn’t run on the first document for new pages opened via popups or context.newPage(url). New targets now pause until init scripts are registered, ensuring scripts run reliably.

- **Bug Fixes**
  - Paused new targets with waitForDebuggerOnStart; resume after script install.
  - Installed init scripts on the CDP session before first document executes.
  - Resumed worker targets immediately so they don’t stay paused.
  - Created newPage() at about:blank, attach, then navigate (fire-and-forget).
  - Seeded scripts/URL on Page to avoid double installs and timing issues.
  - Added tests for newPage(url) and popup pages to verify scripts run on initial load.

<sup>Written for commit 21193c2084408619a2a2e5f0d86c7126310e5e2a. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1642">Review in cubic</a></sup>

### DIFF
--- a/packages/core/lib/v3/understudy/context.ts
+++ b/packages/core/lib/v3/understudy/context.ts
@@ -472,7 +472,10 @@ export class V3Context {
     // for the init-script sends to settle in the background.
     await resume();
     if (installPromises.length) {
-      await Promise.allSettled(installPromises);
+      const results = await Promise.allSettled(installPromises);
+      if (results.some((r) => r.status === "rejected")) {
+        scriptsInstalled = false;
+      }
     }
 
     try {


seanmcguire12:
@cubic-dev-ai

seanmcguire12:
@cubic-dev-ai

seanmcguire12:
@greptileai

packages/core/lib/v3/understudy/cdp.ts
```diff
@@ -109,12 +109,7 @@ export class CdpConnection implements CDPSessionLike {
     await this.send("Target.setAutoAttach", {
       autoAttach: true,
       flatten: true,
-      waitForDebuggerOnStart: false,
-      filter: [
-        { type: "worker", exclude: true },
-        { type: "shared_worker", exclude: true },
-        { type: "service_worker", exclude: true },
-      ],
+      waitForDebuggerOnStart: true,
```
pirate, line 112: I think this might have a stealth/reliability impact, gaps in runtime performance metrics as a result of debuggerOnStart are one of the signals bot-blockers use iirc. Also there are weird situations with using `waitForDebuggerOnStart` on service workers.

might be worth only setting this if there are initScripts needed

Further reading on issues with `waitForDebuggerOnStart`:

- https://github.com/puppeteer/puppeteer/issues/3667#issuecomment-1588632073
- https://issues.chromium.org/issues/40280584
- https://issues.chromium.org/issues/428100629
- https://issues.chromium.org/issues/40269916
- https://issues.chromium.org/issues/376417348 reports one such weirdness:
	> This works fine (as expected) while a new target is opened with a normal click (left click). The new tab is stuck in hanging/loading state until runIfWaitingForDebugger command is sent. We receive waitingForDebugger: true for attachedTotTarget event.
	
	> However, when a tab is opened using ctrl + click (command + click on MAC), browser behaves differently. The tab starts loading at once, without waiting for runIfWaitingForDebugger command. In attachedTotTarget waitingForDebugger parameter is still true.

PR #1672, opened 2026-02-09
[feat]: add `keepAlive` param

seanmcguire12:
# why
- to give users the freedom to decide whether the browser should be kept open or "alive" when their stagehand instance is killed
- to follow best practices by avoiding usage of global signal handlers inside library code
# what changed
new `keepAlive` option:
- added an optional, boolean `keepAlive` parameter in the stagehand constructor params. `keepAlive` defaults to `false`
- `keepAlive` on the constructor overrides `browserbaseSessionCreateParams.keepAlive` when both are provided out-of-process shutdown supervisor (`packages/core/lib/v3/shutdown/`):
- replaced the old in-process global signal handlers (process.once("SIGINT"), etc.) with a detached supervisor process that watches a lifeline (stdin pipe + IPC channel) to the parent
- when `keepAlive` is set to `false`, the supervisor is spawned during `init()` and performs best-effort cleanup if the parent dies unexpectedly:
  - if `env` is `"LOCAL"` the supervisor kills the Chrome PID (with SIGTERM → SIGKILL escalation) and removes any temp user-data directory
  - if `env` is `"BROWSERBASE"` & using the stagehand API, the supervisor sends `REQUEST_RELEASE` to the browserbase API
- the supervisor also includes PID polling to guard against process reuse (killing a recycled PID
- when `keepAlive` is set to `true`, no supervisor is spawned, & the browser is left running
- when `disableAPI` is set to `true` (direct WS to Browserbase), no supervisor is spawned since the browser will auto release when the WS is dropped

refactored `close()` into helpers:
- pulled browser teardown logic into `shutdown/shutdownBrowser.ts` (`shutdownBrowserSession`, `shutdownLocalBrowser`)
- pulled stagehand internal state cleanup into `shutdown/shutdownStagehand.ts` (`shutdownStagehandResources`, `finalizeStagehandShutdown`)
-`.close()` now skips `apiClient.end()` when `keepAlive: true`

local browser launch changes:
- updated local launch to pass chrome launcher’s `handleSIGINT` based on `keepAlive` in `packages/core/lib/v3/v3.ts` and `packages/core/lib/v3/launch/local.ts`. this prevents chrome from getting killed on Ctrl+C when `keepAlive: true`
- updated local launch so that we `unref()` the chrome child process after launch so Node can exit while chrome stays open (`packages/core/lib/v3/v3.ts`)
### behaviour notes:
- setting `keepAlive: true` will keep the browser alive in the following scenarios:
  - stagehand instance gets killed because of an uncaught error (eg, `unhandledRejection`, `unhandledException`)
  - stagehand instance gets killed because of a `SIGTERM` or `SIGINT`
  - stagehand instance gets killed with an explicit call to `stagehand.close()`
# test plan
- added tests in `keep-alive.spec.ts` that validate expected behaviour across the following scenarios: 
  - `unhandledRejection`/`unhandledException`, 
  - `SIGTERM`/`SIGINT`, 
  - `stagehand.close()`
- each scenario is validated in each of the following browser configurations:
  - using a local browser launched by stagehand
  - using a browserbase browser, connected directly over the WS
  - using a browserbase browser + the stagehand API

---
## Summary by cubic
Adds a keepAlive option to Stagehand V3 so you control whether the browser/session stays up after stagehand.close(), signals, or uncaught errors. Replaces global process handlers with a per‑instance crash‑only supervisor with a ready handshake and PID polling to avoid race conditions (STG‑1338).

- **New Features**
  - keepAlive?: boolean (default false) on the V3 constructor; overrides browserbaseSessionCreateParams.keepAlive when set.
  - Local: set handleSIGINT based on keepAlive and unref Chrome when true; when false, a detached supervisor watches a stdin/IPC lifeline and on parent crash/error SIGTERM→SIGKILLs Chrome and removes the temp profile (with PID polling). A shared cleanup helper is used by both close() and the supervisor for consistent local teardown.
  - Browserbase: propagate keepAlive to session creation; when keepAlive is true, close() skips apiClient.end() so the session stays RUNNING; when false (and API enabled), the supervisor requests REQUEST_RELEASE on crash/error; when disableAPI is true (direct WS), no supervisor is spawned and the session ends when the WS drops.

- **Migration**
  - Set keepAlive: true in the V3 constructor to keep the browser/session open on errors, signals, or stagehand.close(); this overrides any value in browserbaseSessionCreateParams.
  - When keepAlive is true, close resources yourself when done (Local: send Browser.close; Browserbase: request release/end the session).

<sup>Written for commit 3887d8f373d27afca6560c9436132fbfbcbb9845. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1672">Review in cubic</a></sup>

### DIFF
--- a/packages/core/lib/v3/shutdown/supervisor.ts
+++ b/packages/core/lib/v3/shutdown/supervisor.ts
@@ -7,12 +7,12 @@
  * - STAGEHAND_API: request session release (when keepAlive is false)
  */
 
-import fs from "node:fs";
 import Browserbase from "@browserbasehq/sdk";
 import type {
   ShutdownSupervisorConfig,
   ShutdownSupervisorMessage,
 } from "../types/private/shutdown";
+import { cleanupLocalBrowser } from "./cleanupLocal";
 
 const SIGKILL_POLL_MS = 500;
 const SIGKILL_TIMEOUT_MS = 10_000;
@@ -81,18 +81,12 @@ const cleanupLocal = async (
   cfg: Extract<ShutdownSupervisorConfig, { kind: "LOCAL" }>,
 ) => {
   if (cfg.keepAlive) return;
-  if (cfg.pid) {
-    if (!pidGone) {
-      await safeKill(cfg.pid);
-    }
-  }
-  if (cfg.createdTempProfile && !cfg.preserveUserDataDir && cfg.userDataDir) {
-    try {
-      fs.rmSync(cfg.userDataDir, { recursive: true, force: true });
-    } catch {
-      // ignore cleanup errors
-    }
-  }
+  await cleanupLocalBrowser({
+    killChrome: cfg.pid && !pidGone ? () => safeKill(cfg.pid) : undefined,
+    userDataDir: cfg.userDataDir,
+    createdTempProfile: cfg.createdTempProfile,
+    preserveUserDataDir: cfg.preserveUserDataDir,
+  });
 };
 
 const cleanupBrowserbase = async (
--- a/packages/core/lib/v3/v3.ts
+++ b/packages/core/lib/v3/v3.ts
@@ -30,6 +30,7 @@ import {
   unbindInstanceLogger,
   withInstanceLogContext,
 } from "./logger";
+import { cleanupLocalBrowser } from "./shutdown/cleanupLocal";
 import { startShutdownSupervisor } from "./shutdown/supervisorClient";
 import { resolveTools } from "./mcp/utils";
 import {
@@ -1430,25 +1431,13 @@ export class V3 {
 
       // Kill local Chrome and clean up temp profile when keepAlive is not enabled
       if (!keepAlive && this.state.kind === "LOCAL") {
-        try {
-          await this.state.chrome.kill();
-        } catch {
-          // best-effort cleanup
-        }
-        if (
-          this.state.createdTempProfile &&
-          !this.state.preserveUserDataDir &&
-          this.state.userDataDir
-        ) {
-          try {
-            fs.rmSync(this.state.userDataDir, {
-              recursive: true,
-              force: true,
-            });
-          } catch {
-            // ignore cleanup errors
-          }
-        }
+        const localState = this.state;
+        await cleanupLocalBrowser({
+          killChrome: () => localState.chrome.kill(),
+          userDataDir: localState.userDataDir,
+          createdTempProfile: localState.createdTempProfile,
+          preserveUserDataDir: localState.preserveUserDataDir,
+        });
       }
     } finally {
       this.stopShutdownSupervisor();
--- a/packages/core/lib/v3/shutdown/cleanupLocal.ts
+++ b/packages/core/lib/v3/shutdown/cleanupLocal.ts
@@ -1,0 +1,35 @@
+import fs from "node:fs";
+
+/**
+ * Shared cleanup logic for locally launched Chrome.
+ *
+ * Used by both `V3.close()` (normal shutdown) and the supervisor process
+ * (crash cleanup). The caller provides a `killChrome` callback since the
+ * kill mechanism differs: chrome-launcher's `chrome.kill()` in-process
+ * vs raw `process.kill(pid)` from the supervisor.
+ */
+export async function cleanupLocalBrowser(opts: {
+  killChrome?: () => Promise<void> | void;
+  userDataDir?: string;
+  createdTempProfile?: boolean;
+  preserveUserDataDir?: boolean;
+}): Promise<void> {
+  if (opts.killChrome) {
+    try {
+      await opts.killChrome();
+    } catch {
+      // best-effort
+    }
+  }
+  if (
+    opts.createdTempProfile &&
+    !opts.preserveUserDataDir &&
+    opts.userDataDir
+  ) {
+    try {
+      fs.rmSync(opts.userDataDir, { recursive: true, force: true });
+    } catch {
+      // ignore cleanup errors
+    }
+  }
+}


seanmcguire12:
@cubic-dev-ai

seanmcguire12:
@greptileai

packages/core/lib/v3/tests/keep-alive.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+  buildKeepAliveCases,
+  getKeepAliveEnvConfig,
+  runKeepAliveCase,
+} from "./keep-alive.helpers";
+
+test.describe.parallel("keepAlive behavior", () => {
+  const { testEnv, apiKey, projectId, hasBrowserbaseCreds } =
+    getKeepAliveEnvConfig();
+  const cases = buildKeepAliveCases(testEnv);
+
```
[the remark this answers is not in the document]
seanmcguire12: each test case spawns its own child process

seanmcguire12:
@greptileai

seanmcguire12:
@cubic-dev-ai

seanmcguire12:
@cubic-dev-ai

seanmcguire12:
@greptileai

seanmcguire12:
@greptileai

seanmcguire12:
@cubic-dev-ai

seanmcguire12:
@cubic-dev-ai

seanmcguire12:
@cubic-dev-ai

seanmcguire12:
@cubic-dev-ai

seanmcguire12:
@cubic-dev-ai

seanmcguire12:
@cubic-dev-ai

seanmcguire12:
@greptileai

seanmcguire12:
@cubic-dev-ai

packages/core/lib/v3/shutdown/supervisor.ts
```diff
@@ -0,0 +1,165 @@
...
+    return;
+  }
+
+  await new Promise((resolve) => setTimeout(resolve, SIGNAL_POLL_MS));
+  if (!isAlive()) return;
+  try {
+    process.kill(pid, "SIGKILL");
+  } catch {
+    // best-effort
+  }
```
pirate: dont just wait one more poll interval, you should loop every 500ms up to 10s and then only send the SIGKILL after 10s.

on slow systems chrome can take 10-15s+ to exit cleanly and save changes to user_data_dir, force-killing during exit is dangerous because it can destroy the user_data_dir and leave the profile in a corrupted state.

the happy path wont be affected if you poll, if it exits cleanly in 500ms it should return, but it's best to keep polling up to 10s before you send the SIGKILL.

packages/core/lib/v3/types/private/shutdown.ts
```diff
@@ -0,0 +1,32 @@
...
+      createdTempProfile?: boolean;
+      preserveUserDataDir?: boolean;
+    }
+  | {
+      kind: "STAGEHAND_API";
+      keepAlive: boolean;
+      sessionId: string;
+      apiKey: string;
+      projectId: string;
+    };
```
pirate, line 20: you should pass the api base url too probably, otherwise it will always assume prod? or does it inherit the env var?

packages/core/lib/v3/v3.ts
```diff
@@ -856,6 +864,18 @@ export class V3 {
...
+          if (!keepAlive && chromePid) {
+            const supervisor = this.startShutdownSupervisor({
+              kind: "LOCAL",
+              keepAlive: false,
+              pid: chromePid,
+              userDataDir,
+              createdTempProfile: createdTemp,
+              preserveUserDataDir: !!lbo.preserveUserDataDir,
+            });
+            await supervisor?.ready;
```
pirate, line 870: might want to promise.race this with a setTimeout so it doesnt block forever if it fails, but not super important

packages/core/lib/v3/v3.ts
```diff
@@ -1361,81 +1397,87 @@ export class V3 {
...
+      if (
+        localState.createdTempProfile &&
+        !localState.preserveUserDataDir &&
+        localState.userDataDir
+      ) {
+        const dir = localState.userDataDir;
+        cleanupUserDataDir = () => {
+          fs.rmSync(dir, { recursive: true, force: true });
+        };
       }
```
pirate: is this not duplicating `shutdown/` `shutdownLocalBrowser()`'s cleanupUserDataDir() logic?

packages/core/lib/v3/v3.ts
```diff
@@ -1361,81 +1397,87 @@ export class V3 {
...
-      }
-      // Clear accumulated data to free memory
-      this._history = [];
-      this.actHandler = null;
-      this.extractHandler = null;
-      this.observeHandler = null;
-      // Remove from global registry
-      V3._instances.delete(this);
+      this.stopShutdownSupervisor();
+      finalizeStagehandShutdown({
```
pirate: same here, this looks very similar to the shutdown reset in `shutdown/`?

packages/core/lib/v3/types/private/shutdown.ts
seanmcguire12: its automatically loaded from env

packages/core/lib/v3/v3.ts
seanmcguire12: it is doing that internally [here](https://github.com/browserbase/stagehand/blob/2080b235d9d27371e41be23fbf892a7d615fa449/packages/core/lib/v3/shutdown/supervisorClient.ts#L95)

packages/core/lib/v3/v3.ts
```diff
@@ -1361,58 +1389,70 @@ export class V3 {
...
+          this.state.userDataDir
+        ) {
+          try {
+            fs.rmSync(this.state.userDataDir, {
+              recursive: true,
+              force: true,
+            });
+          } catch {
+            // ignore cleanup errors
           }
```
pirate: does this duplicate `cleanupLocal()` still?

pirate (approved):

PR #1678, opened 2026-02-11
fix: support custom LLM clients in agent mode

### LINKED ISSUE #1669: The agent cannot use a custom LLM client model.
```tsx
    env: "LOCAL",
    disableAPI: true,
    verbose: 0,
    llmClient: new CustomOpenAIClient({
      modelName: "glm-4-plus",
      client: new OpenAI({
        apiKey: process.env.ZHIPU_API_KEY,
        baseURL: process.env.ZHIPU_URL,
      }),
    }),
  });
  await stagehand.init();

...
     await stagehand.act(`do something`) // right

    const agent = stagehand.agent();
    const img = await agent.execute("find all img : //img[@alt='Stagehand']")
    console.log("🚀 ~ test ~ img:", img) // error


```

Image

   "@browserbasehq/stagehand": "3.1.0-alpha-0e4a144832878aa334b7b18840a22897d86cd509",

shrey150:
## Summary
Adds `getLanguageModel()` to `CustomOpenAIClient` using `@ai-sdk/openai`'s `createOpenAI`, so custom OpenAI-compatible clients (ZhipuAI, Ollama, etc.) work with agent mode.

Closes #1669

## Test plan
- [x] `getLanguageModel()` returns a valid `LanguageModelV2` (correct `modelId`, `doGenerate`, `doStream`, `specificationVersion: "v2"`)
- [x] `getLanguageModel()` preserves custom `baseURL` for OpenAI-compatible providers (ZhipuAI, Ollama)
- [x] Live API call through `generateText()` (the AI SDK path agent mode uses) succeeds with the model from `getLanguageModel()`
- [x] `createChatCompletion()` (act/extract/observe path) still works — no regression
- [x] `MissingLLMConfigurationError` no-arg constructor still works
- [x] All 341 existing vitest tests pass with zero regressions

### DIFF
--- a/packages/core/lib/v3/external_clients/customOpenAI.ts
+++ b/packages/core/lib/v3/external_clients/customOpenAI.ts
@@ -8,6 +8,8 @@
 import type { AvailableModel } from "../types/public/model";
 import { CreateChatCompletionOptions, LLMClient } from "../llm/LLMClient";
 import OpenAI from "openai";
+import { createOpenAI } from "@ai-sdk/openai";
+import type { LanguageModelV2 } from "@ai-sdk/provider";
 import type {
   ChatCompletion,
   ChatCompletionAssistantMessageParam,
@@ -33,6 +35,14 @@ export class CustomOpenAIClient extends LLMClient {
     super(modelName as AvailableModel);
     this.client = client;
     this.modelName = modelName as AvailableModel;
+  }
+
+  public getLanguageModel(): LanguageModelV2 {
+    const provider = createOpenAI({
+      apiKey: this.client.apiKey,
+      baseURL: this.client.baseURL,
+    });
+    return provider(this.modelName);
   }
 
   async createChatCompletion<T = ChatCompletion>({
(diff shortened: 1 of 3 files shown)


packages/core/examples/external_clients/customOpenAI.ts
```diff
@@ -38,6 +40,14 @@ export class CustomOpenAIClient extends LLMClient {
...
   }
 
+  public getLanguageModel(): LanguageModelV2 {
+    const provider = createOpenAI({
+      apiKey: this.client.apiKey,
+      baseURL: this.client.baseURL,
+    });
+    return provider(this.modelName);
+  }
+
```
tkattkat: it seems this just uses openai client meaning the custom one is never actually used?

shrey150: from my understanding, this just creates an OpenAI-compatible client (to populate the base URL), which is what's needed to solve the "failed to execute task: no llm api key or llm client configured" issue

PR #1681, opened 2026-02-11
[feat]: add cookie related getters & setters

tkattkat:
# Why

Browser automation workflows frequently need to manage authentication state, persist sessions across runs, and handle cookie-based features. Currently, users have no direct API to read, write, or clear cookies in Stagehand, forcing workarounds through raw CDP commands or page-level JavaScript injection.

# What Changed

## Cookie Management APIs

Added cookie management methods that mirror the familiar Playwright BrowserContext API:

**On `context` instance:**
- `context.cookies(urls?)` — Get all browser cookies, optionally filtered by URL(s)
- `context.addCookies(cookies)` — Set one or more cookies in the browser context
- `context.clearCookies(options?)` — Clear all cookies or selectively by name/domain/path (supports RegExp)

### Types Added
- `Cookie` — Cookie object returned by the browser
- `CookieParam` — Parameters for setting a cookie (supports `url` OR `domain`+`path`)
- `ClearCookieOptions` — Filter options for selective cookie clearing

# Test Plan

## Cookie Management
- [ ] `context.cookies()` returns all cookies when no URLs provided
- [ ] `context.cookies(url)` filters cookies by single URL
- [ ] `context.cookies([url1, url2])` filters cookies by multiple URLs
- [ ] `context.addCookies([...])` successfully sets cookies
- [ ] `context.addCookies()` validates cookie params (requires url OR domain+path)
- [ ] `context.addCookies()` rejects `sameSite: "None"` without `secure: true`
- [ ] `context.addCookies()` throws on browser rejection (checks CDP success flag)
- [ ] `context.clearCookies()` clears all cookies atomically via single CDP call
- [ ] `context.clearCookies({ name: "session" })` clears by exact name
- [ ] `context.clearCookies({ domain: /\.example\.com/ })` clears by regex
- [ ] Selective clear only deletes matching cookies, leaves others untouched

### DIFF
--- a/.changeset/fast-buses-kneel.md
+++ b/.changeset/fast-buses-kneel.md
@@ -1,5 +1,5 @@
 ---
-"@browserbasehq/stagehand": patch
+"@browserbasehq/stagehand": minor
 ---
 
-Add native support for setting and getting cookies
+Add cookie management APIs: `context.addCookies()`, `context.clearCookies()`, & `context.cookies()`


packages/core/lib/v3/v3.ts
```diff
@@ -1359,6 +1364,45 @@ export class V3 {
...
+   *
+   * - Called with no arguments: clears **all** cookies.
+   * - Called with filter options: only cookies matching every supplied criterion
+   *   are removed.
+   */
+  async clearCookies(options?: ClearCookieOptions): Promise<void> {
+    if (!this.ctx) throw new StagehandNotInitializedError("clearCookies()");
+    return this.ctx.clearCookies(options);
+  }
+
```
seanmcguire12: what is the reason for exposing on the stagehand object? i would think it makes more sense to keep this on the `context` object but open to opinions

tkattkat: word - moved to context

seanmcguire12:
@cubic-dev-ai

packages/core/lib/v3/understudy/context.ts
```diff
@@ -823,4 +838,106 @@ export class V3Context {
...
+      options?.path !== undefined;
+
+    if (!hasFilter) {
+      // Atomic single-call wipe — no race condition, no O(N) roundtrips.
+      await this.conn.send("Storage.clearCookies");
+      return;
+    }
+
+    const current = await this.cookies();
+    const toKeep = current.filter((c) => !cookieMatchesFilter(c, options!));
```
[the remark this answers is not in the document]
seanmcguire12: this should be fine for now since none of these fields are settable by users. this can be fixed when we add support for `partitionKey`

seanmcguire12:
@cubic-dev-ai

pirate (approved):
seems ok but we should really provide a unified API that sets all of these together, cookies alone are not enough for modern sites:

- cookies
- session storage
- local storage
- indexed db

This is why plyawright apis are all stuff like `storageState()` not `cookieState()`:

- https://playwright.dev/docs/auth
- https://playwright.dev/docs/auth#session-storage

image

PR #1692, opened 2026-02-17
[fix]: skip piercer injection for chrome-extension:// and non-HTML targets

shrey150:
## Summary

Navigating to a PDF URL crashes Chrome's built-in PDF viewer because the piercer script gets injected into the PDF viewer's `chrome-extension://` target.

The fix adds an `isNonWebTarget()` predicate (alongside existing `isTopLevelPage()`) that uses an **allowlist** — only `"page"` and `"iframe"` targets receive piercer injection. All other target types (workers, background pages, browser, webviews, worklets, etc.) are resumed but skipped. Additionally, `chrome-extension://` and `chrome://` URLs are always skipped even if the type is `"page"`, since Chrome's PDF viewer reports as `type: "page"` with a `chrome-extension://` URL.

This is future-proof: Chrome currently has [14 target types](https://chromium.googlesource.com/chromium/src/+/main/content/browser/devtools/devtools_agent_host_impl.cc) and adding new ones won't require updating a blocklist.

## Root Cause

`onAttachedToTarget()` only filtered out worker targets. Chrome's PDF viewer (`chrome-extension://mhjfbmdgcfjbbpaeojofohoefgiehjai`) gets auto-attached as a `type: "page"` target — the piercer script patches `Element.prototype.attachShadow` and walks the DOM inside this extension, corrupting the viewer.

## Test plan

- [x] Open a PDF URL (e.g. `https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf`) with Stagehand — verify it loads without error
- [x] Verify normal page navigation still works after viewing a PDF
- [x] Verify OOPIF (cross-origin iframe) targets still get piercer injection
- [x] Verify shadow DOM piercing works (open + closed shadow roots tracked and accessible)

### DIFF
--- a/packages/core/lib/v3/understudy/context.ts
+++ b/packages/core/lib/v3/understudy/context.ts
@@ -69,7 +69,6 @@ function getFirstTopLevelPageTimeoutMs(): number {
   );
 }
 
-
 /**
  * V3Context
  *


packages/core/lib/v3/understudy/context.ts
```diff
@@ -17,6 +17,14 @@ type SessionId = string;
 
 type TargetType = "page" | "iframe" | string;
 
+function isNonWebTarget(info: Protocol.Target.TargetInfo): boolean {
+  return (
+    (info.type !== "page" && info.type !== "iframe") ||
+    info.url?.startsWith("chrome-extension://") ||
+    info.url?.startsWith("chrome://")
```
pirate: there are a bunch of others to exclude here:

`about:*`
`data:*`
`file:*`
`view-source:*`
`javascript:`
`blob:*`
`filesystem:*`

`chrome-error://*`
`devtools://*`
`file://*`
`chrome-untrusted://`
`chrome-search://`
`chrome-native://`
`chrome-distiller://`

might be easier to whitelist http(s)?://

check out [chrome://chrome-urls/](brave://chrome-urls/) for more

---

also other target types you might want to include are `tab` and maybe `webview`:

https://source.chromium.org/chromium/chromium/src/+/main:content/browser/devtools/devtools_agent_host_impl.cc?ss=chromium&q=f:devtools%20-f:out%20%22::kTypeTab%5B%5D%22

pirate (approved):

seanmcguire12 (approved):

shrey150:
### Merge activity

* **Feb 20, 5:11 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/browserbase/stagehand/1692).
* **Feb 20, 5:11 AM UTC**: @shrey150 merged this pull request with [Graphite](https://app.graphite.com/github/pr/browserbase/stagehand/1692).

PR #1689, opened 2026-02-16
Upgrade AI SDK and its providers

dylnslck:
# why

Updates the `ai` SDK and all of its providers to the latest versions. This fixes #1645.

# what changed

- **Dependencies**: Upgraded `ai` from ^5.0.133 to ^6.0.0, `@ai-sdk/provider` from ^2.0.0 to ^3.0.0, and all optional AI provider packages (anthropic, openai, google, etc.) to their v3/v4 counterparts.

- **Language model migration**: Migrated from `LanguageModelV2` to `LanguageModelV3` across AISdkClient, LLMClient, and related typings.

- **API changes**: Replaced deprecated `generateObject` / `streamObject` with `generateText` / `streamText` using `Output.object({ schema })`. Added `objectShims.ts` to preserve the `{ object }` and `partialObjectStream` shapes for callers.

- **Tool integration**: Updated all agent tools (click, type, scroll, dragAndDrop, fillFormVision, screenshot, ariaTree, wait) to use the new tool result shape: `toModelOutput` callbacks now receive `{ output }` instead of `result`.

- **Message types**: Switched from `CoreSystemMessage`, `CoreUserMessage`, and `CoreAssistantMessage` to `ModelMessage`.

- **Flow logger**: Added `specificationVersion: "v3"` to the LLM logging middleware.

- **Usage handling**: Updated usage token access for the new structure (`outputTokenDetails?.reasoningTokens`, `inputTokenDetails?.cacheReadTokens`, etc.).

# backwards compatibility

LLMClient keeps the old `generateObject` and `streamObject` API surface. `objectShims.ts` implements these by wrapping `generateText` / `streamText` with `Output.object({ schema })` and mapping the results:
- `generateObjectShim` returns `{ object, ...rest }` (instead of `{ output }`) so callers that destructure `{ object }` still work
- `streamObjectShim` exposes `partialObjectStream` as an alias for `partialOutputStream`

No changes required for code that uses `llmClient.generateObject()` or `llmClient.streamObject()`.

# test plan

- [x] `pnpm install` and `pnpm build` succeed
- [x] `pnpm test` passes (unit/vitest tests)
- [ ] `pnpm e2e:local` or `pnpm e2e` passes to verify agent flows with the upgraded SDK
- [ ] Verified that AISdkClient accepts `LanguageModelV3` models from current provider packages (e.g. `@ai-sdk/openai` ^3.x) without TypeScript errors

---
## Summary by cubic
[...]

PR #1694, opened 2026-02-18
Upgrade AI SDK v5 → v6 with usage null safety fixes

shrey150:
## Summary

- Upgrades `ai` from ^5.0.133 to ^6.0.0, `@ai-sdk/provider` from ^2.0.0 to ^3.0.0, and all optional AI provider packages to their latest major versions.
- Migrates from `LanguageModelV2` to `LanguageModelV3`, `CoreSystemMessage`/`CoreUserMessage`/`CoreAssistantMessage` to `ModelMessage`, and `experimental_generateImage` to `generateImage`.
- Replaces deprecated `generateObject`/`streamObject` with `generateText`/`streamText` + `Output.object()`, with backwards-compatible shims (`objectShims.ts`) to preserve the existing `LLMClient` API surface.
- Updates all agent tool `toModelOutput` callbacks from `(result)` to `({ output })` to match the v6 tool result shape.
- Adds `specificationVersion: "v3"` to LLM logging middleware.
- **Fixes missing optional chaining and deprecated fallbacks** on `outputTokenDetails` and `inputTokenDetails` access in both `aisdk.ts` and `AISdkClientWrapped.ts`, making token usage handling consistent across generateObject and generateText code paths.
- Renames `u` → `usage` variable in generateText IIFE for consistency with the rest of the codebase.
- Updates stale `LanguageModelV2` comment in test file.

Based on #1689 by @dylnslck — thank you for the original upgrade work!

## Test plan

- [x] `pnpm install` and `pnpm build` succeed
- [x] TypeScript typechecks pass (`tsc --noEmit`)
- [x] `pnpm e2e:local` passes — 326 passed, 2 skipped, 0 failures
- [x] Verify AISdkClient accepts `LanguageModelV3` models from current provider packages without TypeScript errors

## Breaking changes for external users

- `AISdkClient` constructor now requires `LanguageModelV3` instead of `LanguageModelV2`. Users must upgrade their `@ai-sdk/*` provider packages to v3+.


shrey150:
@greptileai

packages/core/lib/v3/flowLogger.ts
```diff
@@ -1117,8 +1117,8 @@ export class SessionFileLogger {
             model: modelId,
             operation: "generateText",
             output: outputPreview || "[empty]",
-            inputTokens: result.usage?.inputTokens,
-            outputTokens: result.usage?.outputTokens,
+            inputTokens: result.usage.inputTokens.total,
+            outputTokens: result.usage.outputTokens.total,
```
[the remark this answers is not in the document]
shrey150: incorrect

shrey150:
@greptileai

packages/core/lib/v3/flowLogger.ts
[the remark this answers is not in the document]
shrey150: Investigated this thoroughly — the original `.inputTokens.total` is actually **correct** for this context, but we added a defensive guard for edge cases.

**Why `.total` is correct here:** This code runs inside `wrapGenerate` middleware (not the public `generateText` API). The AI SDK has two different usage shapes:

- **Middleware layer** (`LanguageModelV3GenerateResult.usage`): `inputTokens` is `{ total, noCache, cacheRead, cacheWrite }` — nested object
- **Public API** (`generateText` result): `inputTokens` is a flat `number`

The `@ai-sdk/openai@3.x` provider (v3 spec) transforms raw API responses into the v3 nested format *before* the middleware sees them. Then `asLanguageModelUsage()` flattens it to the public API shape *after* middleware. So `.inputTokens.total` is the right access pattern in middleware.

**Why Greptile was confused:** It saw flat `number` access in `aisdk.ts` (which operates at the public API layer) and assumed the same shape in `flowLogger.ts` (middleware layer).

**What we did anyway:** Added a `typeof` guard ([79688264](https://github.com/browserbase/stagehand/commit/79688264)) because `wrapLanguageModel` doesn't auto-adapt v2 provider results to v3 format. If a user passes a custom `llmClient` with an older v2-spec AI SDK model, the middleware would see flat numbers and `.total` would silently return `undefined`. The guard handles both shapes:

```ts
inputTokens:
  typeof result.usage.inputTokens === "number"
    ? result.usage.inputTokens
    : result.usage.inputTokens.total,
```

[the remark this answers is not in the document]
shrey150: See detailed analysis in the thread above — the original `.total` access is correct for the middleware context. Added a defensive `typeof` guard in ac2666f5 to also handle v2-spec models.

shrey150:
@greptileai

packages/core/lib/v3/flowLogger.ts
```diff
@@ -1117,8 +1117,14 @@ export class SessionFileLogger {
...
-            inputTokens: result.usage?.inputTokens,
-            outputTokens: result.usage?.outputTokens,
+            inputTokens:
+              typeof result.usage.inputTokens === "number"
+                ? result.usage.inputTokens
+                : (result.usage.inputTokens?.total ?? 0),
+            outputTokens:
+              typeof result.usage.outputTokens === "number"
+                ? result.usage.outputTokens
+                : (result.usage.outputTokens?.total ?? 0),
```
pirate: why would these be empty? does aisdk use different fields now?

maybe better to just do

```suggestion
            usage: result.usage,
```

packages/evals/lib/AISdkClientWrapped.ts
```diff
@@ -214,14 +212,21 @@ You must respond in JSON format. respond WITH JSON. Do not include any other tex
         throw err;
       }
 
+      const usage = objectResponse.usage;
       const result = {
-        data: objectResponse.object,
+        data: objectResponse.output,
         usage: {
```
pirate, line 218:
```suggestion
        usage: {
            ...usage,
```

can we just return the whole obj that aisdk returns as well so users can use fields they might see in aisdk docs

packages/evals/lib/AISdkClientWrapped.ts
```diff
@@ -304,13 +309,22 @@ You must respond in JSON format. respond WITH JSON. Do not include any other tex
...
-        prompt_tokens: textResponse.usage.inputTokens ?? 0,
-        completion_tokens: textResponse.usage.outputTokens ?? 0,
-        reasoning_tokens: textResponse.usage.reasoningTokens ?? 0,
-        cached_input_tokens: textResponse.usage.cachedInputTokens ?? 0,
-        total_tokens: textResponse.usage.totalTokens ?? 0,
-      },
+      usage: (() => {
+        const usage = textResponse.usage;
+        return {
+          prompt_tokens: usage.inputTokens ?? 0,
```
pirate, line 317:
```suggestion
        const usage = textResponse.usage;
        return {
          ...usage,
          prompt_tokens: usage.inputTokens ?? 0,
```
ditto here

pirate (approved):

packages/core/lib/v3/flowLogger.ts
```diff
@@ -1117,8 +1117,8 @@ export class SessionFileLogger {
             model: modelId,
             operation: "generateText",
             output: outputPreview || "[empty]",
-            inputTokens: result.usage?.inputTokens,
-            outputTokens: result.usage?.outputTokens,
+            inputTokens: result.usage.inputTokens.total ?? 0,
+            outputTokens: result.usage.outputTokens.total ?? 0,
```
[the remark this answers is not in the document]
shrey150: False positive

pirate (approved):

packages/core/lib/v3/external_clients/aisdk.ts
```diff
@@ -82,20 +79,29 @@ export class AISdkClient extends LLMClient {
     );
```
tkattkat, line 79: nit: we can continue using the specific types for each of the types of messages rather than the union

seems they get imported as this now -   
SystemModelMessage,
UserModelMessage,
AssistantModelMessage,

packages/core/lib/v3/external_clients/aisdk.ts
```diff
@@ -115,14 +121,21 @@ export class AISdkClient extends LLMClient {
...
-        total_tokens: response.usage.totalTokens ?? 0,
+        prompt_tokens: usage.inputTokens ?? 0,
+        completion_tokens: usage.outputTokens ?? 0,
+        reasoning_tokens:
+          usage.outputTokenDetails?.reasoningTokens ??
+          usage.reasoningTokens ??
+          0,
+        cached_input_tokens:
+          usage.inputTokenDetails?.cacheReadTokens ??
+          usage.cachedInputTokens ??
```
tkattkat, line 136: It seems these fallbacks are deprecated usage.reasoningTokens,  usage.cachedInputTokens . we probably do not need the fallback to reading from them when we check new version already

packages/core/lib/v3/handlers/v3AgentHandler.ts
```diff
@@ -553,8 +557,14 @@ export class V3AgentHandler {
...
-            reasoning_tokens: result.totalUsage.reasoningTokens || 0,
-            cached_input_tokens: result.totalUsage.cachedInputTokens || 0,
+            reasoning_tokens:
+              result.totalUsage.outputTokenDetails?.reasoningTokens ??
+              result.totalUsage.reasoningTokens ??
+              0,
+            cached_input_tokens:
+              result.totalUsage.inputTokenDetails?.cacheReadTokens ??
+              result.totalUsage.cachedInputTokens ??
+              0,
```
tkattkat, line 567: same for the fallbacks here

packages/core/lib/v3/flowLogger.ts
```diff
@@ -1006,17 +1006,17 @@ export class SessionFileLogger {
...
     // No-op middleware when logging is disabled
     if (!CONFIG_DIR) {
       return {
+        specificationVersion: "v3" as const,
         wrapGenerate: async ({ doGenerate }) => doGenerate(),
       };
     }
 
     return {
+      specificationVersion: "v3" as const,
```
tkattkat, line 1019: casting to const is unnecessary in both cases

packages/evals/lib/AISdkClientWrapped.ts
```diff
@@ -1,17 +1,15 @@
...
-  CoreUserMessage,
   ImagePart,
   NoObjectGeneratedError,
+  Output,
   TextPart,
   ToolSet,
   Tool,
 } from "ai";
 import * as ai from "ai";
 import { wrapAISDK } from "braintrust";
```
tkattkat, line 11: Have we confirmed this works with evals CLI. I know between v4 and v5 braintrust changed the way its ai sdk integration works. Unsure if they changed anything for v6 that would cause issues though

packages/evals/lib/AISdkClientWrapped.ts
```diff
@@ -117,7 +115,7 @@ export class AISdkClientWrapped extends LLMClient {
               type: "text" as const,
               text: part.type === "image" ? "[Image]" : part.text,
             }));
-            const assistantMessage: CoreAssistantMessage = {
+            const assistantMessage: ModelMessage = {
               role: "assistant",
               content: textOnlyParts,
             };
```
tkattkat, line 121: nit use types specific to user, system, and assistant message rather than the union

packages/core/lib/v3/external_clients/aisdk.ts
pirate: could someone pass in an older client object when using a custom client?

packages/core/lib/v3/llm/objectShims.ts
```diff
@@ -0,0 +1,118 @@
+/**
+ * Thin shims that wrap generateText and streamText with Output.object({ schema })
+ * as the replacement for deprecated generateObject and streamObject from the AI SDK.
+ * Callers must supply a schema; it is passed through to Output.object() for structured output.
+ */
+
```
tkattkat, line 6: Instead of the shim I would just recommend leaving generateObject as is. It is deprecated so it can still be used, and is currently not used in our package itself. When we eventually move to a version that does not have it, we can likely just tell people we no longer support it and point them to newer methods

tkattkat (changes requested):

packages/core/lib/v3/external_clients/aisdk.ts
tkattkat: I do not think so since its scoped to v3 providers. old ones would be v2

seanmcguire12 (changes requested):
this is breaking for users passing in tools to agent, & also people passing in custom aisdk clients. 

our options are:
1. figure out a way to make this backward compatible
2. hold off on this migration until the next major release

I would rather us hold off until the next major given the complexity & branching logic to required to support both

PR #1708, opened 2026-02-19
[feat]: add `context.setExtraHTTPHeaders()`

seanmcguire12:
# why
- to enable users to set headers to be included in every request created by every page within the `context`
# what changed
- added a new function `context.setExtraHTTPHeaders()` which calls the `Network.setExtraHTTPHeaders` CDP command on all existing sessions
- we also call `Network.setExtraHTTPHeaders` with the user defined headers whenever we attach to a new target (via `onAttachedToTarget`)
- if there is a rejected promise when headers are being set, a `StagehandSetExtraHTTPHeadersError` will be thrown
# test plan
- added an e2e test to confirm that headers are being correctly set by listening to `Network.requestWillBeSent` and inspecting headers
- also added unit tests to confirm that the custom error is being thrown correctly when promises reject

---
## Summary by cubic
Adds context.setExtraHTTPHeaders() to set default HTTP headers on every request from all pages in a context. Implements STG-1411; applies to current sessions and newly attached targets after Network.enable.

- **New Features**
  - New API: context.setExtraHTTPHeaders(headers) sends Network.enable then Network.setExtraHTTPHeaders to each active session; repeats on attach for new targets.
  - Public error: StagehandSetExtraHTTPHeadersError with per-session failure details.
  - Tests: e2e verifies navigation headers; unit tests cover multi-session send and error reporting; public error type is exported.

<sup>Written for commit 21606892ebf693f7798d3d4fea695c1563caa16f. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1708">Review in cubic</a></sup>

### DIFF
--- a/packages/core/lib/v3/understudy/context.ts
+++ b/packages/core/lib/v3/understudy/context.ts
@@ -121,8 +121,6 @@ export class V3Context {
   private pendingCreatedTargetUrl = new Map<TargetId, string>();
   private readonly initScripts: string[] = [];
   private extraHttpHeaders: Record<string, string> | null = null;
-  // Monotonic token to avoid stale rollbacks when overlapping header updates run.
-  private extraHttpHeadersVersion = 0;
 
   private installTargetSessionListeners(session: CDPSessionLike): void {
     const sessionId = session.id;
@@ -342,12 +340,7 @@ export class V3Context {
   public async setExtraHTTPHeaders(
     headers: Record<string, string>,
   ): Promise<void> {
-    const previousHeaders = this.extraHttpHeaders
-      ? { ...this.extraHttpHeaders }
-      : null;
     const nextHeaders = { ...headers };
-    // Capture a version token so overlapping calls don't roll back newer updates.
-    const version = ++this.extraHttpHeadersVersion;
     this.extraHttpHeaders = nextHeaders;
 
     const sessions: CDPSessionLike[] = [];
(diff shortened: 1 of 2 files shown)


packages/core/lib/v3/understudy/context.ts
```diff
@@ -274,6 +281,70 @@ export class V3Context {
...
+
+    const failures = await applyHeaders(nextHeaders, version);
+    if (failures.length) {
+      // Only roll back if this call is still the latest.
+      if (version === this.extraHttpHeadersVersion) {
+        this.extraHttpHeaders = previousHeaders;
+        const rollbackHeaders = previousHeaders ?? {};
+        await applyHeaders(rollbackHeaders, version).catch(() => {});
+      }
+      throw new StagehandSetExtraHTTPHeadersError(failures);
```
seanmcguire12, line 381: @pirate curious if you think it makes sense to throw here? vs maybe a catch & log or a catch & swallow. could be annoying to throw given the level of target attach/detach churn

seanmcguire12:
@cubic-dev-ai

packages/core/lib/v3/understudy/context.ts
```diff
@@ -274,6 +281,70 @@ export class V3Context {
...
+    headers: Record<string, string>,
+  ): Promise<void> {
+    const previousHeaders = this.extraHttpHeaders
+      ? { ...this.extraHttpHeaders }
+      : null;
+    const nextHeaders = { ...headers };
+    // Capture a version token so overlapping calls don't roll back newer updates.
+    const version = ++this.extraHttpHeadersVersion;
+    this.extraHttpHeaders = nextHeaders;
+
```
pirate: I dont think versioning is needed if we just make calls idempotent mapping updates.

Object updates are atomic in JS so idempotent updates should be possible:

```ts
send("Network.enable")

for (const [key, val] of Object.entries(nextHeaders)) {
    try {
		// incrementally update set headers, one new key:val at a time
        send("Network.setExtraHTTPHeaders", { headers: this.extraHttpHeaders })
        this.extraHttpHeaders[key] = val   // last write wins
    } catch(e) {
        // if any header setting fails, dont persist it in the cumulative state
        console.log(`Failed to set header: ...; skipping`)
    }
}
```

That way a user can do:

```ts
ctx.setExtraHTTPHeaders({'x-some-key': 'some-val'})
// then later
ctx.setExtraHTTPHeaders({'x-some-other-key': 'some-otherval'})
```

and the second will merge cleanly with the first instead of resetting the entire headers state.

If they want to clear existing headers they can always do:

```ts
ctx.setExtraHTTPHeaders({'x-some-key': null})
ctx.setExtraHTTPHeaders({'x-some-other-key': null})

// or nuke everything
const headers_reset = Object.fromEntries(Object.entries(ctx.extraHttpHeaders).map(([key, val]) => [key, null]))
ctx.etExtraHTTPHeaders(headers_reset)
```

seanmcguire12: merge behaviour diverges from playwright behaviour, which is what users are likely to expect. every subsequent call to `context.setExtraHTTPHeaders()` overwrites whatever was there before. merging only happens with `page.setExtraHTTPHeaders()`. clearing headers should be done like this `context.setExtraHTTPHeaders({})`

seanmcguire12:
@cubic-dev-ai

packages/core/lib/v3/understudy/context.ts
seanmcguire12: removed the versioning token logic in [this](https://github.com/browserbase/stagehand/pull/1708/commits/21606892ebf693f7798d3d4fea695c1563caa16f) one

PR #1719, opened 2026-02-20
Solve init script race with Debugger.resume at frame init script injection time

pirate:
# why

Init script injection was racing with Debugger.resume() sometimes, causing frames to load without init scripts running sometimes. This led to flaky init script tests, which were legitimately catching the issue.

- https://github.com/browserbase/stagehand/actions/runs/22233062982/job/64336364420?pr=1580

image

# what changed

- queues calls on page load to run before we resume
- catches oopifs and lazy frames and click-triggered popups the same way playwright does
- removes flaky timeout/retry based prior approach

https://deepwiki.com/search/how-does-playwright-guarantee_8cf2339b-c060-4cfc-bc62-f3baaf57b229?mode=deep

# test plan

---
## Summary by cubic
Fixes the init‑script race by guaranteeing pre‑resume setup and correcting popup attach order. Init scripts now run reliably in same‑ and cross‑process popups, OOPIF iframes, and across reloads; race tests verify addScript is sent before resume per session.

- **Bug Fixes**
  - Enforce pre‑resume ordering: per‑session dispatch waiters ensure Page/Runtime enables, Target.setAutoAttach(waitForDebuggerOnStart), Network.enable/setExtraHTTPHeaders, and Page.addScriptToEvaluateOnNewDocument(runImmediately) are sent before Runtime.runIfWaitingForDebugger; resume only after dispatch; log ordering issues only for top‑level pages.
  - Stabilize attach and evaluation: fix popup attach ordering; fan out Target.* events to root listeners; retry Runtime.evaluate once on stale context ids; pre‑register the piercer script before resume and lazy‑install if needed.
  - Harden lifecycle: convert detach errors to PageNotFoundError and propagate; treat Page.enable/lifecycle acks as best‑effort; never drop top‑level Page.create due to local timeouts.
  - Expand tests and deflake: add delayed‑CDP‑send popup/iframe race repro with real URLs; assert addScript precedes resume per session; cover in‑process and cross‑process popups, window.open, OOPIF iframes, and reload persistence; update detach expectations and timeouts.

<sup>Written for commit 6f464d377a74c8430c526d06d2d3def005416e28. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1719">Review in cubic</a></sup>

### DIFF
--- a/packages/core/tests/integration/context-addInitScript.spec.ts
+++ b/packages/core/tests/integration/context-addInitScript.spec.ts
@@ -4,25 +4,31 @@ import { v3TestConfig } from "./v3.config.js";
 import { V3Context } from "../../lib/v3/understudy/context.js";
 import type { Page as V3Page } from "../../lib/v3/understudy/page.js";
 
+const POPUP_TIMEOUT_MS = 20_000;
+
 const toDataUrl = (html: string): string =>
   `data:text/html,${encodeURIComponent(html)}`;
 
 const waitForPopupPage = async (
   ctx: V3Context,
-  opener: V3Page,
-  timeoutMs = 5000,
+  knownTargetIds: Set<string>,
+  timeoutMs = POPUP_TIMEOUT_MS,
 ): Promise<V3Page> => {
-  const openerId = opener.targetId();
   const deadline = Date.now() + timeoutMs;
-  let popup = ctx.pages().find((p) => p.targetId() !== openerId);
-  while (!popup && Date.now() < deadline) {
-    await opener.waitForTimeout(25);
-    popup = ctx.pages().find((p) => p.targetId() !== openerId);
+  while (Date.now() < deadline) {
+    const popup = ctx
+      .pages()
+      .find((page) => !knownTargetIds.has(page.targetId()));
+    if (popup) return popup;
+    try {
+      const active = await ctx.awaitActivePage(500);
+      if (!knownTargetIds.has(active.targetId())) return active;
+    } catch {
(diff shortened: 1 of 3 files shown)


packages/core/tests/integration/context-addInitScript.spec.ts
seanmcguire12, line 1: lets try & get test that fails on main so we can make sure the cdp cmd sequencing stays bulletproof

pirate: added in `iframe-ctx-addInitScript-race.spec.ts`

packages/core/tests/integration/iframe-ctx-addInitScript-race.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+
+const isBrowserbase =
+  (process.env.STAGEHAND_BROWSER_TARGET ?? "local").toLowerCase() ===
+  "browserbase";
+
+const POPUP_URL_SUBSTRING =
+  "/stagehand-eval-sites/sites/closed-shadow-dom-in-spif/";
+const POPUP_IFRAME_URL =
+  "https://browserbase.github.io/stagehand-eval-sites/sites/closed-shadow-dom-in-spif/embedded.html";
+const INIT_SCRIPT_DELAY_MS = Number(
```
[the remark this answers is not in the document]
pirate: Whoops, didn't catch that, will fix.

PR #1730, opened 2026-02-22
[chore]: configure test coverage with env var

seanmcguire12:
# why
- to make test coverage outputs configurable via an environment variable so that you can run tests locally without clogging up storage unnecessarily
# what changed
- moved `packages/core/scripts/test-utils.ts` to th workspace root level `scripts/test-utils.ts` so that utils can be shared across other test scripts (`test-core.ts`, `test-e2e.ts`, `test-server.ts`, `test-evals.ts`)
- added `scripts/test-artifacts.ts` which handles conditionally writing test artifacts based on the `TEST_COVERAGE` env var

---
## Summary by cubic
Make test coverage and JUnit/CTRF artifact generation opt-in via the TEST_COVERAGE env var to avoid writing artifacts during local runs. CI sets TEST_COVERAGE=1 to keep coverage and reports in pipelines.

- **New Features**
  - Added scripts/test-artifacts.ts with helpers to conditionally create coverage dirs, JUnit paths, and env (NODE_V8_COVERAGE, CTRF_JUNIT_PATH), and to unset them when disabled.
  - Updated core, e2e, evals, and server test scripts to only set reporters/paths and write coverage/JUnit/CTRF when TEST_COVERAGE is enabled; env vars are ignored when no path is provided.
  - Moved test-utils to workspace root for sharing; updated imports accordingly.
  - CI config sets TEST_COVERAGE=1; removed hardcoded NODE_V8_COVERAGE in evals job.
  - Excluded scripts from package tsconfigs.

- **Migration**
  - To generate coverage locally: export TEST_COVERAGE=1 (or true) before running test scripts.

<sup>Written for commit 9852ae822e6f3bf2f14224ac66810f32b6c3ab90. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1730">Review in cubic</a></sup>

### DIFF
--- a/scripts/test-artifacts.ts
+++ b/scripts/test-artifacts.ts
@@ -28,12 +28,22 @@ export const initJunitPath = (
 export const withCoverageEnv = (
   env: Record<string, string | undefined>,
   coverageDir: string | null,
-) => (coverageDir ? { ...env, NODE_V8_COVERAGE: coverageDir } : env);
+) => {
+  if (coverageDir) return { ...env, NODE_V8_COVERAGE: coverageDir };
+  const cleaned = { ...env };
+  delete cleaned.NODE_V8_COVERAGE;
+  return cleaned;
+};
 
 export const withCtrfEnv = (
   env: Record<string, string | undefined>,
   junitPath: string | null,
-) => (junitPath ? { ...env, CTRF_JUNIT_PATH: junitPath } : env);
+) => {
+  if (junitPath) return { ...env, CTRF_JUNIT_PATH: junitPath };
+  const cleaned = { ...env };
+  delete cleaned.CTRF_JUNIT_PATH;
+  return cleaned;
+};
 
 export const maybeWriteCtrf = (
   junitPath: string | null,


seanmcguire12:
@cubic-dev-ai

packages/core/tsconfig.json
```diff
@@ -13,5 +13,5 @@
     }
   },
   "include": ["**/*.ts", "lib/v3/cli.js"],
-  "exclude": ["node_modules", "dist"]
+  "exclude": ["node_modules", "dist", "scripts"]
```
pirate, line 16: why exclude scripts?

pirate:
> moved packages/core/scripts/test-utils.ts to th workspace root level scripts/test-utils.ts so that utils can be shared across other test scripts (test-core.ts, test-e2e.ts, test-server.ts, test-evals.ts)

I dont think this is a good idea, it significantly complicates the build because it introduces a cross-repo dependency at the tsx level, it's not worth it for simple helpers.

pirate:
here this is a simpler way to accomplish the same thing without branching in all the test scripts: https://github.com/browserbase/stagehand/pull/1733

packages/core/tsconfig.json
seanmcguire12: the scripts all use tsx

seanmcguire12:
> here this is a simpler way to accomplish the same thing without branching in all the test scripts: #1733

yeah i was intentionally shooting for "don't write in the first place" behaviour, instead of "write, then remove"

PR #1744, opened 2026-02-24
When cdp url is set, but browser type is not set throw a 400

tkattkat:
# why

Currently, if a cdp url is set, but browser.type is not set to local a session is created on api 

# what changed

We now warn people that they need to set the browser.type to local to properly use their own custom cdp url 

# test plan

tested locally with http requests + from go client + added tests

---
## Summary by cubic
Return 400 on POST /v1/sessions/start when `browser.cdpUrl` is provided but `browser.type` is not "local", preventing unintended session creation and guiding correct local CDP usage.

- **Bug Fixes**
  - Reject requests with `cdpUrl` when `browser.type` is missing or not "local"; respond 400 with a clear message to set `browser.type` to "local".
  - Keep local validation: when `browser.type` is "local", require `cdpUrl` or `launchOptions`.
  - Add integration tests for missing type and `type="browserbase"` with `cdpUrl`.

<sup>Written for commit 58a6c551387d168ceb0f5273e4bb95965771eead. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1744">Review in cubic</a></sup>


packages/server-v3/src/routes/v1/sessions/start.ts
```diff
@@ -29,12 +29,21 @@ const startBodySchema = z
     return value;
   }, Api.SessionStartRequestSchema)
   .superRefine((value, ctx) => {
+    if (value.browser?.cdpUrl && value.browser.type !== "local") {
```
seanmcguire12, line 32: @tkattkat this is strict inequality so it would throw if `"LOCAL"` is passed in. does browser.type get normalized? 

would be worth adding a test to prove that this behaves as expected with both upper & lowercase

seanmcguire12 (changes requested):

packages/server-v3/src/routes/v1/sessions/start.ts
tkattkat: the zod enum will reject it if it is not properly cased, so it can never reach here improperly cased

seanmcguire12 (approved):

PR #1765, opened 2026-02-26
bump @browserbasehq/sdk to ^2.7.0, make projectId optional

derekmeegan:
# why

SDK v2.7.0 makes projectId optional for session creation. Update all references to treat projectId as optional rather than required.

# what changed

new browserbase sdk version and optional project id

# test plan


shrey150:
Closing — these changes were already landed via #1800.

PR #1766, opened 2026-02-26
[feat]: add configurable timeout to agent tools

tkattkat:
# Why

Agent tool calls can hang indefinitely when the underlying operation stalls (e.g. waiting on a slow LLM inference, an unresponsive page, or a network timeout). When this happens the entire agent loop blocks and the user has no recourse other than killing the process. We need a safety net that bounds how long any single tool call can take and returns control to the LLM so it can adapt.

# What changed

We now pass timeout into the tools for ones that involve llm inference within the tool call 

---
## Summary by cubic
Adds a configurable per-tool timeout to agent tools to prevent hangs, defaulting to 45s. Implements real cancellation for inference-backed tools and standardized JSON error outputs, with clear timeout warnings so the agent can adapt.

- **New Features**
  - agent.execute/stream accept toolTimeout (ms); default 45s; values ≤ 0 are ignored.
  - Timeouts forwarded to v3 for act, extract, fillForm (observe + act), and ariaTree; TimeoutError handled with helpful messages; error outputs use consistent JSON via toModelOutput.
  - AriaTree supports toolTimeout and returns JSON errors on timeout.

- **Bug Fixes**
  - Standardized success/error payloads across tools (click, dragAndDrop, type, fillFormVision, screenshot, scroll, wait, ariaTree).
  - Centralized timeout handling with withTimeout/createTimeoutGuard; non-positive timeouts bypassed; errors returned to the agent instead of thrown.
  - Tests: added e2e toolTimeout coverage for act/extract/fillForm/ariaTree; removed msw; relaxed screenshot timeout regex; updated test strings to match timeout errors.

<sup>Written for commit 1f0cb4f8f0fb6c39d7110032bb74181f625ed4d9. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1766">Review in cubic</a></sup>


packages/core/lib/v3/agent/tools/ariaTree.ts
```diff
@@ -1,33 +1,52 @@
...
+            "\n\n[CONTENT TRUNCATED: Exceeded 70,000 token limit]";
+        }
 
-      return { content, pageUrl };
+        return { content, pageUrl };
+      } catch (error) {
+        if (error instanceof TimeoutError) {
+          return {
+            content: `ariaTree timed out — the page may be too large, try using extract with a targeted selector instead`,
+            pageUrl: "",
```
pirate:
```suggestion
			error: `TimeoutError: ariaTree() timed out — the page may be too large, try using extract with a targeted selector instead`,
			success: false,
            content: "",
            pageUrl,
```

packages/core/lib/v3/agent/tools/extract.ts
```diff
@@ -94,11 +96,17 @@ export const extractTool = (
...
+          timeout: toolTimeout,
         });
         return { success: true, result };
       } catch (error) {
-        const err = error as Error;
-        return { success: false, error: err?.message ?? String(error) };
+        if (error instanceof TimeoutError) {
+          return {
+            success: false,
+            error: `extract timed out — try using a smaller or simpler schema`,
```
pirate:
```suggestion
            error: `TimeoutError: extract() timed out — try using a smaller or simpler schema`,
```

packages/core/lib/v3/agent/tools/act.ts
```diff
@@ -60,7 +63,16 @@ export const actTool = (
         }
         return response;
       } catch (error) {
-        return { success: false, error: error?.message ?? String(error) };
+        if (error instanceof TimeoutError) {
+          return {
+            success: false,
+            error: `act timed out — try using a different description for the action`,
```
pirate:
```suggestion
            error: `TimeoutError: act() timed out — try using a different description for the action`,
```

packages/core/lib/v3/agent/tools/fillform.ts
```diff
@@ -31,47 +33,62 @@ export const fillFormTool = (
...
+        return {
+          success: true,
+          actions: completed,
+          playwrightArguments: replayableActions,
+        };
+      } catch (error) {
+        if (error instanceof TimeoutError) {
+          return {
+            success: false,
+            error: `fillForm timed out — try filling fewer fields at once or use a different tool`,
```
pirate:
```suggestion
            error: `TimeoutError: fillForm() timed out — try filling fewer fields at once or use a different tool`,
```

packages/core/lib/v3/agent/tools/ariaTree.ts
```diff
@@ -1,33 +1,55 @@
...
+        if (error instanceof TimeoutError) {
+          return {
+            content: "",
+            error: `TimeoutError: ariaTree() timed out — the page may be too large`,
+            success: false,
+            pageUrl: "",
+          };
+        }
+        return {
+          content: `Error: ${error?.message ?? String(error)}`,
```
pirate: I would also not put the error in the content here
```suggestion
          content: "",
```

packages/core/lib/v3/agent/tools/extract.ts
```diff
@@ -94,11 +96,17 @@ export const extractTool = (
...
+          timeout: toolTimeout,
         });
         return { success: true, result };
       } catch (error) {
-        const err = error as Error;
-        return { success: false, error: err?.message ?? String(error) };
+        if (error instanceof TimeoutError) {
+          return {
+            success: false,
+            error: `TimeoutError: extract() timed out — try using a smaller or simpler schema`,
```
pirate:
```suggestion
            error: `TimeoutError: extract() timed out — try using a smaller or simpler schema`,
            content: "",
```

Do we want to provide an empty content here too for consistency?

packages/core/lib/v3/agent/tools/ariaTree.ts
```diff
@@ -1,33 +1,56 @@
...
 
-      return { content, pageUrl };
+        return { content, pageUrl };
+      } catch (error) {
+        if (error instanceof TimeoutError) {
+          return {
+            content: "",
+            error: `TimeoutError: ariaTree() timed out — the page may be too large`,
+            success: false,
+            pageUrl: "",
```
pirate, line 50:
```suggestion
            pageUrl,
```

packages/core/lib/v3/agent/tools/ariaTree.ts
```diff
@@ -1,33 +1,56 @@
...
+          };
+        }
+        return {
+          content: "",
+          error: error?.message ?? String(error),
+          success: false,
+          pageUrl: "",
```
pirate, line 57:
```suggestion
          pageUrl,
```

packages/core/lib/v3/agent/tools/index.ts
```diff
@@ -90,15 +95,16 @@ export function createAgentTools(v3: V3, options?: V3AgentToolOptions) {
...
   const excludeTools = options?.excludeTools;
   const variables = options?.variables;
+  const toolTimeout = options?.toolTimeout;
 
   const allTools: ToolSet = {
-    act: actTool(v3, executionModel, variables),
-    ariaTree: ariaTreeTool(v3),
+    act: actTool(v3, executionModel, variables, toolTimeout),
+    ariaTree: ariaTreeTool(v3, toolTimeout),
     click: clickTool(v3, provider),
```
seanmcguire12, line 103: should these other tools have timeouts as well?

pirate:
Sean caught some good things:

- we should change the wording to tell the LLM for all timeout errors: "waiting for xyz() timed out, but may continue in the background -- try ..." because we dont actually cancel it (it's impossible in JS, we just stop waiting for the result)

- we should add a `logger.warn(...)` log about the timeout being hit

packages/core/lib/v3/agent/tools/act.ts
```diff
@@ -60,7 +63,16 @@ export const actTool = (
         }
         return response;
       } catch (error) {
-        return { success: false, error: error?.message ?? String(error) };
+        if (error instanceof TimeoutError) {
+          return {
+            success: false,
+            error: `TimeoutError: act() timed out — try using a different description for the action`,
```
seanmcguire12: maybe change the wording here to let the LLM know that the action may have actually still taken place even if the timeout was hit. 

also, we should tell the LLM how long the timeout was: eg "act timed out after x milliseconds"

packages/core/lib/v3/agent/tools/index.ts
pirate: imo yes but I didnt want to blow up @tkattkat's PR too much lol

packages/core/lib/v3/agent/tools/act.ts
seanmcguire12: we might also want to log something here so that the user can see that the tool timed out

packages/core/lib/v3/agent/tools/ariaTree.ts
```diff
@@ -1,36 +1,76 @@
...
+          const maxChars = MAX_TOKENS * 4;
+          content =
+            content.substring(0, maxChars) +
+            "\n\n[CONTENT TRUNCATED: Exceeded 70,000 token limit]";
+        }
+
+        return { success: true, content, pageUrl };
+      } catch (error) {
+        if (error instanceof TimeoutError) {
+          const timeoutMessage = `TimeoutError: ariaTree() timed out — the page may be too large`;
```
seanmcguire12, line 40: approved but would be good to include the number of milliseconds in the timeout in both the log & the message to the agent. eg, "x timed out after y milliseconds"

this info is already included in `error.message` here

PR #1781, opened 2026-03-04
[STG-1499] Reduce Google CUA payload size to mitigate INVALID_ARGUMENT errors

derekmeegan:
## Problem

Customers running Google CUA agents (`google/gemini-2.5-computer-use-preview-10-2025`) at high concurrency (40+ parallel agents) are seeing ~7% failure rates with opaque `INVALID_ARGUMENT` 400 errors from the Gemini API.

**Root cause**: Screenshots sent as raw uncompressed PNG base64 inline data in `functionResponse` parts. A 1288×711 viewport PNG can be 1-4 MB depending on page content. With conversation history keeping 2 full screenshots (via `compressGoogleConversationImages`), the total payload can exceed undocumented limits on the Gemini CUA preview model. The non-deterministic failure pattern (7% of runs) matches this theory — screenshot sizes vary dramatically based on page visual complexity at capture time.

**Ref**: Pylon #14952, Linear [STG-1499](https://linear.app/browserbase/issue/STG-1499)

### 1. JPEG screenshots for Google CUA (`v3CuaAgentHandler.ts`)
- Screenshot provider now captures JPEG at 80% quality when the agent is a `GoogleCUAClient`
- Also applies to `captureAndSendScreenshot()` (used after each action)
- **Impact**: Reduces each screenshot from ~1-4 MB (PNG) to ~100-300 KB (JPEG 80%), roughly a 5-10x reduction in payload size
- Other CUA clients (OpenAI, Anthropic, Microsoft) continue using PNG

### 2. Payload size logging (`GoogleCUAClient.ts`)
- Logs estimated payload size in MB and history item count before each `generateContent` call
- Uses `estimatePayloadSize()` helper that measures base64 image data + text + function call JSON
- Enables diagnosis of payload-size-related failures without code changes

### 3. Skip retries for INVALID_ARGUMENT (`GoogleCUAClient.ts`)
- Detects `INVALID_ARGUMENT` in error messages and fails immediately instead of retrying 5 times with exponential backoff
- These errors are deterministic for the same payload — retrying wastes 1+2+4+8+16 = 31 seconds before failing anyway
- Other transient errors (rate limits, network issues) still retry normally

### 4. JPEG format detection (`GoogleCUAClient.ts`)
- `captureScreenshot()` now detects JPEG vs PNG from base64 data header (`/9j` prefix = JPEG)
- `functionResponse` uses the correct `mimeType` (`image/jpeg` or `image/png`) based on actual format
- Backwards compatible — still works with PNG screenshots from other sources

## Testing

- [ ] Run Google CUA agent evals to verify screenshots render correctly as JPEG
- [ ] Verify JPEG quality at 80% provides sufficient visual fidelity for the model
- [ ] Check that payload size logging appears in agent output at log level 2
- [ ] Confirm INVALID_ARGUMENT errors fail fast (no retry delay)
- [ ] Verify OpenAI/Anthropic CUA agents still use PNG (no regression)

### DIFF
--- a/packages/core/lib/v3/handlers/v3CuaAgentHandler.ts
+++ b/packages/core/lib/v3/handlers/v3CuaAgentHandler.ts
@@ -63,11 +63,18 @@ export class V3CuaAgentHandler {
 
   private setupAgentClient(): void {
     // Provide screenshots to the agent client
+    // Use JPEG for Google CUA to reduce payload size significantly
+    // (PNG: 1-4 MB per screenshot, JPEG at 80%: 100-300 KB)
+    // This helps avoid INVALID_ARGUMENT errors from the Gemini API
+    // when conversation history accumulates large screenshot payloads
+    const useJpeg = this.agentClient instanceof GoogleCUAClient;
     this.agentClient.setScreenshotProvider(async () => {
       this.ensureNotClosed();
       const page = await this.v3.context.awaitActivePage();
-      const screenshotBuffer = await page.screenshot({ fullPage: false });
-      return screenshotBuffer.toString("base64"); // base64 png
+      const screenshotBuffer = useJpeg
+        ? await page.screenshot({ fullPage: false, type: "jpeg", quality: 80 })
+        : await page.screenshot({ fullPage: false });
+      return screenshotBuffer.toString("base64");
     });
 
     // Provide action executor
@@ -594,7 +601,10 @@ export class V3CuaAgentHandler {
     });
     try {
       const page = await this.v3.context.awaitActivePage();
-      const screenshotBuffer = await page.screenshot({ fullPage: false });
+      const useJpeg = this.agentClient instanceof GoogleCUAClient;
+      const screenshotBuffer = useJpeg
+        ? await page.screenshot({ fullPage: false, type: "jpeg", quality: 80 })
+        : await page.screenshot({ fullPage: false });
 
       // Emit screenshot event via the bus
       this.v3.bus.emit("agent_screenshot_taken_event", screenshotBuffer);


shrey150:
## Review: Converted to Draft — Needs Investigation Before Merging

### Build is Broken
`payloadSizeMB` is declared with `const` inside `try {}` (line 433) but referenced in `catch {}` (line 479). This is a compile error — TS2304. Fix: move the declaration above the `for` loop.

### Core Hypothesis (Payload Size) is Plausible But Unproven

After deep investigation, here's what I found:

**The FunctionResponse structure is correct.** I traced through the `@google/genai` v1.24.0 SDK serialization chain (`partToMldev$1` → `contentToMldev$1` → JSON.stringify). The `functionResponse` field gets a **direct passthrough** with zero transformation — whatever JS object we put in goes straight to the wire. The field names (`parts`, `inlineData`, `mimeType`, `data`) match the Python reference implementation exactly.

**Intermittent INVALID_ARGUMENT is a known Gemini API-side issue.** Multiple users on the [Google AI Developers Forum](https://discuss.ai.google.dev/t/getting-400-invalid-argument-intermittently/122930) reported the same non-deterministic 400 errors starting around Feb 2026, with no code changes on their side, and some cases resolved spontaneously. This matches the customer's 7% pattern.

**There's also a [known Gemini bug with function call/response parity](https://github.com/google-gemini/gemini-cli/issues/16135):** the API returns `"Please ensure that the number of function response parts is equal to the number of function call parts"` when there's a mismatch. While Stagehand's parity logic *looks* correct, there are edge cases (e.g., if `convertFunctionCallToAction` returns null for an unsupported function name, the `result.actions.length > 0` guard at line 549 could skip function response creation entirely, violating parity). Worth investigating whether the customer's tasks trigger any unusual function calls.

**Google's reference implementation uses `MAX_RECENT_TURN_WITH_SCREENSHOTS = 3`** while Stagehand uses `keepMostRecentCount = 2`. Both strip older screenshots. The Google impl also retries ALL exceptions indiscriminately (no INVALID_ARGUMENT skip) — suggesting Google doesn't consider this error deterministic.

### What I'd Recommend Instead

1. **Ask the customer to log the full error response** — not just `error.message`, but the raw HTTP response body. Google sometimes includes `details[]` with field-level validation info that gets lost in the SDK's error wrapping.

2. **Ask the customer to upgrade to 3.0.8+** — the coordinate sanitization fix (PR #1560) changed how out-of-range coordinates are handled in history. This is a separate potential cause.

3. **Ship JPEG optimization separately** as a general performance improvement (not as a fix for this specific bug). It's the right thing to do regardless — PNG screenshots are wastefully large.

4. **Don't skip INVALID_ARGUMENT retries** — Google's own reference implementation retries all errors. The error may not be deterministic for the same payload (server-side A/B testing of validation rules for preview models).

PR #1792, opened 2026-03-06
feat: add support for gpt 5.4 native computer use

Kylejeong2:
# why

# what changed

# test plan


miguelg719:
getting #1892 in instead

PR #1807, opened 2026-03-11
fix(agent): fillForm tool drops caller-provided values for LLM-hallucinated ones

### LINKED ISSUE #1789: fillForm agent tool drops value parameter, causing LLM to hallucinate fill values
## Summary

The `fillForm` agent tool receives a `value` parameter but never uses it. When the tool calls `observe()`, it only passes the `action` description, causing the LLM to hallucinate placeholder values like `test@example.com` instead of using the actual provided value. This bug also affects custom tool results (e.g., TOTP codes from 2FA tools).

## Affected File

`packages/core/lib/v3/agent/tools/fillform.ts`

## Reproduction

Using the agent to fill a login form:

```typescript
const agent = stagehand.agent({
  model: { modelName: "openai/gpt-5.1-2025-11-13" },
  systemPrompt: "Log into the portal",
});

await agent.execute({
  instruction: `Enter "tiwa@trysplendor.com" in the email field and "mypassword" in the password field`,
  page,
});
```

The agent correctly parses the instruction and calls `fillForm` with the right values:
```
Agent calling tool: fillForm
arguments: [{"action": "type email into the One Healthcare ID or Email Address textbox", "value": "tiwa@trysplendor.com"}]
```

But the `observe()` response shows a hallucinated placeholder:
```json
{
  "elements": [{
    "elementId": "0-395",
    "description": "Textbox labeled 'One Healthcare ID or Email Address'",
    "method": "fill",
    "arguments": ["test@example.com"]
  }]
}
```

The wrong value `test@example.com` is then typed into the field.

## Root Cause

The tool receives fields with both `action` and `value`:
```typescript
inputSchema: z.object({
  fields: z.array(z.object({
    action: z.string().describe('Description of typing action'),
    value: z.string().describe("Text to type into the target")  // <-- This exists!
  }))
})
```

But the execute function only uses `action`:
```typescript
execute: async ({ fields }) => {
  // BUG: Only uses action, completely ignores value
  const instruction = `Return observation results for the following actions: ${fields
    .map((f) => f.action)
    .join(", ")}`;
  // ...
}
```

## Flaky Behavior

This manifests as **flaky behavior** because it depends on how the agent formulates the `action` string:

- **Works** when agent embeds value in action: `{"action": "type tiwa@example.com into email", "value": "tiwa@example.com"}`
- **Fails** when agent separates them: `{"action": "type into email field", "value": "tiwa@example.com"}`

## Additional Impact: Custom Tool Results Are Also Dropped

This affects values returned by custom tools. For example, when using a TOTP tool for 2FA, the agent correctly calls `fillForm` with `{ action: "type code into verification field", value: "123456" }`, but the value is dropped and the LLM hallucinates a placeholder, **breaking 2FA flows entirely**.

## Suggested Fix

Override the `arguments` in observe results with the original `value` before calling `act()`:

```typescript
for (let i = 0; i < observeResults.length; i++) {
  const res = observeResults[i];

  // FIX: Inject the actual value from fields into the observe result
  if (res.method === "fill" && fields[i]?.value) {
    res.arguments = [fields[i].value];
  }

  const actResult = await v3.act(res, actOptions);
  // ...
}
```

## Environment

- Stagehand version: 3.0.6 (also verified bug exists in 3.1.0 source)
- Using `stagehand.agent()` API

## Impact

This bug makes the `fillForm` tool effectively broken for any use case where specific values matter:
- Login forms
- Search queries
- Data entry
- 2FA/OTP flows
- Any workflow using custom tool results

shrey150:
## Summary

Fixes #1789 — supersedes #1805 with a more robust implementation.

Based on the original fix by @elliotllliu in #1805 — thank you for identifying and diagnosing this bug!

The `fillForm` tool was passing only field `action` strings to `observe()`, causing the LLM to hallucinate placeholder values (e.g. `test@example.com`) instead of using the actual `value` provided by the caller. This broke login forms, 2FA flows, search queries, and any workflow where specific values matter.

### What this PR does

- Uses a `fillIndex` counter to align `fill` results from `observe()` with caller-provided field values, correctly handling interleaved non-fill actions (e.g. click-to-focus)
- Skips extra fills when `observe()` returns more fill actions than fields provided, with a warning log
- Warns when `observe()` returns fewer fills than fields, so dropped values are visible in logs

### Changes

| File | Change |
|------|--------|
| `packages/core/lib/v3/agent/tools/fillform.ts` | Override observe result arguments with caller-provided values using fillIndex counter; skip/warn on mismatches |
| `packages/core/tests/unit/fillform-value-override.test.ts` | Unit tests covering value override, interleaved actions, empty string, non-fill passthrough, and extra fills |
| `.changeset/fix-fillform-value-hallucination.md` | Patch changeset |

## Test plan

- [x] Unit test: hallucinated values are overridden with caller-provided values (fails on main, passes here)
- [x] Unit test: non-fill methods (click) are not affected
- [x] Unit test: interleaved non-fill actions don't misalign values (fails on main, passes here)
- [x] Unit test: empty string `value: ""` correctly overrides (fails on main, passes here)
- [x] Unit test: extra fills from observe are skipped with warning
- [x] Existing `agent-execution-model` tests pass

### DIFF
--- a/packages/core/lib/v3/agent/tools/fillform.ts
+++ b/packages/core/lib/v3/agent/tools/fillform.ts
@@ -81,6 +81,13 @@ export const fillFormTool = (
             replayableActions.push(...(actResult.actions as Action[]));
           }
         }
+        if (fillIndex < fields.length) {
+          v3.logger({
+            category: "agent",
+            message: `fillForm: observe returned fewer fill actions (${fillIndex}) than provided fields (${fields.length}); ${fields.length - fillIndex} field value(s) were not applied`,
+            level: 1,
+          });
+        }
         v3.recordAgentReplayStep({
           type: "fillForm",
           fields,


shrey150:
@greptileai

shrey150:
@greptileai

shrey150:
@greptileai

shrey150:
@greptileai

packages/core/lib/v3/agent/tools/fillform.ts
```diff
@@ -56,7 +56,22 @@ export const fillFormTool = (
...
+              fillIndex++;
+            } else {
+              v3.logger({
+                category: "agent",
+                message: `fillForm: observe returned more fill actions than provided fields (${fields.length}); skipping extra fill`,
+                level: 1,
+              });
+              continue;
+            }
+          }
```
tkattkat, line 73: This assumes agent will provide the values in same order as the observe results are returned in .This likely will not be consistently true.

pirate:
@filip-michalsky did your PRs fix this already? or do we still need this one?

filip-michalsky:
@pirate @shrey150 this can be closed - should be addressed with variables on observe here https://github.com/browserbase/stagehand/pull/1808

PR #1811, opened 2026-03-11
remove unnecssary value param from fillform tool in dom mode

tkattkat:
# why

The `fillForm` tool (DOM mode) had a `value` field in its input schema that the LLM would populate, but it was never actually used during execution. The `execute` function only read `f.action` to build the observe instruction — the `value` was completely ignored. This meant the tool relied entirely on the action string (e.g. `"type foo into the email field"`) to carry both the target and the value, while `value` was dead weight.

This was also a problem for the `variables` feature: the schema told the LLM to use `%variableName%` syntax in the `value` field, but since `value` was never consumed, variable substitution through it could never work.

# what changed

- **`fillform.ts`**: Removed the `value` field from the input schema. The `action` field now explicitly guides the LLM to include the value to type *within* the action description (e.g. `"type foo@bar.com into the email field"`). When variables are available, the description shows `%variableName%` syntax with an example. This aligns the schema with how the tool actually works — `observe` parses the action string and returns `Action` objects with the correct `method`, `selector`, and `arguments`, which `act` then executes.
- **`agentSystemPrompt.ts`**: Removed the stale reference to `value` in the variable usage instructions for DOM mode (`"value/action fields"` → `"action fields"`).
- **`cache.ts`**: Updated `AgentReplayFillFormStep.fields` type to drop `value`.
- **Tests**: Removed `value` from test fixtures in `timeouts.spec.ts` and `agent-execution-model.test.ts`.

# test plan

- [ ] Run unit tests: `agent-execution-model.test.ts` passes with the updated fillForm field shape
- [ ] Run integration tests: `timeouts.spec.ts` fillForm timeout test passes
- [ ] Run agent eval with a form-filling task (e.g. `sign_in`) to verify the LLM correctly embeds values in the action string and fields get filled as expected
- [ ] Test with variables: confirm `%variableName%` in the action string gets substituted correctly (e.g. `"type %password% into the password field"`)

---
## Summary by cubic
Removed the unused `value` field from the DOM‑mode `fillForm` tool; values must now be included in the `action` string using a strict "type <value> into the <field>" pattern. Also removed the emoji from the tool description and aligned variable guidance, types, and tests.

- **Migration**
  - Update `fillForm` inputs to only use `action` that follows: "type <exact value> into the <field name> <fieldType>", e.g. `{ fields: [{ action: "type john@example.com into the email input" }] }`.
  - Use `%variableName%` inside the `action` string, e.g. `type %password% into the password input`.
  - If you reference replay types, `AgentReplayFillFormStep.fields` now only includes `{ action: string }`.

<sup>Written for commit aaf6e67d6e351a81e23aab006fb69a716835ab49. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1811">Review in cubic</a></sup>

### DIFF
--- a/packages/core/lib/v3/agent/tools/fillform.ts
+++ b/packages/core/lib/v3/agent/tools/fillform.ts
@@ -18,7 +18,7 @@ export const fillFormTool = (
 
   return tool({
     description:
-      '📝 FORM FILL - MULTI-FIELD INPUT TOOL\nFill 2+ form inputs/textareas at once. Each action MUST include the exact text to type and the target field, e.g. "type john@example.com into the email field".',
+      'FORM FILL - MULTI-FIELD INPUT TOOL\nFill 2+ form inputs/textareas at once. Each action MUST include the exact text to type and the target field, e.g. "type john@example.com into the email field".',
     inputSchema: z.object({
       fields: z
         .array(


packages/core/lib/v3/agent/tools/fillform.ts
```diff
@@ -12,22 +12,18 @@ export const fillFormTool = (
...
-    ? `Text to type into the target. Use %variableName% to substitute a variable value. Available: ${Object.keys(variables).join(", ")}`
-    : "Text to type into the target";
+  const actionDescription = hasVariables
+    ? `Must follow the pattern: "type <exact value> into the <field name> <fieldType>". Use %variableName% to substitute a variable value. Available: ${Object.keys(variables).join(", ")}. Examples: "type %email% into the email input", "type %password% into the password input"`
+    : 'Must follow the pattern: "type <exact value> into the <field name> <fieldType>". Examples: "type john@example.com into the email input", "type John into the first name input"';
 
   return tool({
-    description: `📝 FORM FILL - MULTI-FIELD INPUT TOOL\nFor any form with 2+ inputs/textareas. Faster than individual typing.`,
+    description:
+      '📝 FORM FILL - MULTI-FIELD INPUT TOOL\nFill 2+ form inputs/textareas at once. Each action MUST include the exact text to type and the target field, e.g. "type john@example.com into the email field".',
```
shrey150: Do these emojis do anything? Wouldn't they just confuse agents / be unnecessary? We should remove and standardize if so

tkattkat: lmaooo yea we dont need these

PR #1830, opened 2026-03-15
feat(cli): add --connect flag for daemon mode with existing Chrome

peytoncasper:
## Summary

- Adds `--connect <url>` CLI option that starts the daemon attached to an existing Chrome instance via CDP WebSocket URL
- The daemon persists between commands, caching accessibility tree refs from snapshots
- Unlike `--ws` (stateless, no ref caching), `--connect` gives you persistent daemon mode without launching Chrome

## Use case

Remote node management where Chrome is launched externally with custom flags (anti-bot detection, persistent profiles, specific ports) and the browse CLI needs to interact with it while preserving refs between commands.

Example:
```bash
# External Chrome is running on port 9300 with custom flags
browse --connect ws://127.0.0.1:9300/devtools/browser/abc123 open https://example.com
browse --connect ws://127.0.0.1:9300/devtools/browser/abc123 snapshot  # refs cached
browse --connect ws://127.0.0.1:9300/devtools/browser/abc123 click @0-5  # refs work!
```

## Changes

- `packages/cli/src/index.ts`: Added `--connect` option to `GlobalOpts`, program options, `runDaemon`, `ensureDaemon`, `sendCommand`, and `runCommand`
- When `--connect` is set, `ensureBrowserInitialized` uses `localBrowserLaunchOptions: { cdpUrl: connectUrl }` instead of launching Chrome
- Retry logic in `sendCommand` skips `killChromeProcesses` when using `--connect` (Chrome is externally managed)

## Test plan

- [x] `browse --connect <ws_url> open <url>` navigates successfully
- [x] `browse --connect <ws_url> snapshot -c` returns accessibility tree with refs
- [x] `browse --connect <ws_url> click @0-1` uses cached refs from previous snapshot
- [x] Daemon persists between invocations (refs survive across CLI calls)
- [ ] Existing `--ws` behavior unchanged
- [ ] Local daemon mode (no flags) unchanged

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/core/lib/v3/understudy/context.ts
+++ b/packages/core/lib/v3/understudy/context.ts
@@ -231,6 +231,22 @@ export class V3Context {
       });
     }
 
+    // Close unnavigable chrome:// tabs so the new about:blank tab becomes
+    // the only visible tab. Without this, --connect mode leaves the original
+    // chrome://newtab tab visible while the new page hides behind it.
+    try {
+      const targets = await this.conn.getTargets();
+      for (const t of targets) {
+        if (t.type === "page" && t.url?.startsWith("chrome://")) {
+          await this.conn
+            .send("Target.closeTarget", { targetId: t.targetId })
+            .catch(() => {});
+        }
+      }
+    } catch {
+      // best effort
+    }
+
     await this.newPage("about:blank");
   }
 
--- a/.changeset/clean-dogs-chew.md
+++ b/.changeset/clean-dogs-chew.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/browse-cli": patch
+---
+
+feat(cli): add --connect flag for daemon mode with existing Chrome
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -288,7 +288,11 @@ interface DaemonResponse {
 // Default viewport matching Stagehand core
 const DEFAULT_VIEWPORT = { width: 1288, height: 711 };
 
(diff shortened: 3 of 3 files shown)


miguelg719 (approved):

miguelg719:
@peytoncasper changeset for release

miguelg719 (approved):

packages/core/lib/v3/understudy/context.ts
```diff
@@ -231,6 +231,22 @@ export class V3Context {
       });
     }
 
+    // Close unnavigable chrome:// tabs so the new about:blank tab becomes
+    // the only visible tab. Without this, --connect mode leaves the original
+    // chrome://newtab tab visible while the new page hides behind it.
+    try {
+      const targets = await this.conn.getTargets();
+      for (const t of targets) {
+        if (t.type === "page" && t.url?.startsWith("chrome://")) {
```
pirate, line 240:
```suggestion
		if (t.type === "page" && t.url?.startsWith("chrome://")) {
          // TODO: consider expanding this list if needed, see chrome://chrome-urls
          // many things can cause weird unusable targets to appear at launch.
          // e.g. chrome extensions, new feature info pages, chrome error pages 
```

pirate (approved):

shrey150:
Thanks for this, @peytoncasper — and sorry it sat. Since it opened, daemon mode was reworked into the `ConnectionTarget`/`--target` model (and `--connect` is now evolving via #2059/#2060), so this approach is superseded. Closing as such — happy to revisit if I'm missing a gap here that the new model doesn't cover.

PR #1834, opened 2026-03-16
update search tool

tkattkat:
# Why

Add Browserbase Search API as the primary search tool for agents (`POST /v1/search`), leveraging the Browserbase API key users already have configured. Brave Search remains fully supported as a backwards compatible fallback for existing users with `BRAVE_API_KEY` set.

# What Changed

- **Browserbase Search tool**: New search tool powered by the Browserbase Search API, enabled via `useSearch: true` in `agent.execute()`.
- **Brave Search (backwards-compatible)**: Existing users with `BRAVE_API_KEY` in their environment continue to get the search tool automatically with no code changes required.
- **Priority**: `useSearch: true` with a Browserbase API key takes precedence. If not set, falls back to Brave if `BRAVE_API_KEY` is present.
- **Validation**: If `useSearch: true` is set without a valid Browserbase API key, a clear error is thrown at preparation time (before the agent loop starts).
- **Separate files**: `browserbaseSearch.ts` and `braveSearch.ts` keep the implementations cleanly isolated.

```typescript
const agent = stagehand.agent({ mode: "hybrid" });

  instruction: "use the search tool to find me the capital of france",
  useSearch: true,
});
```

# Test Plan

- [x] Run agent with `useSearch: true` and a valid `BROWSERBASE_API_KEY` — Browserbase search tool should return results
- [x] Run agent with `useSearch: true` and no Browserbase API key — should throw `MissingEnvironmentVariableError`

### DIFF
--- a/packages/core/lib/v3/agent/tools/browserbaseSearch.ts
+++ b/packages/core/lib/v3/agent/tools/browserbaseSearch.ts
@@ -19,6 +19,7 @@ interface BrowserbaseApiResponse {
 }
 
 async function performBrowserbaseSearch(
+  v3: V3,
   query: string,
   apiKey: string,
   numResults: number = 5,
@@ -51,7 +52,11 @@ async function performBrowserbaseSearch(
 
     return { results };
   } catch (error) {
-    console.error("Search error", error);
+    v3.logger({
+      category: "agent",
+      message: `Search error: ${(error as Error).message}`,
+      level: 0,
+    });
     return {
       results: [],
       error: `Error performing search: ${(error as Error).message}`,
@@ -79,7 +84,7 @@ export const searchTool = (v3: V3, apiKey: string) =>
         },
       });
 
-      const result = await performBrowserbaseSearch(query, apiKey);
+      const result = await performBrowserbaseSearch(v3, query, apiKey);
 
       v3.recordAgentReplayStep({
         type: "search",


packages/docs/v3/basics/agent.mdx
```diff
@@ -472,10 +472,25 @@ const result = await agent.execute({
...
+  useSearch: true,
   excludeTools: ["search"],
 });
 ```
 
+## Web Search
+
+Enable the `search` tool by setting `useSearch: true` in `agent.execute()`. This gives the agent the ability to perform web searches using the Browserbase Search API, which is useful when the agent needs to find URLs or gather information before navigating.
+
+<Note>Requires a valid Browserbase API key. Set `BROWSERBASE_API_KEY` (or `BB_API_KEY`) in your environment, or pass `apiKey` in the Stagehand constructor.</Note>
```
miguelg719: is `BB_API_KEY` valid?

packages/core/lib/v3/agent/tools/browserbaseSearch.ts
```diff
@@ -2,83 +2,64 @@ import { tool } from "ai";
...
-      }
-    }
+    const data = (await response.json()) as BrowserbaseApiResponse;
+    const results: SearchResult[] = (data?.results ?? []).map(
+      ({ title, url, publishedDate }) => ({
+        title: title ?? "",
+        url: url ?? "",
+        ...(publishedDate && { publishedDate }),
+      }),
+    );
```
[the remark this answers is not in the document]
tkattkat: instead I changed to not fallback, as the title and url will always be present ( required fields in response )

packages/core/lib/v3/agent/tools/browserbaseSearch.ts
```diff
@@ -0,0 +1,93 @@
...
+      ({ title, url, publishedDate }) => ({
+        title: title,
+        url: url,
+        ...(publishedDate && { publishedDate }),
+      }),
+    );
+
+    return { results };
+  } catch (error) {
+    console.error("Search error", error);
```
miguelg719: should this be log? @tkattkat

tkattkat: good catch yea

PR #1846, opened 2026-03-18
[STG-1607] Yield finished SSE event instead of silently dropping it

aq17:
# why

Customer reported a bug in the Python SDK (stagehand-python v3.6.0).

In src/stagehand/_streaming.py, both the sync Stream.__stream__() (line 62-63) and async AsyncStream.__stream__() (line 147-148) have:

```
python
if sse.data.startswith('{"data":{"status":"finished"'):
    break  # ← drops the event without yielding it
```

When the server sends the finished SSE event (which contains success, message, actions, usage, and messages), the SDK immediately breaks out of the loop without yielding the event to the caller. This means the final result payload is silently dropped.

# what changed

The streaming on_event config had `handle: done` for the `{"data":{"status":"finished"...}}` event. Stainless generates this as a bare break, which exits the SSE loop without yielding the event – silently dropping the final result payload (success status, message, actions, usage, and messages) from every streaming response.           

Fix – Changed to `handle: yield` so the finished event is delivered to callers before the stream terminates. This is safe because the server explicitly calls `reply.raw.end()` immediately after sending the finished event (packages/server-v3/src/lib/stream.ts), so the SSE connection closes right after the yield and the loop exits naturally on EOF – no hang risk.

# test plan

Regenerate SDKs and confirm fix

### DIFF
--- a/.changeset/fix-streaming-finished-event.md
+++ b/.changeset/fix-streaming-finished-event.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+Fix streaming finished event being silently dropped. The final SSE event containing the result payload (success status, message, actions, usage, and messages) was previously discarded instead of being yielded to the caller.


pirate:
great catch, thank you for debugging + fixing this @aq17!

PR #1849, opened 2026-03-18
Add explicit SSE event names for v3 streaming

monadoid:
## Summary
- add explicit SSE `event:` names to the shared v3 streaming helper in `packages/server-v3/src/lib/stream.ts`
- switch Stainless streaming handlers from JSON-prefix/null matching to explicit `event_type` matching in `stainless.yml`
- add an integration assertion that streamed frames now carry explicit SSE event names
- update the public v3 stream-event docs in `packages/core` and regenerate `packages/server-v3/openapi.v3.yaml`

## Why
The current SDK generation path has trouble distinguishing the terminal `finished` event from the generic catch-all when the server only emits `data:` frames. Giving each streamed frame an explicit SSE event name makes the protocol more structured and lets Stainless match on non-overlapping `event_type` handlers instead of `data_starts_with` plus a generic null-event rule.

This change is intentionally scoped to the shared v3 stream helper, which is used by all streamed v3 routes:
- `act`
- `extract`
- `observe`
- `agentExecute`
- `navigate`

## Testing
- `pnpm --dir /Users/samfinton/Documents/Programming/browserbase/stagehand --filter @browserbasehq/stagehand lint`
- `pnpm --dir /Users/samfinton/Documents/Programming/browserbase/stagehand --filter @browserbasehq/stagehand-server-v3 lint`
- `pnpm --dir /Users/samfinton/Documents/Programming/browserbase/stagehand exec prettier --check stainless.yml`

---
## Summary by cubic
Add explicit SSE event names to all v3 streaming responses so clients and SDKs can match on `event_type` instead of brittle JSON prefixes. This fixes misclassification of the `finished` event and makes handlers non-overlapping.

- **New Features**
  - `packages/server-v3/src/lib/stream.ts`: emit `event: starting|connected|running|finished|error` with JSON `data: { data, type: "system"|"log", id }`.
  - `stainless.yml`: replace `data_starts_with` with `event_type`; route `error` to `error`, others to `yield`.
  - Docs/schema: update v3 stream event docs in `packages/core` and regenerate `packages/server-v3/openapi.v3.yaml`.
  - Tests: add integration check that each SSE frame includes `event` and it matches the payload `status`.

<sup>Written for commit 0faec064c7fb77aa4cee7d8a6ee9025ec472503b. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1849">Review in cubic</a></sup>

### DIFF
--- a/packages/core/lib/v3/types/public/api.ts
+++ b/packages/core/lib/v3/types/public/api.ts
@@ -901,11 +901,9 @@ export const StreamEventLogDataSchema = z
 /**
  * SSE stream event sent during streaming responses.
  *
- * IMPORTANT: Key ordering matters for Stainless SDK generation.
- * The `data` field MUST be serialized first, with `status` as the first key within it.
- * This allows Stainless to use `data_starts_with: '{"data":{"status":"finished"'` for event handling.
- *
- * Expected serialization order: {"data":{"status":...},"type":...,"id":...}
+ * The SSE wire format includes an `event:` line that mirrors the stream status
+ * (`starting`, `connected`, `running`, `finished`, or `error`) followed by a
+ * JSON `data:` line containing the typed payload below.
  */
 export const StreamEventSchema = z
   .object({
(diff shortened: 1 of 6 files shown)


packages/core/lib/v3/types/public/api.ts
```diff
@@ -901,11 +901,9 @@ export const StreamEventLogDataSchema = z
 /**
  * SSE stream event sent during streaming responses.
  *
- * IMPORTANT: Key ordering matters for Stainless SDK generation.
```
pirate, line 904: I would keep this note, unless we are no longer matching on this pattern at all

monadoid:
Superseded by the split rollout in #1856 and #1858.

PR #1859, opened 2026-03-19
fix: fallback to zod-to-json-schema when Zod v4 toJSONSchema is missing

### LINKED ISSUE #1845: Zod v4 schema detection can fail when root zod resolves without toJSONSchema
## Summary

We are seeing a runtime failure in Stagehand `3.1.0` when `stagehand.extract()` receives a Zod v4-style schema object, but Stagehand's own root `zod` import does not expose `z.toJSONSchema()`.

The failure comes from `dist/esm/lib/v3/zodCompat.js`:

```js
import { z } from "zod";
import zodToJsonSchema from "zod-to-json-schema";
export const isZod4Schema = (schema) => typeof schema._zod !== "undefined";
export function toJsonSchema(schema) {
  if (!isZod4Schema(schema)) {
    return zodToJsonSchema(schema);
  }
  const zodWithJsonSchema = z;
  if (zodWithJsonSchema.toJSONSchema) {
    return zodWithJsonSchema.toJSONSchema(schema);
  }
  throw new Error("Zod v4 toJSONSchema method not found");
}
```

In our deployment artifact, this throws:

```text
Error: Zod v4 toJSONSchema method not found
```

## Why this is surprising

Stagehand declares support for:

- `zod ^3.25.76 || ^4.2.0`

But there is a runtime shape where:

- the schema object looks like Zod v4 (`_zod` exists)
- Stagehand's own root `zod` import does **not** expose `toJSONSchema`

That means Stagehand detects the schema as v4, but then assumes its own imported `z` also has the v4 root API surface.

## Reproduction shape

This is enough to reproduce the failure in our built output:

```js
import { z } from './node_modules/zod/v4/index.js';
import { toJsonSchema } from './node_modules/@browserbasehq/stagehand/dist/esm/lib/v3/zodCompat.js';

toJsonSchema(z.object({ ok: z.boolean() }));
```

This throws:

```text
Error: Zod v4 toJSONSchema method not found
```

## What seems to be happening

In our case, the deploy artifact ends up with a transitional `zod@3.25.76` package at the root, where:

- `import { z } from "zod"` does not provide `toJSONSchema`
- `import { z } from "zod/v4"` does provide `toJSONSchema`

So the mismatch is:

- Stagehand gets a v4 schema object
- Stagehand imports root `zod`
- root `zod` is not the v4 root surface Stagehand expects

## Related upstream packaging issue

We also opened an issue with the Mastra team because the mismatch appears during their generated build output packaging:

- mastra-ai/mastra#14460
- https://github.com/mastra-ai/mastra/issues/14460

That said, this still seems like Stagehand could be more defensive here, since downstream build systems may legitimately hand it a v4 schema object while root `zod` resolves differently.

## Questions

1. Is this a known issue?
2. Do you recommend always using `zod/v3` schemas with `stagehand.extract()` as the safest option in production?
3. Would you consider making the v4 path more defensive, for example by:
   - importing from `zod/v4` when available, or
   - falling back more gracefully instead of throwing when `schema` is v4-shaped but root `z` lacks `toJSONSchema`?
4. Is there a canonical way you want consumers to avoid this class of drift in bundled/deployed environments?

Happy to provide more detail if useful.

ghost:
## Summary

Fixes a runtime crash in `stagehand.extract()` when a Zod v4 schema is provided but the root `zod` import doesn't expose `z.toJSONSchema()`.

## Problem

```
Error: Zod v4 toJSONSchema method not found
```

This happens when the bundler resolves the root `zod` import to a build that doesn't include `toJSONSchema` (e.g., certain Zod v4 builds or bundler configurations that tree-shake it away).

## Fix

Instead of throwing, fall back to the `zod-to-json-schema` library which is already imported and can handle both v3 and v4 schemas. This makes the code resilient to different Zod resolution scenarios.

## Before / After

```typescript
// Before: throws if z.toJSONSchema not available
throw new Error("Zod v4 toJSONSchema method not found");

// After: falls back to the library that's already imported
return zodToJsonSchema(schema);
```

Fixes #1845

---
## Summary by cubic
Prevent crash in `stagehand.extract()` when Zod v4’s `toJSONSchema()` is unavailable by checking at runtime and falling back to `zod-to-json-schema`. Keeps schema conversion working for Zod v3 and v4 across bundlers. Fixes #1845.

<sup>Written for commit 516678cf886c129f74aae534bd079b95f73b8e67. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1859">Review in cubic</a></sup>

### DIFF
--- a/packages/core/lib/v3/zodCompat.ts
+++ b/packages/core/lib/v3/zodCompat.ts
@@ -1,4 +1,4 @@
-import { z, toJSONSchema as zodV4ToJsonSchema } from "zod";
+import { z } from "zod";
 import type {
   ZodObject as Zod4Object,
   ZodRawShape as Zod4RawShape,
@@ -35,8 +35,17 @@ export function toJsonSchema(schema: StagehandZodSchema): JsonSchemaDocument {
     return zodToJsonSchema(schema);
   }
 
-  // Use the named import directly, which is resolved at module load time
-  // and not susceptible to tree-shaking or bundler resolution issues
-  // that can strip z.toJSONSchema from the namespace object.
-  return zodV4ToJsonSchema(schema as Zod4TypeAny) as JsonSchemaDocument;
+  // For v4 schemas, use built-in z.toJSONSchema() method
+  const zodWithJsonSchema = z as typeof z & {
+    toJSONSchema?: (schema: Zod4TypeAny) => JsonSchemaDocument;
+  };
+
+  if (zodWithJsonSchema.toJSONSchema) {
+    return zodWithJsonSchema.toJSONSchema(schema as Zod4TypeAny);
+  }
+
+  // Fallback to zod-to-json-schema when the native method isn't available.
+  // This can happen when the root zod import resolves to a build that
+  // doesn't expose toJSONSchema (e.g., some bundler configurations).
+  return zodToJsonSchema(schema as Parameters<typeof zodToJsonSchema>[0]);
 }


ghost:
The bot flagged that zod-to-json-schema and z.toJSONSchema may produce slightly different output for v4 schemas. That's true in theory, but the alternative is throwing an error that crashes the user's app entirely. A working fallback with minor schema differences is strictly better than no fallback at all.

In practice, zod-to-json-schema has supported Zod v4 since its 3.24.x release, so the output should be equivalent for the schemas Stagehand uses (objects with primitives, arrays, enums).

PR #1861, opened 2026-03-19
[v4]: add index to selector schemas

seanmcguire12:
# why
- selectors can resolve to multiple nodes
- we need to give the user freedom to optionally provide an `index`, indicating the desired element if multiple are found
# what changed
- added an optional `index` field to `XPathSelectorSchema`, `CssSelectorSchema` & `TextSelectorSchema`
- note: `CoordinateSelectorSchema` remains unchanged, since coordinates are inherently specific

### selector schema before:
Screenshot 2026-03-19 at 4 31 50 PM

### selector schema after: 
Screenshot 2026-03-19 at 4 33 32 PM

# test plan
- added new test case in `packages/server-v4/src/schemas/v4/page.ts`

---
## Summary by cubic
Adds an optional non-negative idx to XPath, CSS, and text selectors so callers can target a specific element when a selector matches multiple nodes. Updates the v4 schema and OpenAPI spec without breaking existing clients.

- **New Features**
  - Added optional non-negative `idx` to `XPathSelectorSchema`, `CssSelectorSchema`, and `TextSelectorSchema`.
  - Left `CoordinateSelectorSchema` unchanged.
  - Updated `openapi.v4.yaml` with the new `idx` field and examples.
  - Expanded integration tests to cover `css`, `xpath`, and `text` selectors with `idx`.

<sup>Written for commit 0c25f7442cfd0bdc11fb6da761bddf4fe1375699. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1861">Review in cubic</a></sup>

### DIFF
--- a/packages/server-v4/src/schemas/v4/page.ts
+++ b/packages/server-v4/src/schemas/v4/page.ts
@@ -112,7 +112,7 @@ export const PageActionStatusSchema = z
 export const XPathSelectorSchema = z
   .object({
     xpath: z.string().min(1).meta({ example: "//button[text()='Submit']" }),
-    index: z.number().int().nonnegative().optional().meta({ example: 0 }),
+    idx: z.number().int().nonnegative().optional().meta({ example: 0 }),
   })
   .strict()
   .meta({ id: "XPathSelector" });
@@ -120,7 +120,7 @@ export const XPathSelectorSchema = z
 export const CssSelectorSchema = z
   .object({
     css: z.string().min(1).meta({ example: ".btn-submit" }),
-    index: z.number().int().nonnegative().optional().meta({ example: 0 }),
+    idx: z.number().int().nonnegative().optional().meta({ example: 0 }),
   })
   .strict()
   .meta({ id: "CssSelector" });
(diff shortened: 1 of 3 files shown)


packages/server-v4/src/schemas/v4/page.ts
```diff
@@ -112,20 +112,23 @@ export const PageActionStatusSchema = z
 export const XPathSelectorSchema = z
   .object({
     xpath: z.string().min(1).meta({ example: "//button[text()='Submit']" }),
+    index: z.number().int().nonnegative().optional().meta({ example: 0 }),
```
pirate: lets call it `idx` for brevity? `idx` in my experience always means "small integer value describing which element to choose", but `index` is more ambiguous because it could mean "use some search index to get this value"

monadoid (approved):

packages/server-v4/src/schemas/v4/page.ts
seanmcguire12: i like

miguelg719 (approved):

PR #1877, opened 2026-03-23
[v4]: add `/elementInfo` route stub

seanmcguire12:
# why
- to add a single `/elementInfo` route which encapsulates multiple `locator` functions
# what changed
- added an `/elementInfo` route, which accepts a `Selector`, and returns the following info:
  - `count`
  - `isVisible`
  - `isChecked`
  - `inputValue`
  - `textContent`
  - `innerHTML`
  - `innerText`
  - `centroid`

### screenshot:
Screenshot 2026-03-23 at 5 10 55 PM

# test plan
- added a test which verifies that the `/elementInfo` route accepts a request

---
## Summary by cubic
Adds the v4 `page.elementInfo` endpoint (`POST /v4/page/elementInfo`) to fetch common element details in one call. The handler is a stub that returns defaulted fields to unblock client work.

- **New Features**
  - Implemented `page.elementInfo` route that accepts a `Selector` and optional `fields` filter; returns: `count`, `selector`, `tagName`, `backendNodeId`, `visibility`, `domRects`, `content`, and optional `inputInfo`, `ariaInfo`, `attributes`, `styles`.
  - Wired the route into v4 page routes, added request/action/result/response schemas to OpenAPI, and added an integration test posting an XPath selector to assert a successful action.

<sup>Written for commit ebdb216d1ff3aa6b7c890a3d42047463acc98c14. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1877">Review in cubic</a></sup>

### DIFF
--- a/packages/server-v4/src/schemas/v4/page.ts
+++ b/packages/server-v4/src/schemas/v4/page.ts
@@ -488,8 +488,21 @@ export const PageCloseParamsSchema = PageWithPageIdSchema.meta({
   id: "PageCloseParams",
 });
 
+export const ElementInfoFieldSchema = z
+  .enum([
+    "visibility",
+    "domRects",
+    "content",
+    "inputInfo",
+    "ariaInfo",
+    "attributes",
+    "styles",
+  ])
+  .meta({ id: "ElementInfoField" });
+
 export const PageElementInfoParamsSchema = PageWithPageIdSchema.extend({
-  selector: ElementSelectorSchema,
+  selector: SelectorSchema,
+  fields: z.array(ElementInfoFieldSchema).optional(),
 })
   .strict()
   .meta({ id: "PageElementInfoParams" });
@@ -862,28 +875,90 @@ export const PageCloseResultSchema = z
   .strict()
   .meta({ id: "PageCloseResult" });
 
+export const ElementInfoVisibilitySchema = z
+  .object({
+    isInViewport: z.boolean(),
+    isOccluded: z.boolean(),
+  })
+  .strict()
+  .meta({ id: "ElementInfoVisibility" });
+
+export const ElementInfoDomRectSchema = z
+  .object({
+    x: z.number(),
+    y: z.number(),
+    width: z.number(),
+    height: z.number(),
+  })
+  .strict()
+  .meta({ id: "ElementInfoDomRect" });
+
+export const ElementInfoDomRectsSchema = z
+  .object({
+    rects: z.array(ElementInfoDomRectSchema),
+  })
+  .strict()
+  .meta({ id: "ElementInfoDomRects" });
+
+export const ElementInfoContentSchema = z
+  .object({
+    innerText: z.string(),
+    textContent: z.string(),
+    innerHTML: z.string(),
+  })
+  .strict()
+  .meta({ id: "ElementInfoContent" });
+
+export const ElementInfoInputInfoSchema = z
+  .object({
+    value: z.string(),
+    isChecked: z.boolean(),
+  })
+  .strict()
+  .meta({ id: "ElementInfoInputInfo" });
+
+export const ElementInfoAriaInfoSchema = z
+  .object({
+    role: z.string(),
+    attributes: z.record(z.string(), z.string()),
+  })
+  .strict()
+  .meta({ id: "ElementInfoAriaInfo" });
+
+export const ElementInfoAttributesSchema = z
+  .object({
+    values: z.record(z.string(), z.string()),
+  })
+  .strict()
+  .meta({ id: "ElementInfoAttributes" });
+
+export const ElementInfoStylesSchema = z
+  .object({
+    computed: z.record(z.string(), z.string()),
+  })
+  .strict()
+  .meta({ id: "ElementInfoStyles" });
+
 export const PageElementInfoResultSchema = z
   .object({
     count: z.number().int().nonnegative().default(0),
-    isVisible: z.boolean().default(false),
-    isChecked: z.boolean().default(false),
-    inputValue: z.string().default(""),
-    textContent: z.string().default(""),
-    innerHTML: z.string().default(""),
-    innerText: z.string().default(""),
-    centroid: z
-      .object({
-        x: z.number(),
-        y: z.number(),
-      })
-      .strict()
-      .default({ x: 0, y: 0 }),
-    selector: ResultSelectorSchema.optional(),
-    elementType: z.string().optional(),
-    backendNodeId: z.number().int().optional(),
-    ariaRole: z.string().optional(),
-    ariaAttributes: z.record(z.string(), z.string()).optional(),
-    attrs: z.record(z.string(), z.string()).optional(),
+    selector: ResultSelectorSchema.default({}),
+    tagName: z.string(),
+    backendNodeId: z.number().int().nonnegative(),
+    visibility: ElementInfoVisibilitySchema.default({
+      isInViewport: false,
+      isOccluded: false,
+    }),
+    domRects: ElementInfoDomRectsSchema.default({ rects: [] }),
+    content: ElementInfoContentSchema.default({
+      innerText: "",
+      textContent: "",
+      innerHTML: "",
+    }),
+    inputInfo: ElementInfoInputInfoSchema.optional(),
+    ariaInfo: ElementInfoAriaInfoSchema.optional(),
+    attributes: ElementInfoAttributesSchema.optional(),
+    styles: ElementInfoStylesSchema.optional(),
   })
   .strict()
   .meta({ id: "PageElementInfoResult" });
@@ -1494,6 +1569,15 @@ export const pageOpenApiComponents = {
     PageEvaluateResponse: PageEvaluateResponseSchema,
     PageSendCDPResponse: PageSendCDPResponseSchema,
     PageCloseResponse: PageCloseResponseSchema,
+    ElementInfoField: ElementInfoFieldSchema,
+    ElementInfoVisibility: ElementInfoVisibilitySchema,
+    ElementInfoDomRect: ElementInfoDomRectSchema,
+    ElementInfoDomRects: ElementInfoDomRectsSchema,
+    ElementInfoContent: ElementInfoContentSchema,
+    ElementInfoInputInfo: ElementInfoInputInfoSchema,
+    ElementInfoAriaInfo: ElementInfoAriaInfoSchema,
+    ElementInfoAttributes: ElementInfoAttributesSchema,
+    ElementInfoStyles: ElementInfoStylesSchema,
     PageElementInfoResult: PageElementInfoResultSchema,
     PageElementInfoResponse: PageElementInfoResponseSchema,
     PageActionIdParams: PageActionIdParamsSchema,
--- a/packages/server-v4/openapi.v4.yaml
+++ b/packages/server-v4/openapi.v4.yaml
@@ -2856,7 +2856,11 @@ components:
         pageId:
           $ref: "#/components/schemas/PageId"
         selector:
-          $ref: "#/components/schemas/ElementSelector"
+          $ref: "#/components/schemas/Selector"
+        fields:
+          type: array
+          items:
+            $ref: "#/components/schemas/ElementInfoField"
       required:
         - selector
       additionalProperties: false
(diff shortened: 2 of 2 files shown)


pirate:
I would also add:

- `selector`: fully hydrated selector `{xpath,css,x,y,text,elementType,backendNodeId}`
- `ariaRole`
- `ariaAttributes`
- `attrs` (element attributes like `{name,id,data-*,href,style,...}`)

packages/server-v4/src/schemas/v4/page.ts
```diff
@@ -850,6 +862,32 @@ export const PageCloseResultSchema = z
   .strict()
   .meta({ id: "PageCloseResult" });
 
+export const PageElementInfoResultSchema = z
+  .object({
+    count: z.number().int().nonnegative().default(0),
+    isVisible: z.boolean().default(false),
+    isChecked: z.boolean().default(false),
+    inputValue: z.string().default(""),
```
pirate: these are redundant with `attrs` below, I dont think we need to send both.

If you do want to break it out then I think it should just be `value` and `value=1` or `value=checked` for checkboxes that are checked

packages/server-v4/src/schemas/v4/page.ts
```diff
@@ -850,6 +862,32 @@ export const PageCloseResultSchema = z
   .strict()
   .meta({ id: "PageCloseResult" });
 
+export const PageElementInfoResultSchema = z
+  .object({
+    count: z.number().int().nonnegative().default(0),
+    isVisible: z.boolean().default(false),
```
pirate:
```suggestion
    isInViewport: z.boolean().default(false),   // e.g. scrolled out of view
    isHidden: z.boolean().default(false),       // e.g. display: hidden | none
    isOccluded: z.boolean().default(false),     // e.g. elem is behind a modal, not clickable until modal is dismissed
```
`isVisible` is misleadingly simple, we should split this up so consumers can actually determine what is needed in order to click an element (closing some occluding element, scrolling, un-hiding, etc. all require different steps)

packages/server-v4/src/schemas/v4/page.ts
```diff
@@ -850,6 +862,32 @@ export const PageCloseResultSchema = z
...
+        y: z.number(),
+      })
+      .strict()
+      .default({ x: 0, y: 0 }),
+    selector: ResultSelectorSchema.optional(),
+    elementType: z.string().optional(),
+    backendNodeId: z.number().int().optional(),
+    ariaRole: z.string().optional(),
+    ariaAttributes: z.record(z.string(), z.string()).optional(),
+    attrs: z.record(z.string(), z.string()).optional(),
```
pirate:
```suggestion
    attributes: z.record(z.string(), z.string()),
    style: z.record(z.string(), z.unknown()),
```

We should add computed CSS styles so that the agent can check for things like `display: none`, `display: hidden`, `opacity: 0`, `color: green`, etc.).

Also `attributes` should never be none, we should just return `{}` to avoid more unions. Try to match the naming and shapes that the DOM `Element` API uses to minimize confusion for agents and users: https://developer.mozilla.org/en-US/docs/Web/API/Element

`style` should be the same object shape as the DOM API: [`CSSStyleDeclaration`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration)

packages/server-v4/src/schemas/v4/page.ts
```diff
@@ -850,6 +862,32 @@ export const PageCloseResultSchema = z
...
+    centroid: z
+      .object({
+        x: z.number(),
+        y: z.number(),
+      })
+      .strict()
+      .default({ x: 0, y: 0 }),
+    selector: ResultSelectorSchema.optional(),
+    elementType: z.string().optional(),
+    backendNodeId: z.number().int().optional(),
```
pirate:
```suggestion
    backendNodeId: z.bigint().positive(),
```
I think this is always available, no? Also it should never be negative.

packages/server-v4/src/schemas/v4/page.ts
```diff
@@ -850,6 +862,32 @@ export const PageCloseResultSchema = z
   .strict()
   .meta({ id: "PageCloseResult" });
 
+export const PageElementInfoResultSchema = z
+  .object({
+    count: z.number().int().nonnegative().default(0),
+    isVisible: z.boolean().default(false),
+    isChecked: z.boolean().default(false),
+    inputValue: z.string().default(""),
+    textContent: z.string().default(""),
```
pirate:
```suggestion
```
This is redundant with `selector.text` and `innerText` I think

packages/server-v4/src/schemas/v4/page.ts
```diff
@@ -850,6 +862,32 @@ export const PageCloseResultSchema = z
...
+    innerHTML: z.string().default(""),
+    innerText: z.string().default(""),
+    centroid: z
+      .object({
+        x: z.number(),
+        y: z.number(),
+      })
+      .strict()
+      .default({ x: 0, y: 0 }),
```
pirate:
```suggestion
    domrects: z.array(z
      .object({
        x: z.number(),
        y: z.number(),
        top: z.number().optional(),
        bottom: z.number().optional(),
        right: z.number().optional(),
        left: z.number().optional(),
      })
      .strict())
```
This should be an array of [`DOMRect`](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect)s matching the shape returned by [`elem.getClientRects()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getClientRects). I has to be an array because many inline elements like `<span>`, `<p>` etc. often wrap across multiple lines and have multiple bounding boxes. Centroid-based clicking is useless for most inline elements subject to wrapping unless you provide the full list of bounding rects.

packages/server-v4/src/schemas/v4/page.ts
```diff
@@ -850,6 +862,32 @@ export const PageCloseResultSchema = z
...
+    selector: ResultSelectorSchema.optional(),
+    elementType: z.string().optional(),
```
pirate:
```suggestion
    elementType: z.string(),
```

packages/server-v4/src/schemas/v4/page.ts
```diff
@@ -850,6 +862,32 @@ export const PageCloseResultSchema = z
   .strict()
   .meta({ id: "PageCloseResult" });
 
+export const PageElementInfoResultSchema = z
+  .object({
+    count: z.number().int().nonnegative().default(0),
```
pirate, line 944: If we dont return the whole list of matching elements here then we need to a separate endpoint to get the whole list, or at least a param like `idx` that we can pass to "get the second matching element".

packages/server-v4/src/schemas/v4/page.ts
```diff
@@ -850,6 +862,32 @@ export const PageCloseResultSchema = z
...
+        x: z.number(),
+        y: z.number(),
+      })
+      .strict()
+      .default({ x: 0, y: 0 }),
+    selector: ResultSelectorSchema.optional(),
+    elementType: z.string().optional(),
+    backendNodeId: z.number().int().optional(),
+    ariaRole: z.string().optional(),
+    ariaAttributes: z.record(z.string(), z.string()).optional(),
```
pirate:
```suggestion
    aria: z.record(z.string(), z.unknown()),
```
Just return `{}` if empty, not undefined. Also all aria-related stuff should be under a single key imo, there are lots of attrs we might want to add in the future and it's easier if they're nested under one type instead of flat at the top level:

- [`aria-autocomplete`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-autocomplete)
- [`aria-checked`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-checked)
- [`aria-disabled`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-disabled)
- [`aria-errormessage`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-errormessage)
- [`aria-expanded`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-expanded)
- [`aria-haspopup`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-haspopup)
- [`aria-hidden`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-hidden)
- [`aria-invalid`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-invalid)
- [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label)
- [`aria-level`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-level)
- [`aria-modal`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-modal)
- [`aria-multiline`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-multiline)
- [`aria-multiselectable`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-multiselectable)
- [`aria-orientation`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-orientation)
- [`aria-placeholder`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-placeholder)
- [`aria-pressed`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-pressed)
- [`aria-readonly`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-readonly)
- [`aria-required`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-required)
- [`aria-selected`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-selected)
- [`aria-sort`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-sort)
- [`aria-valuemax`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-valuemax)
- [`aria-valuemin`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-valuemin)
- [`aria-valuenow`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-valuenow)
- [`aria-valuetext`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-valuetext)
- and lots more...

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes

packages/server-v4/src/schemas/v4/page.ts
```diff
@@ -850,6 +862,32 @@ export const PageCloseResultSchema = z
...
+      .object({
+        x: z.number(),
+        y: z.number(),
+      })
+      .strict()
+      .default({ x: 0, y: 0 }),
+    selector: ResultSelectorSchema.optional(),
+    elementType: z.string().optional(),
+    backendNodeId: z.number().int().optional(),
+    ariaRole: z.string().optional(),
```
pirate:
```suggestion
```
I think this should be under a single `aria: {...}` shape, see below.

packages/server-v4/src/schemas/v4/page.ts
seanmcguire12: i think this route should only return details for a single element: 
- if there are multiple matches, this route will just take the first match
- eg, if the user wants info on the third match, they must provide a selector that is specific enough to reach the third match

pirate: thats really difficult to do if you dont know the nesting level at which they are peers, e.g.

`/div/div/div`  to get the second match thee is no way to know which is needed:

- `/div/div/div[2]`
- `/div/div[2]/div`
- `/div[2]/div/div`

packages/server-v4/src/schemas/v4/page.ts
pirate: BU tried a bunch of really intricate heuristics to try and solve this problem because it was a big issue with links that wrap across lines. e.g. try 2px in from the upper left corner of the first bounding box, then try 4px in from the bottom right corner of the last bounding box, etc. based on various conditions

a naiive centroid click would land in the middle of the space between the two bounding boxex, which is actually outside both of them, so the click does nothing.

seanmcguire12: yeah, `centroid` is a bit too leaky. can make it an array with multiple rects. do you think we need top, right, bottom, & left though? or should we just make it x, y, width, height ?

packages/server-v4/src/schemas/v4/page.ts
seanmcguire12: innerText & textContent are different https://stackoverflow.com/questions/35213147/difference-between-textcontent-vs-innertext

packages/server-v4/src/schemas/v4/page.ts
seanmcguire12: - `isChecked` needs to look at the `checked` property on the `HTMLInputElement`, so it wouldn't show in `attrs`
- `inputValue` also reads from the `value` property, so it also wouldn't show in `attrs`

packages/server-v4/src/schemas/v4/page.ts
seanmcguire12: thats why the `ElementSelectorSchema` provides an `idx` param

packages/server-v4/src/schemas/v4/page.ts
seanmcguire12: down to have multiple visibility related fields. should we have them all nested in an outer `visibility` object though?

packages/server-v4/src/schemas/v4/page.ts
seanmcguire12: put these into a nested obj

packages/server-v4/src/schemas/v4/page.ts
seanmcguire12: put these into a nested obj

packages/server-v4/src/schemas/v4/page.ts
seanmcguire12: add computed styles object, rename attrs -> attributes

packages/server-v4/src/schemas/v4/page.ts
seanmcguire12: innerText could be huge, maybe add depth arg or think about putting this into a separate route.

packages/server-v4/src/schemas/v4/page.ts
pirate: just do the same shape as [DOMRect](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect), the native browser type for this

PR #1777, opened 2026-03-03
fix: add variable substitution to keys tool

trillville:
- [x] Check the [documentation](https://docs.stagehand.dev/) for relevant information
- [x] Search existing [issues](https://github.com/browserbase/stagehand/issues) to avoid duplicates

Fixes #1776

### Fix

- Accept `variables` parameter in `keysTool` (matching `typeTool`)
- Call `substituteVariables()` before `page.type()` in the `method === "type"` branch
- Pass `variables` to `keysTool` in `createAgentTools`
- Return original token in result to avoid exposing sensitive values to LLM

---
## Summary by cubic
Fixes variable substitution in the keys tool when method="type" so %variableName% tokens are resolved before typing. Aligns behavior with the type tool and prevents secret leakage in tool outputs.

- **Bug Fixes**
  - Substitute %variableName% in keys(method="type") via substituteVariables before page.type().
  - Accept and pass variables to keysTool; update input schema to list available variables so the agent knows %var% syntax.
  - Return the original token string in the tool result to avoid exposing substituted secrets.

<sup>Written for commit 9c30883ba601e39d6e17ddfe91dfd72f00d4f7cd. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1777">Review in cubic</a></sup>

PR #1978, opened 2026-04-08
fix: add variable substitution to keys step during cache replay

### LINKED ISSUE #1776: keys tool does not perform variable substitution
### Environment Information

**Stagehand:**

- Language/SDK: TypeScript
- Stagehand version: 3.1.0

**AI Provider:**

- Provider: Anthropic
- Model: claude-sonnet-4-6

### Issue Description

The keys tool with method: "type" does not call substituteVariables() before typing, so %variableName% tokens are typed literally instead of being resolved. The type tool correctly calls 
https://github.com/browserbase/stagehand/blob/main/lib/v3/agent/tools/type.ts#L28, but the keys tool sends the raw value directly to https://github.com/browserbase/stagehand/blob/main/lib/v3/agent/tools/keys.ts#L33.           

Additionally, variables is passed to type, act, fillForm, and fillFormVision in
https://github.com/browserbase/stagehand/blob/main/lib/v3/agent/tools/index.ts, but not to keys.

This is somewhat easy to trigger in practice: when the agent needs to clear an existing field value before typing a new one, it naturally uses the keys tool for the whole flow (Ctrl+A → Delete → type), and the final type step skips substitution.

### Steps to Reproduce

1. Create an agent with variables (e.g. { token: "my-secret-value" })
2. Give it a task that involves clearing an existing input field and replacing it with %token%
3. The agent uses keys for the flow: keys(press, Ctrl+A) → keys(press, Delete) → keys(type, %token%)
4. The literal string %token% is typed instead of my-secret-value

### Suggested fix 
In keys.ts, the keysTool factory should accept variables (like typeTool does) and substitute before typing

a7med3liamin:
- [x] Check the [documentation](https://docs.stagehand.dev/) for relevant information
- [x] Search existing [issues](https://github.com/browserbase/stagehand/issues) to avoid duplicates

Fixes #1776

## Problem

The `keys` tool has no variable substitution in either the live execution or cache replay paths. When the agent uses `%variableName%` tokens with the keys tool, the literal token string gets typed instead of the resolved value.

## Fix

This PR combines two fixes into one:

### 1. Live execution (original fix by @trillville from #1777)
- Accept `variables` parameter in `keysTool` (matching `typeTool`)
- Call `substituteVariables()` before `page.type()` in the `method === "type"` branch
- Pass `variables` to `keysTool` in `createAgentTools`
- Update schema description to advertise available variables to the LLM
- Return original token in result to avoid exposing sensitive values to LLM

### 2. Cache replay (new fix)
- Import `substituteVariables` in `AgentCache.ts`
- Pass `variables` through to `replayAgentKeysStep`
- Call `substituteVariables(text, variables)` before `page.type()` in the replay path

Without fix #2, cached `keys` steps with `method="type"` replay by typing literal `%variableName%` tokens even when variables are provided, since `replayAgentKeysStep` had no access to the variables map.

## Credit

The live execution fix (part 1) is from @trillville's work in #1777/#1813. We merged it here with the cache replay fix per @pirate's request to consolidate into a single PR.

### DIFF
--- a/.changeset/keys-cache-replay-variables.md
+++ b/.changeset/keys-cache-replay-variables.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+Add variable substitution to the keys tool in both live execution and cache replay paths. When keys steps with `method="type"` contain `%variableName%` tokens, they are now resolved against the provided variables. This brings the keys tool to parity with the type tool's variable handling.
--- a/packages/core/lib/v3/agent/tools/index.ts
+++ b/packages/core/lib/v3/agent/tools/index.ts
@@ -168,7 +168,7 @@ export function createAgentTools(v3: V3, options?: V3AgentToolOptions) {
     fillForm: fillFormTool(v3, executionModel, variables, toolTimeout),
     fillFormVision: fillFormVisionTool(v3, provider, variables),
     goto: gotoTool(v3),
-    keys: keysTool(v3),
+    keys: keysTool(v3, variables),
     navback: navBackTool(v3),
     screenshot: screenshotTool(v3),
     scroll: mode === "hybrid" ? scrollVisionTool(v3, provider) : scrollTool(v3),
--- a/packages/core/lib/v3/agent/tools/keys.ts
+++ b/packages/core/lib/v3/agent/tools/keys.ts
@@ -1,9 +1,16 @@
 import { tool } from "ai";
 import { z } from "zod";
 import type { V3 } from "../../v3.js";
+import type { Variables } from "../../types/public/agent.js";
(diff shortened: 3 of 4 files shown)


pirate (approved):

pirate:
sorry for the trouble can you actually merge your two PRs into a single unified pr / this PR. then I will claim it.

Our github workflow is a little clunky right now because we have to "claim" external contributor PRs in order to run CI, so it's easier to do fewer PRs that have more changes all in one.

a7med3liamin:
No worries at all @pirate.

Done, merged @trillville's live execution fix from #1777 into this PR alongside the cache replay fix.

pirate (approved):

PR #2027, opened 2026-04-22
Remove error thrown on no action taken

miguelg719:
# why

After adding regex schema validation for element ids in `act`, the models were left with no choice to say 'no candidate found/no action to be taken' since an empty string on element id would fail validation. This PR corrects the regressed behavior

# what changed
`act()` no longer throws when the model can't find a target. Callers get the structured `success: false` result, matching the behavior `observe()` already has ([]) when nothing matches.   
- `packages/core/lib/inference.ts` — wrap the act fields (elementId, description, method, arguments, twoStep) under action: {...}.nullable(). When the model returns action: null, inference returns element: undefined and the existing handler branch returns { success: false, actions: [], message: "Failed to perform act: No action found" }.
- Lift `twoStep` out of the nullable action object and back to the top level of the act schema — small models (gpt-4.1-nano) flatten booleans out of anyOf variants during structured output, causing spurious AI_NoObjectGeneratedError  on otherwise-valid responses.
- `packages/core/lib/prompt.ts` — update buildActSystemPrompt, buildActPrompt, and buildStepTwoPrompt to instruct the model to set action: null when no element matches, with an explicit ban on empty strings / placeholder values.
- `packages/core/lib/v3/handlers/actHandler.ts` — no logic change; the existing no-action branch is now reachable.    

Act will return the following on these cases:
Screenshot 2026-04-22 at 1 19 33 PM

# test plan

---
## Summary by cubic
Stops `act` from throwing when no element matches by making the response `action` nullable and updating prompts/parsing to treat “no action” as valid. Fixes the regression where empty element IDs failed validation. Aligns with Linear STG-1849.

- **Bug Fixes**
  - Changed `act` schema to `{ action: { elementId, description, method, arguments } | null, twoStep: boolean }` with the element ID regex; `twoStep` is top‑level and defaults to `false`.
  - Updated prompts to set `action: null` when no element matches and to not fabricate elements or use empty/placeholder values.
  - Adjusted parsing to handle `action === null` and return `element: undefined`, letting the existing no‑action path return `success: false`.
  - Updated integration tests and test utils to use the new `action` shape and top‑level `twoStep`.

<sup>Written for commit 1cb51fa47ffefe7b7b931eb011ca6c468aa0bf97. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/2027">Review in cubic</a></sup>

### DIFF
--- a/packages/core/lib/inference.ts
+++ b/packages/core/lib/inference.ts
@@ -440,12 +440,12 @@ export async function act({
               "the arguments to pass to the method. For example, for a click, the arguments are empty, but for a fill, the arguments are the value to fill in.",
             ),
         ),
-        twoStep: z.boolean(),
       })
       .nullable()
       .describe(
         "The element to act on. Return null if no element on the page matches the instruction — do NOT fabricate or guess an element, and never emit empty strings or placeholder values.",
       ),
+    twoStep: z.boolean(),
   });
 
   type ActResponse = z.infer<typeof actSchema>;
@@ -535,6 +535,6 @@ export async function act({
     reasoning_tokens: reasoningTokens,
     cached_input_tokens: cachedInputTokens,
     inference_time_ms: usageTimeMs,
-    twoStep: actData.action?.twoStep ?? false,
+    twoStep: actData.twoStep,
   };
 }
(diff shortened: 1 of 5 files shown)


miguelg719:
@cubic-dev-ai

monadoid (approved):

PR #2062, opened 2026-04-28
fix(core): gracefully handle runtimes that omit AsyncLocalStorage.enterWith() (fixes #2055)

### LINKED ISSUE #2055: FlowLogger breaks on Cloudflare Workers: enterWith() is not supported
## Problem

Starting in v3.2.0, Stagehand crashes immediately on Cloudflare Workers with:

```
asyncLocalStorage.enterWith() is not implemented
```

This happens during `stagehand.init()` before any browser interaction occurs.

## Root cause

`FlowLogger.init()` uses `AsyncLocalStorage.enterWith()` to set session-wide logging context:

```ts
// packages/core/lib/v3/flowlogger/FlowLogger.ts
static init(sessionId, eventBus) {
  const ctx = { sessionId, eventBus, parentEvents: [] };
  loggerContext.enterWith(ctx);  // <-- this line
  return ctx;
}
```

Cloudflare Workers [intentionally omits `enterWith()`](https://developers.cloudflare.com/workers/runtime-apis/nodejs/asynclocalstorage/) from their `AsyncLocalStorage` implementation because it mutates context for the entire remaining async chain, which is unsafe across concurrent requests. They support `run()`, `getStore()`, `exit()`, `bind()`, and `snapshot()` — but not `enterWith()`.

This means any Stagehand user running on Workers (which includes browser automation via the [Browser Rendering](https://developers.cloudflare.com/browser-rendering/) binding) hits this immediately on v3.2.0+.

## Suggested fix

Replace the `enterWith()` call with `run()`. Since `FlowLogger.init()` is called at session start and the context just needs to be available for the duration of the Stagehand session, wrapping the session lifecycle in a `loggerContext.run(ctx, callback)` would work and is compatible with all runtimes.

Alternatively, a try/catch around the `enterWith()` call with a fallback would at least allow Stagehand to function (without flow logging) on runtimes that don't support it.

## Workaround

We're currently shimming `enterWith()` on the prototype before importing Stagehand:

```ts
import { AsyncLocalStorage } from 'node:async_hooks'

if (typeof AsyncLocalStorage.prototype.enterWith !== 'function') {
  const fallbackStores = new WeakMap()
  const runDepth = new WeakMap()
  const originalGetStore = AsyncLocalStorage.prototype.getStore
  const originalRun = AsyncLocalStorage.prototype.run

  AsyncLocalStorage.prototype.enterWith = function (store) {
    fallbackStores.set(this, store)
  }

  AsyncLocalStorage.prototype.run = function (store, callback, ...args) {
    runDepth.set(this, (runDepth.get(this) ?? 0) + 1)
    try {
      return originalRun.call(this, store, callback, ...args)
    } finally {
      const d = (runDepth.get(this) ?? 1) - 1
      if (d <= 0) runDepth.delete(this)
      else runDepth.set(this, d)
    }
  }

  AsyncLocalStorage.prototype.getStore = function () {
    const active = originalGetStore.call(this)
    if ((runDepth.get(this) ?? 0) > 0) return active
    return active !== undefined ? active : fallbackStores.get(this)
  }
}
```

This works but it's a monkey-patch we'd rather not carry.

## Environment

- Stagehand: 3.2.1
- Runtime: Cloudflare Workers (Durable Objects) with `nodejs_compat` flag
- `compatibility_date`: 2026-04-01
- Using `WorkersAIClient` (custom `LLMClient`) + `@cloudflare/playwright`

## Versions affected

- v3.1.x and earlier: **not affected** (FlowLogger didn't exist)
- v3.2.0+: **affected** (FlowLogger introduced in PR #1836)

octo-patch:
Fixes #2055

## Problem

Since v3.2.0, `FlowLogger.init()` calls `AsyncLocalStorage.enterWith()` to set the session-wide logging context. Cloudflare Workers intentionally omits `enterWith()` from their `AsyncLocalStorage` implementation — calling it throws `asyncLocalStorage.enterWith() is not implemented`, which crashes Stagehand immediately during `init()` before any browser interaction occurs.

## Solution

Wrap the `enterWith()` call in a `try/catch`. If the runtime does not support `enterWith()`, we fall through silently. The `FlowLoggerContext` is still created and returned, so it can be stored by the caller (`this.flowLoggerContext` in v3.ts). All other ALS usage inside FlowLogger already uses `loggerContext.run()` (in `runWithLogging()`, `withContext()`, `logCdpEvent()`, etc.), which is supported everywhere — so the logging pipeline continues to work through the explicit-context path on runtimes without `enterWith()`.

## Testing

Added a unit test that simulates a runtime without `enterWith()` (by temporarily setting `AsyncLocalStorage.prototype.enterWith = undefined`) and asserts that `FlowLogger.init()` completes without throwing and returns a valid context.

```
✓ FlowLogger.init() does not throw when enterWith() is not implemented (e.g. Cloudflare Workers)
```

All existing flowlogger unit tests continue to pass.

---
## Summary by cubic
Fixes #2055. Prevents crashes on runtimes that omit `AsyncLocalStorage.enterWith()` (e.g., Cloudflare Workers) by safely initializing `FlowLogger`. Logging continues via `loggerContext.run()` paths.

- **Bug Fixes**
  - Wrap `enterWith()` in try/catch in `FlowLogger.init()`; fall through when missing.
  - Add a unit test that simulates missing `enterWith()` and verifies init succeeds and returns a valid `FlowLoggerContext`.

<sup>Written for commit 7211ad50befdcd3d9fbc1709b2ee460e703d7090. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/2062?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/core/lib/v3/flowlogger/FlowLogger.ts
+++ b/packages/core/lib/v3/flowlogger/FlowLogger.ts
@@ -413,7 +413,15 @@ export class FlowLogger {
       parentEvents: [],
     };
 
-    loggerContext.enterWith(ctx);
+    try {
+      loggerContext.enterWith(ctx);
+    } catch {
+      // Some runtimes (e.g. Cloudflare Workers) do not implement enterWith()
+      // because it mutates context across concurrent requests, which is unsafe
+      // in that environment. Fall through: the context is still returned and
+      // callers that need ALS can use loggerContext.run() via runWithLogging()
+      // or withContext().
+    }
     return ctx;
   }
 
--- a/packages/core/tests/unit/flowlogger-capturing-llm.test.ts
+++ b/packages/core/tests/unit/flowlogger-capturing-llm.test.ts
@@ -1,5 +1,7 @@
+import { AsyncLocalStorage } from "node:async_hooks";
 import { describe, expect, it } from "vitest";
 import { FlowLogger } from "../../lib/v3/flowlogger/FlowLogger.js";
+import { EventEmitterWithWildcardSupport } from "../../lib/v3/flowlogger/EventEmitter.js";
 
 describe("flow logger llm logging", () => {
   it("no-ops direct llm logging calls when no flow context is active", () => {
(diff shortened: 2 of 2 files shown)


dontfeedthecode:
Can this be reviewed? It's a blocker for me.

PR #2068, opened 2026-04-29
remove hardcoded agent temp

filip-michalsky:
# why                                                                
  Reasoning models (e.g. `openai/gpt-5-mini`) emit `AI SDK Warning: The "temperature" setting is not supported by this model` whenever the v3 agent
   runs against them, because `V3AgentHandler` hardcoded `temperature: 1` on every `generateText`/`streamText` call. There was no constructor or   
  agent-config option to override it, so users running reasoning models with `experimental: true` had no way to silence the warning.
                                                                                                                                                   
  # what changed                                                       
  - Removed the hardcoded `temperature: 1` from both the non-streaming and streaming agent paths in
  `packages/core/lib/v3/handlers/v3AgentHandler.ts`.                                                                                               
  - Agent calls now omit `temperature` entirely, so the AI SDK no longer flags it as user-set on models that don't accept it. Models that do accept
   temperature still default to the AI SDK provider's own default.                                                                                 
  - Added unit coverage in `packages/core/tests/unit/agent-temperature.test.ts` asserting neither code path passes a `temperature` to the LLM
  client.                                                                                                                                          
  - Added a patch changeset (`.changeset/remove-agent-temperature-default.md`).
                                                                                                                                                   
  # test plan
  - [ ] `pnpm --filter @browserbasehq/stagehand test agent-temperature` — new unit tests pass                                                      
  - [ ] Existing v3 agent unit suite stays green                                                                                                   
  - [ ] Manual: `agent.execute(...)` against `openai/gpt-5-mini` with `experimental: true` — confirm no `"temperature" setting is not supported`
  warning in stderr                                                                                                                                
  - [ ] Manual: same call against a non-reasoning model (e.g. `openai/gpt-4.1-mini`) — confirm agent still completes a multi-step task end-to-end
  - [ ] Manual: hybrid-mode agent run on a reasoning model — verify streaming path is also clean


filip-michalsky:
That failing auto-captcha test is going to get removed.

miguelg719 (approved):

PR #2084, opened 2026-05-06
[feat]: add `ignoreSelectors` to `extract()`

seanmcguire12:
# why
- to allow users to specify selectors whose content should be omitted from the LLM request inside of `extract()`
# what changed
- added `ignoreSelectors` to `extract()` so callers can exclude parts of the page from the extraction snapshot
- `packages/core/lib/v3/understudy/a11y/snapshot/capture.ts`:
  - added `resolveIgnoreSelectorRoots()` to resolve ignore selectors into `frameId + backendNodeId` 
  - added `buildFrameExclusionIntervals()` to turn those roots into per-frame subtree ranges
  - updated `collectPerFrameMaps()` / `tryScopedSnapshot()` to skip ignored nodes when building snapshot data
- `packages/core/lib/v3/understudy/a11y/snapshot/domTree.ts`:
  - updated `buildSessionDomIndex()` to record dfs entry/exit positions for each backend node so ignored subtrees can be filtered by interval membership instead of having to re-traverse the tree
- `packages/core/lib/v3/understudy/a11y/snapshot/a11yTree.ts`: 
  - updated `a11yForFrame()` to accept an ignore predicate and remove ignored nodes from the outline (the stringified tree) and `urlMap`
- `packages/core/lib/v3/handlers/extractHandler.ts` and `packages/core/lib/v3/v3.ts`: 
  - passed `ignoreSelectors` down through the extract path into `captureHybridSnapshot()`
- kept `xpathMap` values based on the original dom structure and only skipped ignored entries, so surviving selectors do not get renumbered

# test plan
- updated `packages/core/tests/unit/public-api/public-types.test.ts` to cover the new `ignoreSelectors` option on `extract()`
- updated `packages/core/tests/unit/snapshot-a11y-resolvers.test.ts` to cover ignored backend nodes being removed from the a11y outline and `urlMap`, and to cover scoped snapshots dropping ignored xpath entries
- updated `packages/core/tests/unit/snapshot-capture-orchestration.test.ts` to cover the new capture flow, including filtered per-frame maps and merged snapshot output when `ignoreSelectors` is used

---
## Summary by cubic
Adds `ignoreSelectors` to `extract()` so callers can exclude elements and their subtrees from snapshots and the LLM input. Supports CSS and XPath (with iframe hops) across shadow DOM and iframes, without renumbering existing XPaths; also adds a fast path for scoped snapshots and excludes child frame subtrees when an iframe host is ignored.

- **New Features**
  - `extract({ ignoreSelectors: string[] })`: drops matching nodes and descendants from DOM maps, a11y outline, and `urlMap`; `xpathMap` keeps original values but omits ignored entries.
  - Applies to full-page and scoped snapshots; filters resolved per frame, excludes child frames when their iframe host is ignored, and enforces DFS enter/exit intervals for speed.
  - A11y tree accepts an ignore predicate and filters before building the outline and links.
  - Scoped snapshots return early when only `selector` is provided (no `ignoreSelectors`) for faster capture.
  - Plumbed through `extractHandler`, `v3`, and `captureHybridSnapshot()`; public types and OpenAPI updated in `@browserbasehq/stagehand`; unit tests added.

<sup>Written for commit 0ad2f67d8fe81d15945599c60351ac45f66f1277. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/blue-suns-post.md
+++ b/.changeset/blue-suns-post.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": minor
+---
+
+add ignoreSelectors param to extract()
--- a/packages/core/lib/v3/handlers/extractHandler.ts
+++ b/packages/core/lib/v3/handlers/extractHandler.ts
@@ -109,7 +109,15 @@ export class ExtractHandler {
   async extract<T extends StagehandZodSchema>(
     params: ExtractHandlerParams<T>,
   ): Promise<InferStagehandSchema<T> | { pageText: string }> {
-    const { instruction, schema, page, selector, timeout, model } = params;
+    const {
+      instruction,
+      schema,
+      page,
+      selector,
+      ignoreSelectors,
+      timeout,
+      model,
+    } = params;
 
     const llmClient = this.resolveLlmClient(model);
 
@@ -126,6 +134,7 @@ export class ExtractHandler {
       const snap = await captureHybridSnapshot(page, {
         experimental: this.experimental,
         focusSelector: focusSelector || undefined,
+        ignoreSelectors,
       });
       ensureTimeRemaining();
 
@@ -147,6 +156,7 @@ export class ExtractHandler {
     const { combinedTree, combinedUrlMap } = await captureHybridSnapshot(page, {
       experimental: this.experimental,
       focusSelector: focusSelector,
+      ignoreSelectors,
     });
 
     v3Logger({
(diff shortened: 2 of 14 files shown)


seanmcguire12:
@cubic-dev-ai

seanmcguire12:
@cubic-dev-ai

miguelg719 (approved):

PR #2130, opened 2026-05-15
feat(verifier): add verifier evaluator shell and types

miguelg719:
# Why

The verifier pipeline needs a stable public contract before the judgment engine lands. This PR introduces the verifier-facing API and types while keeping runtime behavior reviewable and compatible with the legacy backend.

# What Changed

- Added public verifier types for trajectories, task specs, rubrics, criteria, verdicts, findings, and evidence.
- Consolidated the verifier-facing type surface through `packages/core/lib/v3/verifier/types.ts` with public re-exports only from `verifier/index.ts`.
- Removed implementation-level proxy type barrels, including the unused `verifier.ts` file and `trajectory.ts` type re-exports.
- Kept public `RubricCriterion.earnedPoints` numeric-only; serialized empty-string values are normalized away at the IO boundary.
- Constrained trajectory asset paths so `screenshotPath` cannot escape the trajectory directory during offline loading.
- Added `V3Evaluator.verify()` and `V3Evaluator.generateRubric()` facade methods.
- Added legacy compatibility behavior for `verify()` over saved trajectories.
- Exported verifier types and helpers from the public v3 entrypoint.
- Replaced the temporary stub reason with the stable `stub-verifier` reason.
- Added public API and trajectory utility coverage for the new verifier surface.

# Tests

- `pnpm --filter @browserbasehq/stagehand run build:esm`
- `pnpm --filter @browserbasehq/stagehand run test:core -- packages/core/dist/esm/tests/unit/public-api/v3-core.test.js`
- `pnpm --filter @browserbasehq/stagehand run test:core -- packages/core/dist/esm/tests/unit/verifier-trajectory.test.js`
- `git diff --check`


packages/core/lib/v3/verifier/trajectory.ts
```diff
@@ -0,0 +1,332 @@
...
+  earnedPoints?: number | string;
+}
+
+/** A rubric — list of criteria for a task. */
+export interface Rubric {
+  items: RubricCriterion[];
+}
+
+/**
+ * FARA/upstream rubric item shape as stored in datasets and prompt responses.
```
miguelg719:
```suggestion
 * Rubric item shape as stored in datasets and prompt responses.
```

pirate (approved):

PR #2131, opened 2026-05-15
feat(verifier): record agent trajectories

miguelg719:
# Why

The new verifier needs richer evidence than a final screenshot, especially for DOM and Hybrid agent modes where the important facts often live in tool returns, ARIA snapshots, and per-step observations. This PR adds trajectory recording without changing the verifier judgment engine.

# What Changed

- Added typed agent events for screenshot, step-finished, step-observed, and final-answer events.
- Added listener-gated post-step probes for screenshots and ARIA trees.
- Attached the settled post-turn probe to every tool call in a DOM/Hybrid turn.
- Added CUA step evidence pairing and final answer capture.
- Added `TrajectoryRecorder` persistence and a smoke script for trajectory shape and disk layout.

# Tests

- `pnpm --filter @browserbasehq/stagehand-evals run typecheck`
- `node --import tsx packages/evals/scripts/verify-trajectory-recorder.ts`
- `git diff --check`


packages/core/lib/v3/agent/AnthropicCUAClient.ts
```diff
@@ -902,6 +902,10 @@ export class AnthropicCUAClient extends AgentClient {
             ...input,
           };
         } else if (action === "triple_click" || action === "tripleClick") {
+          // Anthropic's computer_20250124 tool emits `triple_click` with
+          // `coordinate: [x, y]`. Without this branch the snake_case name +
+          // raw coordinate array fall through to the generic `else` and
+          // executeAction logs "Unknown action type: triple_click".
```
miguelg719:
```suggestion
```

miguelg719:
```suggestion
```

.changeset/verifier-trajectory-events.md
```diff
@@ -0,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
```
monadoid, line 2: Do we need these three different patch files or just 1?

packages/core/lib/v3/handlers/v3AgentHandler.ts
```diff
@@ -292,8 +311,90 @@ export class V3AgentHandler {
             action.timestamp = Date.now();
             state.actions.push(action);
           }
+
+          const stepIndex = stepCounter++;
+          stepIndicesInTurn.push(stepIndex);
+          const toolOk =
```
monadoid: I'm not sure if this covers all possible cases - seems like the toolresult shape can sometimes look like `toolResult.output.success` - so this might need to cover that? Might be useful to transform / standardize tool result shapes first to avoid this logic where we have to check a bunch of different fields, but that might be more complicated.

packages/evals/framework/trajectoryRecorder.ts
```diff
@@ -0,0 +1,333 @@
...
+  private ensurePartial(stepIndex: number): Partial<PartialStep> {
+    let p = this.partialSteps.get(stepIndex);
+    if (!p) {
+      p = { index: stepIndex };
+      this.partialSteps.set(stepIndex, p);
+    }
+    return p;
+  }
+
+  private assembleSteps(): TrajectoryStep[] {
```
monadoid: Could we see a scenario where a model sees a screenshot and then does a final answer immediately after (i.e. no action) - in this case, I think the screenshot event might not make it into the trajectory?

.changeset/verifier-trajectory-events.md
miguelg719: I can consolidate at the end if we want to have all be 1 patch

packages/core/lib/v3/agent/utils/toolOutputEvidence.ts
```diff
@@ -0,0 +1,25 @@
...
+  const isError =
+    toolResult &&
+    typeof toolResult === "object" &&
+    "isError" in toolResult &&
+    Boolean((toolResult as { isError?: unknown }).isError);
+
+  return {
+    ok: error === undefined && !isError,
+    result: toolResult,
+    error,
```
seanmcguire12, line 74: this means tools that return `{ success: false }` without an `error` field are marked as `ok: true` . is this intentional?

packages/evals/framework/trajectoryRecorder.ts
```diff
@@ -0,0 +1,283 @@
...
+      }
+      out.push({
+        index: i,
+        actionName: p.actionName,
+        actionArgs: p.actionArgs ?? {},
+        reasoning: p.reasoning ?? "",
+        agentEvidence: p.agentEvidence ?? { modalities: [] },
+        probeEvidence: p.probeEvidence ?? {},
+        toolOutput: p.toolOutput,
+        startedAt: this.startedAt,
```
seanmcguire12: is `startedAt` supposed to mean the start of a step? or the start of an entire run? right now its only recording the start of an entire run

packages/core/lib/v3/handlers/v3AgentHandler.ts
```diff
@@ -292,8 +312,44 @@ export class V3AgentHandler {
             action.timestamp = Date.now();
             state.actions.push(action);
           }
+
+          const stepIndex = stepCounter++;
+          stepIndicesInTurn.push(stepIndex);
+          await evidenceCallback?.({
```
seanmcguire12, line 323: should this be wrapped in try/catch? or do we want to treat as hard failure?

packages/core/lib/v3/handlers/v3CuaAgentHandler.ts
```diff
@@ -767,6 +791,96 @@ export class V3CuaAgentHandler {
...
+        ? action.pageUrl
+        : (this.cuaEvidenceSteps.latestScreenshotUrl ?? "");
+    try {
+      pageUrl = (await this.v3.context.awaitActivePage()).url();
+    } catch {
+      // Keep the best pre-action URL fallback.
+    }
+    const { stepIndex, replayScreenshot } = this.cuaEvidenceSteps.pairAction();
+    if (replayScreenshot) {
+      await this.evidenceCallback?.(replayScreenshot);
```
seanmcguire12: a bit confused on what the policy is for when something should be treated as fatal vs non fatal. eg, `emitCuaScreenshotNonFatal()` is called and errors are caught, but replaying it with `await this.evidenceCallback?.(replayScreenshot)` , we let the error propogate?

packages/core/lib/v3/handlers/v3CuaAgentHandler.ts
```diff
@@ -145,10 +154,13 @@ export class V3CuaAgentHandler {
             [action],
           );
         } else {
-          await this.executeAction(action);
+          executionResult = await this.executeAction(action);
         }
 
         action.timestamp = Date.now();
+        if (shouldLog) {
+          await this.emitCuaActionStep(action, executionResult);
```
seanmcguire12: right now this is getting called for all non-screenshot actions. do we need to do this even if the `onEvidence` is not supplied?

miguelg719:
@cubic-dev-ai review

packages/core/lib/v3/handlers/v3CuaAgentHandler.ts
miguelg719: addressed in [3dfa861](https://github.com/browserbase/stagehand/pull/2131/commits/3dfa861419c405f8c0b7bb8d03425c8a6ef5be8c)

packages/core/lib/v3/handlers/v3AgentHandler.ts
miguelg719: addressed in [3dfa861](https://github.com/browserbase/stagehand/pull/2131/commits/3dfa861419c405f8c0b7bb8d03425c8a6ef5be8c)

packages/core/lib/v3/agent/utils/toolOutputEvidence.ts
miguelg719: addressed in [3dfa861](https://github.com/browserbase/stagehand/pull/2131/commits/3dfa861419c405f8c0b7bb8d03425c8a6ef5be8c)

packages/evals/framework/trajectoryRecorder.ts
miguelg719: we now emit a final observation on every agent loop [2418db3](https://github.com/browserbase/stagehand/pull/2131/commits/2418db391edd981c3adbdd57e97ad941001c3cc2)

packages/evals/framework/trajectoryRecorder.ts
miguelg719: addressed in [2418db3](https://github.com/browserbase/stagehand/pull/2131/commits/2418db391edd981c3adbdd57e97ad941001c3cc2)

packages/core/lib/v3/handlers/v3CuaAgentHandler.ts
miguelg719: addressed

seanmcguire12 (approved):

PR #2132, opened 2026-05-15
feat(verifier): normalize canonical evidence

miguelg719:
# Why

The verifier needs a mode-neutral evidence layer before scoring. DOM/Hybrid tasks can only be judged correctly if text, JSON, and tool-output evidence are considered alongside screenshots and ARIA evidence, while core installs should not pick up image-processing runtime dependencies.

# What Changed

- Added canonical screenshot loading, deduplication, resizing, and step-index mapping.
- Added canonical text evidence for ARIA snippets, agent text, agent JSON, and native tool output.
- Added combined chronological canonical evidence collection.
- Moved canonical evidence and evidence-load result types into `verifier/types.ts`; implementation files import those types directly.
- Kept image reduction behind a dynamic `sharp` import so core remains dependency-light; `sharp` is owned by evals where verifier tooling runs.
- Added a smoke check proving `probe-aria`, `agent-text`, `agent-json`, and `tool-output` evidence are collected.
- Removed upstream verifier references from comments.

# Tests

- `pnpm -w exec prettier --check packages/core/lib/v3/verifier packages/core/tests/unit/verifier-failure-step-parser.test.ts packages/core/tests/unit/verifier-trajectory.test.ts packages/evals/cli.ts packages/evals/framework packages/evals/tests/framework/rubricCache.test.ts packages/evals/tests/framework/verifierAdapter.test.ts packages/evals/tests/tui/commandTree.test.ts packages/evals/tui packages/evals/tasks/bench/agent`
- Canonical evidence smoke: `collectCanonicalEvidence` produced `probe-aria,agent-text,agent-json,tool-output`
- `git diff --check`


miguelg719:
@cubic-dev-ai

packages/core/lib/v3/verifier/evidence.ts
```diff
@@ -0,0 +1,555 @@
...
+
+  // Collect raw frames in chronological order. Per step we take tier-1
+  // agent-mirrored screenshots first (the exact bytes a CUA provider saw, plus
+  // any inline act() screenshots lifted into image modalities), then the
+  // tier-2 post-action probe. Buffers are populated either live or after
+  // loadTrajectoryFromDisk(); absent/empty buffers are skipped. Dedup collapses
+  // near-identical frames downstream, so including both tiers is safe.
+  const rawFrames: Array<{
+    bytes: Buffer;
+    originalStepIndex: number;
```
monadoid, line 141: I think (could be wrong) that `originalStepIndex` and `trajectoryStepPosition` are the same thing and one can be removed?

packages/core/lib/v3/verifier/evidence.ts
```diff
@@ -0,0 +1,555 @@
...
+  for (const shot of loaded.screenshots) {
+    pending.push({ kind: "image", shot });
+  }
+
+  const seenText = new Set<string>();
+  const addTextEvidence = (
+    stepPos: number,
+    stepIdx: number,
+    source: CanonicalTextEvidence["source"],
+    raw: unknown,
```
monadoid: Optional nit but I'd make this more strongly typed so then you can remove the whole checking typeof raw thing and the safeStringifyEvidence helper as well - could use a zod object or raw could be `string | undefined` (and maybe renamed)

packages/core/lib/v3/verifier/evidence.ts
miguelg719: Good catch — they were always assigned the same value (they could differ back when `TrajectoryStep.index` existed, but that field was removed in the verifier-03 follow-ups). Removed `trajectoryStepPosition` in 7ca3221a6; `originalStepIndex` stays since the engine prompts cite it.

packages/core/lib/v3/verifier/evidence.ts
miguelg719: Tightened in 7ca3221a6: `addTextEvidence` now takes `string | undefined` and the typeof check is gone. Kept `safeStringifyEvidence` at the two call sites where the payload is natively `unknown` (`agent-json` modality content and `toolOutput.result` are arbitrary extract/tool returns), since the stringification has to live somewhere — but it's explicit at the source now instead of hidden in the helper.

packages/core/lib/v3/verifier/evidence.ts
```diff
@@ -0,0 +1,553 @@
...
+    pending.push({ kind: "image", shot });
+  }
+
+  const seenText = new Set<string>();
+  const addTextEvidence = (
+    stepIdx: number,
+    source: CanonicalTextEvidence["source"],
+    text: string | undefined,
+  ) => {
+    if (!text || text.length === 0) return;
```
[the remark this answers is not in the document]
miguelg719: Hardened — the guard is now `typeof text !== "string"` so a malformed non-string field in on-disk trajectory JSON is skipped instead of crashing collection. The `agent-json`/`tool-output` paths still stringify their natively-unknown payloads via `safeStringifyEvidence`.

pirate (approved):

PR #2149, opened 2026-05-20
Add `screenshot` option to Extract

miguelg719:
# why
Some pages do not properly encapsulate the necessary information required to extract the content requested by the user, and can benefit from a hybrid approach with vision.

# what changed
- New `extract({ options: { screenshot: true } })` flag. Captures the current viewport as PNG and sends it as an `image_url` part alongside the a11y tree text in the extraction LLM call.
- Default is `false` — existing callers see no behavior change.
- Gated to AI SDK clients; throws `StagehandInvalidArgumentError` otherwise (validated in both the handler and `inference.extract`, with the handler check running before the screenshot is taken).
- Screenshot capture is bracketed by `ensureTimeRemaining()` so it respects the extract timeout.
- Wired end-to-end: public `ExtractOptions` type, `ExtractOptionsSchema`, OpenAPI `ExtractOptions`, `StagehandAPIClient.extract` wire payload, `V3.extract` → `ExtractHandler` → `inference.extract` → `buildExtract{System,User}Prompt`.

# test plan
- `extract-screenshot.test.ts` (new) — multimodal message shape on happy path; rejection for non-AI SDK clients.
- `timeout-handlers.test.ts` — handler captures viewport when enabled, skips capture by default, rejects screenshot + non-AI SDK client without touching `page.screenshot` or the snapshot pipeline.
- `api-client-serialization.test.ts` (renamed from `api-client-observe-variables.test.ts`) — wire payload preserves `options.screenshot`.

---
## Summary by cubic
Adds a screenshot option to `extract()` that sends the current viewport screenshot with the accessibility tree to improve results on visually-driven pages.

- New Features
  - `extract({ options: { screenshot: boolean } })` captures a PNG of the current viewport and includes it with the accessibility tree; defaults to false; only supported with `aisdk` clients (throws `StagehandInvalidArgumentError` otherwise).
  - Prompts are consolidated and multimodal when enabled: system and user prompts mention the screenshot and using it with the accessibility tree; attaches an `image_url` data URI; logs when a screenshot is used.
  - Public API, types, and OpenAPI updated (`ExtractOptions.screenshot`); the API client forwards the flag; tests cover prompt shape, viewport capture, default-off behavior, timeout guards, client-type validation, and serialization.

<sup>Written for commit d0c75002884eb42fcebb132c0d333259e8d9efbf. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/2149?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.changeset/extract-screenshot-option.md
+++ b/.changeset/extract-screenshot-option.md
@@ -2,4 +2,4 @@
 "@browserbasehq/stagehand": minor
 ---
 
-Add a `screenshot` option to `extract()` that sends the current viewport screenshot with the accessibility tree for AI SDK LLM extraction.
+Added a `screenshot` option to `extract()` that sends the current viewport screenshot with the a11y tree for extraction.
--- a/.changeset/extract-screenshot-option.md
+++ b/.changeset/extract-screenshot-option.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": minor
+---
+
+Add a `screenshot` option to `extract()` that sends the current viewport screenshot with the accessibility tree for AI SDK LLM extraction.
--- a/packages/core/lib/prompt.ts
+++ b/packages/core/lib/prompt.ts
@@ -21,6 +21,7 @@ ${userProvidedInstructions}`;
 export function buildExtractSystemPrompt(
   isUsingPrintExtractedDataTool: boolean = false,
   userProvidedInstructions?: string,
+  includeScreenshot: boolean = false,
 ): ChatMessage {
   const baseContent = `You are extracting content on behalf of a user.
   If a user asks you to extract a 'list' of information, or 'all' information, 
@@ -30,7 +31,9 @@ export function buildExtractSystemPrompt(
 1. An instruction
 2. `;
 
-  const contentDetail = `A list of DOM elements to extract from.`;
+  const contentDetail = includeScreenshot
+    ? `A hierarchical accessibility tree and a screenshot of the current viewport to extract from. Use them together to extract content from the page.`
+    : `A list of DOM elements to extract from.`;
 
   const instructions = `
 Print the exact text from the DOM elements with all symbols, characters, and endlines as is.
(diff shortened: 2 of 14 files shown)


.changeset/extract-screenshot-option.md
```diff
@@ -0,0 +1,5 @@
+---
+"@browserbasehq/stagehand": minor
+---
+
+Add a `screenshot` option to `extract()` that sends the current viewport screenshot with the accessibility tree for AI SDK LLM extraction.
```
miguelg719:
```suggestion
Added a `screenshot` option to `extract()` that sends the current viewport screenshot with the a11y tree for extraction.
```

packages/core/tests/unit/api-client-observe-variables.test.ts
```diff
@@ -52,6 +52,41 @@ describe("StagehandAPIClient variable serialization", () => {
...
+      client as unknown as {
+        execute: typeof executeMock;
+      }
+    ).execute = executeMock;
+
+    await client.extract({
+      instruction: "extract the title",
+      options: {
+        screenshot: true,
+        serverCache: true,
```
miguelg719:
```suggestion
```

packages/core/lib/prompt.ts
```diff
@@ -30,7 +31,9 @@ export function buildExtractSystemPrompt(
 1. An instruction
 2. `;
 
-  const contentDetail = `A list of DOM elements to extract from.`;
+  const contentDetail = includeScreenshot
+    ? `A hierarchical accessibility tree and a screenshot of the current viewport to extract from. Use them together to extract content from the page.`
+    : `A list of DOM elements to extract from.`;
```
miguelg719:
```suggestion
    ? `A list of DOM elements to extract from and a screenshot of the current viewport to extract from. Use them together to extract content from the page.`
    : `A list of DOM elements to extract from.`;
```

pirate (approved):
looks good, just make sure we dont emit full base64 to any DB/logs

packages/core/tests/unit/extract-screenshot.test.ts
```diff
@@ -0,0 +1,99 @@
...
+        usage: {
+          prompt_tokens: 4,
+          completion_tokens: 2,
+          total_tokens: 6,
+        },
+      });
+
+    const llmClient = {
+      type: "aisdk",
+      modelName: "gpt-4o",
```
pirate, line 33: maybe worth using one of the env vars we have to select preferred vision model?

miguelg719: no need

PR #2157, opened 2026-05-22
feat(verifier): add verifier evaluator shell and types

miguelg719:
Replacement for #2130, which was merged into the PR1 branch instead of main. This branch is rebased onto current main and contains the PR2 verifier evaluator shell/type changes.

---
## Summary by cubic
Adds a verifier evaluator shell with public trajectory/rubric/result types and utilities, plus a `V3Evaluator.verify(trajectory)` facade that uses the legacy backend by default without breaking existing flows.

- **New Features**
  - New public types under `v3/verifier`: `Trajectory`, `Rubric`, `EvaluationResult`, `Verifier`, and more.
  - Utilities: `normalizeRubric`, `loadTrajectoryFromDisk` (rehydrates screenshots and image modalities), `nextResultFilename`.
  - `V3Evaluator` supports `verify()` and `generateRubric()`. Backend selectable via `STAGEHAND_EVALUATOR_BACKEND` (`legacy` default; `verifier` is stubbed for future). Legacy path maps trajectory screenshots/final answer/reasoning to the old evaluator and returns an `EvaluationResult`.
  - Re-exported on the V3 public API: `Stagehand.loadTrajectoryFromDisk`, `Stagehand.nextResultFilename`, `Stagehand.normalizeRubric`.

- **Bug Fixes**
  - Normalizes dataset rubrics to public camelCase and validates required fields; strips serialized `earned_points` noise.
  - Secures `screenshotPath` to stay within the trajectory directory and decodes on-disk `bytesBase64` to `Buffer`.
  - Derives the verification task from the `Trajectory` to prevent mismatched task specs and simplify `verify()`.

<sup>Written for commit 3d8c324046d7c12b9a436dea4985841312c906dc. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/2157?utm_source=github">Review in cubic</a></sup>


packages/core/lib/v3/verifier/trajectory.ts
```diff
@@ -0,0 +1,189 @@
...
+ * @param dir absolute or cwd-relative path to a `<run-id>/<task-id>/` directory.
+ */
+export async function loadTrajectoryFromDisk(dir: string): Promise<Trajectory> {
+  const fs = await import("node:fs/promises");
+  const path = await import("node:path");
+  const trajectoryDir = path.resolve(dir);
+
+  const trajectoryPath = path.join(trajectoryDir, "trajectory.json");
+  const raw = await fs.readFile(trajectoryPath, "utf8");
+  const parsed = JSON.parse(raw) as Trajectory & {
```
monadoid, line 107: This could be made more typesafe at runtime if we used zod at the parsing boundary, like: `TrajectorySchema.safeParse(JSON.parse(raw))` 

and then you could z.infer to still have a Trajectory type (the array of trajectory steps could all be part of the zod schema too)

but might be a nit!

packages/core/lib/v3/verifier/types.ts
```diff
@@ -0,0 +1,306 @@
...
+  /** Debugging summary from the active evaluator backend. */
+  rawSteps?: VerifierRawSteps;
+}
+
+/**
+ * Verifier interface. Implementations consume a Trajectory + TaskSpec and
+ * return an EvaluationResult — they MUST NOT touch a live browser.
+ */
+export interface Verifier {
+  verify(trajectory: Trajectory, taskSpec: TaskSpec): Promise<EvaluationResult>;
```
monadoid: Since the trajectory already contains a taskspec within it, can we remove the additional taskSpec that gets passed in here? This would make it simpler, but maybe i'm missing something.

monadoid (approved):

packages/core/lib/v3/verifier/trajectory.ts
```diff
@@ -0,0 +1,189 @@
...
+type RawRubric = {
+  items?: unknown;
+};
+
+/**
+ * Convert dataset or generated rubric JSON into the public Stagehand shape.
+ * Snake-case dataset fields are accepted here so serialized quirks do not leak
+ * into the canonical rubric type.
+ */
+export function normalizeRubric(rubric: unknown): Rubric | undefined {
```
monadoid, line 26: Also here - a rubric could just be a zod object + a z.infer type, and then this logic could be built into the zod object (via superrefine or otherwise), and it might be a bit simpler, but optional nit!

packages/core/lib/v3/verifier/trajectory.ts
miguelg719: going to add an extra pr at the end to parse as much as possible, this one is tricky because we use downstream some precomputed rubrics from `webtailbench` specifically that don't match our schema (snake)

packages/core/lib/v3/verifier/types.ts
miguelg719: good catch 🤝

monadoid (approved):

PR #2167, opened 2026-05-26
STG-1746 Add SEA --version flag

monadoid:
# why

The SEA binary needs a lightweight way to report the commit and timestamp it was built from.

# what changed

Added a `--version` path that prints build metadata baked into the SEA binary during build.

### DIFF
--- a/.changeset/true-pillows-prove.md
+++ b/.changeset/true-pillows-prove.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand-server-v3": patch
+---
+
+Add SEA binary --version build metadata output.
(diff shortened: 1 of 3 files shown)


pirate (approved):

turbo.json
```diff
@@ -256,6 +257,7 @@
     },
     "build:sea:cjs": {
       "dependsOn": ["^build:cjs"],
+      "cache": false,
```
pirate: This will slow down builds quite a bit, no? Isn't the whole point of turbo to track input/output hashes well enough that this should be cachable, because this is the most expensive step in the chain.

pirate (approved):

PR #2171, opened 2026-05-29
[feat]: add `clipboard` API

seanmcguire12:
# why

- to give users a first class primitive for basic clipboard functionality: cut, copy, paste, read, write, & clear

# what changed

- added `context.clipboard` which has the following methods:
  - `readText`, `writeText`, `clear`, `paste`, `copy`, & `cut`
- clipboard read/write runs in the active page context, with browser permissions granted for the page origin
- paste/copy/cut focus the target page before dispatching keyboard shortcuts
- clipboard methods use `context.activePage()` by default, but callers can pass `{ page }` to target a specific `Page`
- added an integration spec for text clipboard behavior across active & explicit pages

# test plan

- added `clipboard.spec.ts` to cover `readText`, `writeText`, `paste`, `copy`, & `cut`
- the spec asserts the exact clipboard & textarea values for pasted, copied, cut, & read text
- also verifies clipboard actions use the active page by default
- also verifies passing `{ page }` targets that specific `Page`, even when another page is active

---
## Summary by cubic
Adds a native clipboard API to the context for read/write and cut/copy/paste. Defaults to the active page, handles permissions, focuses the target page, and logs actions via FlowLogger without duplicates.

- New Features
  - New `context.clipboard` with `readText`, `writeText`, `clear`, `paste`, `copy`, and `cut`.
  - Grants origin clipboard permissions and brings the page to front before sending shortcuts; emits FlowLogger events for each action.
  - Uses `context.activePage()` by default; pass `{ page }` to target a specific `Page`. `paste` supports a `shortcut` override.
  - Exports public types (`BrowserClipboard`, options), and includes an example and integration tests verifying active vs explicit page behavior.

- Bug Fixes
  - Prevents duplicate FlowLogger events for clipboard actions.

<sup>Written for commit d7103c28ac1ebe8340fe05793ce7fc3a29cb69b1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2171?utm_source=github">Review in cubic</a>

### DIFF
--- a/packages/core/lib/v3/types/public/clipboard.ts
+++ b/packages/core/lib/v3/types/public/clipboard.ts
@@ -1,0 +1,19 @@
+import type { Page } from "../../understudy/page.js";
+
+export interface ClipboardOptions {
+  page?: Page;
+  timeout?: number;
+}
+
+export interface ClipboardPasteOptions extends ClipboardOptions {
+  shortcut?: "ControlOrMeta+V" | "Meta+V" | "Control+V";
+}
+
+export interface BrowserClipboard {
+  readText(options?: ClipboardOptions): Promise<string>;
+  writeText(text: string, options?: ClipboardOptions): Promise<void>;
+  clear(options?: ClipboardOptions): Promise<void>;
+  paste(options?: ClipboardPasteOptions): Promise<void>;
+  copySelection(options?: ClipboardOptions): Promise<string>;
+  cutSelection(options?: ClipboardOptions): Promise<string>;
+}
--- a/packages/core/lib/v3/types/public/clipboard.ts
+++ b/packages/core/lib/v3/types/public/clipboard.ts
@@ -14,6 +14,6 @@ export interface BrowserClipboard {
   writeText(text: string, options?: ClipboardOptions): Promise<void>;
   clear(options?: ClipboardOptions): Promise<void>;
   paste(options?: ClipboardPasteOptions): Promise<void>;
-  copySelection(options?: ClipboardOptions): Promise<string>;
-  cutSelection(options?: ClipboardOptions): Promise<string>;
+  copy(options?: ClipboardOptions): Promise<void>;
+  cut(options?: ClipboardOptions): Promise<void>;
 }
--- a/packages/core/tests/integration/clipboard.spec.ts
+++ b/packages/core/tests/integration/clipboard.spec.ts
@@ -1,0 +1,129 @@
+import { test, expect } from "@playwright/test";
+import { V3 } from "../../lib/v3/v3.js";
+import type { Page } from "../../lib/v3/types/public/page.js";
+import { v3DynamicTestConfig } from "./v3.dynamic.config.js";
+import { closeV3 } from "./testUtils.js";
+
+const TEST_URL =
+  "https://browserbase.github.io/stagehand-eval-sites/sites/example/";
+
+async function setupTextarea(
+  page: Page,
+  options: { id: string; value?: string },
+): Promise<void> {
+  await page.goto(TEST_URL, {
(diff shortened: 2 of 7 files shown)


seanmcguire12:
@cubic-dev-ai

packages/core/tests/integration/clipboard.spec.ts
pirate, line 1: You should add a test for clicking an element that triggers copy to clipboard in JS. Then confirm users can read the copied value via the new APIs. Chrome uses user-initiated input actions to gate clipboard behavior, so it's important to test it works with their gating system.

e.g.

```js
var copyTextareaBtn = document.querySelector('.js-textareacopybtn');

copyTextareaBtn.addEventListener('click', function(event) {
  var copyTextarea = document.querySelector('.js-copytextarea');
  copyTextarea.focus();
  copyTextarea.select();

  try {
    var successful = document.execCommand('copy');
    var msg = successful ? 'successful' : 'unsuccessful';
    console.log('Copying text command was ' + msg);
  } catch (err) {
    console.log('Oops, unable to copy');
  }
});
```
```html
<p>
  <button class="js-textareacopybtn" style="vertical-align:top;">Copy Textarea</button>
  <textarea class="js-copytextarea">Hello I'm some text</textarea>
</p>
```

pirate (approved):

packages/core/lib/v3/types/public/clipboard.ts
```diff
@@ -0,0 +1,19 @@
...
+  shortcut?: "ControlOrMeta+V" | "Meta+V" | "Control+V";
+}
+
+export interface BrowserClipboard {
+  readText(options?: ClipboardOptions): Promise<string>;
+  writeText(text: string, options?: ClipboardOptions): Promise<void>;
+  clear(options?: ClipboardOptions): Promise<void>;
+  paste(options?: ClipboardPasteOptions): Promise<void>;
+  copy(options?: ClipboardOptions): Promise<void>;
+  cut(options?: ClipboardOptions): Promise<void>;
```
miguelg719, line 18: should cut/copy return the text? prevents having to read after

packages/core/lib/v3/types/public/clipboard.ts
```diff
@@ -0,0 +1,19 @@
...
+  timeout?: number;
+}
+
+export interface ClipboardPasteOptions extends ClipboardOptions {
+  shortcut?: "ControlOrMeta+V" | "Meta+V" | "Control+V";
+}
+
+export interface BrowserClipboard {
+  readText(options?: ClipboardOptions): Promise<string>;
+  writeText(text: string, options?: ClipboardOptions): Promise<void>;
```
miguelg719, line 14: is Text explicit because this doesn't support images?

pirate: yeah also clipboard supports all kinds of extra interesting things, like multiple mimetype representations for the same text that consumers can choose between, images, etc.

we may want to think about how to design it to not preclude supporting those things in the future with optional args.

packages/core/lib/v3/types/public/clipboard.ts
seanmcguire12: initially i had them returning a string, but then this kind of paints the API into a corner. the thinking was to keep`copy` & `cut` generic APIs, so we can support more content types in the future

packages/core/lib/v3/types/public/clipboard.ts
seanmcguire12: yeah, I intentionally made `Text` explicit so that we can add future read/write functions for the various content types, or add generic `read`/`write` APIs which return an object containing nullable fields of all the possible types.

PR #2174, opened 2026-05-29
fix example script by appending .env

darren-trellis:
# why

`pnpm run example` fails because it does not have secrets.

# what changed

the run script now appends secrets from `.env`

# test plan

N/A

---
## Summary by cubic
Fixes the example runner by optionally loading environment variables from the repo `.env`. Updates the `example` script in `packages/core/package.json` to pass `--env-file-if-exists=../../.env`, so `pnpm run example` picks up secrets when present and doesn’t fail if the file is missing.

<sup>Written for commit 050c72bbd5e2aeb1cb7f9e2224e89a479d1f89d3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2174?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


shrey150 (approved):
Verified locally: with shell env stripped and a populated repo-root `.env`, the current `example` script gets MISSING for `BROWSERBASE_API_KEY`/`OPENAI_API_KEY`; with `--env-file-if-exists=../../.env` they load correctly. Path is right (script cwd is `packages/core`, so `../../.env` = repo root, matching the README's `cp .env.example .env`). Flag is supported across the full engines range — backported in Node v20.19.0 (nodejs/node#53060), exactly the engines floor.

darren-trellis:
@shrey150 could we merge

PR #2200, opened 2026-06-06
[STG-2191] feat(cli): static browse skill nudge + de-spam update notice

shrey150:
Port of [browserbase/cli#142](https://github.com/browserbase/cli/pull/142) into the monorepo, de-scoped to a static model after team review.

## How it works

| When | What you see |
| --- | --- |
| `browse` or `browse --help` | Help opens with a **"Start here (for AI agents)"** banner pointing to `browse skills install`. Always shown. |
| Starting a browser session without the skill installed | A stderr tip — run `browse skills install` — on the command that starts each session (when the driver daemon spawns). Repeats once per session until the skill is installed. Skipped in CI and with `BROWSE_DISABLE_SKILL_NUDGE=1`. |
| Any command while the CLI is outdated | **"Update available"** on stderr — at most **once per 20 hours, repeating until you upgrade** (same semantics as Codex's upgrade banner). `browse --help` and `browse doctor` always show it. |
| Everything else | Nothing. |

Today the update notice prints on **every command** while outdated; with this PR it's one line per ~day. The nudge is stateless — no marker files; the daemon spawn is the session boundary. Skill detection is two filesystem checks (`~/.agents/skills/browse` and `<cwd>/.agents/skills/browse` — the two locations `npx skills` installs to). `skill_present` is reported on telemetry so adoption is measurable.

## E2E Test Matrix

Run against the local build. `<fake home>` = machine without the skill; `<real home>` has it.

| Command / flow | Observed output | Proves |
| --- | --- | --- |
| `HOME=<fake home> browse open <url> --local` (daemon spawns) | stderr tip; stdout clean JSON | Tip fires at browser-session start |
| `browse get title` in the same session | No tip | One tip per session, not per command |
| `browse stop`, then `open` again | Tip again | Repeats each new session until installed |
| Real home (skill installed), fresh session | No tip | Detection suppresses the tip |
| `browse` / `--help` ×2, both homes | Banner present on every run | Banner is static, always-on |
| Fake cache `version: 99.0.0` → `browse status` ×2 → rewind `lastNotifiedAt` 21h → run again → run as v99 | Prints once → silent same sitting → reminds at +21h → silent once upgraded | Update reminder: once per 20h until upgrade |
| `pnpm test` / `pnpm lint` (packages/cli) | 225/225 tests, format+eslint+tsc clean | Full suite incl. nudge, presence, and reminder-semantics tests |

https://linear.app/browserbase/issue/STG-2191


.changeset/browse-skill-nudge.md
```diff
@@ -0,0 +1,7 @@
+---
```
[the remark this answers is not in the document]
shrey150: Addressed in [598d2c6](https://github.com/browserbase/stagehand/commit/598d2c60239395a6079f1faed80b73b25526be24) — dropped the changeset. `browse` is in the `ignore` list in `.changeset/config.json`, so the changeset was a no-op (no version bump / release). Leaving the ignore list as-is intentionally: putting the CLI on the stagehand changeset release train is a release-process decision, not something to flip via this PR. The CLI is versioned/published outside the changeset flow.

packages/cli/src/lib/help.ts
```diff
@@ -0,0 +1,54 @@
...
+      this.log(AGENT_START_HERE);
+    }
+    await super.showRootHelp();
+    await this.writeUpdateNotice();
+  }
+
+  private async skillBannerNeeded(): Promise<boolean> {
+    try {
+      const agent = await detectAgent();
+      return !(await isBrowseSkillInstalled(agent ?? "", process.env));
```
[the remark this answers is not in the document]
shrey150: Working as intended here — leaving as-is. The root-help banner is deliberately gated on skill **presence**, not agent detection: it only appears on the opt-in `browse` / `browse --help` surface (which a user explicitly invokes), and there's an explicit test asserting it shows when the skill is absent regardless of agent. The thing that could actually spam — the unsolicited per-command stderr nudge in `skill-nudge.ts` — *is* correctly agent-gated (`if (!agent) return`) and throttled. Gating the help banner on agent detection too would also make the surface test environment-dependent (agent detection varies by CI/host). Happy to switch it to agent-gated if we'd rather not show the pointer to humans on `--help`.

.changeset/browse-skill-nudge.md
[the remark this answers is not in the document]
shrey150: Correction to my earlier reply — restored the changeset in [3362936](https://github.com/browserbase/stagehand/commit/336293616b6c3fc15575fe546e3b01ad6a079354). I was wrong that it was a no-op: `browse` is released by the decoupled `.github/workflows/release-cli.yml`, which scans `.changeset/*.md` for `"browse"` entries and bumps the CLI from them. The `ignore` entry in `.changeset/config.json` only keeps the **core** stagehand release from bumping `browse`; the CLI release flow temporarily re-scopes ignore to consume browse-only changesets. This changeset is browse-only, so it satisfies that workflow's no-mixed-changeset rule.

packages/cli/src/commands/doctor.ts
```diff
@@ -42,6 +45,20 @@ export default class Doctor extends BrowseCommand {
...
 
     this.log(renderDoctorReport(report));
+    await this.writeUpdateNotice();
     if (report.verdict === "fail") this.exit(1);
   }
+
+  private async writeUpdateNotice(): Promise<void> {
+    try {
+      const notice = await getUpdateNotice(this.config.version, process.env, {
+        cacheFile: join(this.config.cacheDir, "update-check.json"),
```
Kylejeong2, line 55: do we save a cache file to tmp dir?

shrey150: Good question — no, nothing goes to a tmp dir. Both cache files live in oclif's per-CLI cache directory (`this.config.cacheDir`): `update-check.json` and `skill-nudge.json`. That resolves to `~/Library/Caches/browse` on macOS, `~/.cache/browse` on Linux, and `%LOCALAPPDATA%\browse` on Windows — stable across invocations (so the 24h TTL works) and cleaned up with the usual OS cache semantics. Same location the pre-existing update check already used on main (`src/hooks/init.ts`).

shrey150:
@Kylejeong2 + team — de-scoped per the review discussion (19921e1 and follow-ups): all dynamic nudge logic (agent detection, session keys, TTL windows) is gone. Current behavior and test evidence are in the PR description.

packages/cli/src/commands/doctor.ts
```diff
@@ -42,6 +45,20 @@ export default class Doctor extends BrowseCommand {
     }
 
     this.log(renderDoctorReport(report));
+    await this.writeUpdateNotice();
     if (report.verdict === "fail") this.exit(1);
   }
+
+  private async writeUpdateNotice(): Promise<void> {
```
shrey150, line 52: what is this? why do we need it?

packages/cli/src/commands/doctor.ts
shrey150: does skill-nudge.json still exist after simplifying the impl here?

packages/cli/src/lib/update.ts
```diff
@@ -38,15 +66,19 @@ export async function maybeAutoUpdateCli(
...
     return;
   }
 
   spawnBackgroundUpdateCheck(env, cachePath);
 }
 
+function isUpdateCheckDisabled(env: NodeJS.ProcessEnv): boolean {
+  return (
+    env.BROWSE_DISABLE_UPDATE_CHECK === "1" ||
+    env.BB_DISABLE_UPDATE_CHECK === "1"
```
shrey150, line 137: why are both these env vars set? are they both in use?

packages/cli/src/commands/doctor.ts
shrey150: This is the doctor half of the de-spam: on main, the "Update available" notice prints to stderr on **every command** (init hook). This PR makes the init hook silent (cache refresh only) and renders the notice on exactly two surfaces — root help and `browse doctor`. Doctor gets it because it's the "check my setup" command and an outdated CLI is genuinely diagnostic info there. Note it's best-effort stderr and skipped entirely in `--json` mode (the early return above it), so machine output stays clean. If you'd rather the notice live on root help only, deleting this method is a 1-hunk change.

packages/cli/src/commands/doctor.ts
shrey150: The file on this line is `update-check.json` (the update-notice cache, a concept that already exists on main) — different file from the nudge marker. But to the question: yes, `skill-nudge.json` still exists post-simplification (written by the init hook, see hooks/init.ts:28) — it's now just a once-per-install marker `{"shownAt": ...}`. What got deleted was the machinery *inside* it: the per-agent/session-key map, 4h TTL window, and 7d pruning. Two cache files total, one job each: `update-check.json` = update state, `skill-nudge.json` = "first-run hint already shown".

packages/cli/src/lib/update.ts
shrey150: Pre-existing, not added here — main's update.ts:28-29 has the identical pair; the diff only shows it because the check moved into the extracted `isUpdateCheckDisabled()` helper. `BB_*` is the legacy env-prefix family from the bb CLI heritage (also honored in functions/dev.ts), kept as a backcompat alias alongside `BROWSE_*`. Both are "in use" in the sense that both are honored; if you want to retire the `BB_*` aliases that's a separate cleanup across the package rather than this PR.

PR #2201, opened 2026-06-06
[STG-2080] feat: add managed local Chrome args

shrey150:
## Summary

Port of [browserbase/cli#127](https://github.com/browserbase/cli/pull/127) into the stagehand monorepo, where the browse CLI now lives at `packages/cli` (package `browse`). The old `browserbase/cli` repo is deprecated.

Adds a managed-local Chrome launch-arg surface to the browse CLI, matching the Playwright / Puppeteer / chrome-launcher convention (three repeatable/boolean flags, no JSON):

- `--chrome-arg <flag>` (repeatable) — append a launch arg on top of Chrome's defaults
- `--ignore-default-chrome-arg <flag>` (repeatable) — drop a specific default arg
- `--no-default-chrome-args` — launch without any of Chrome's default args

All three are **managed-local only** (rejected with `--remote` / `--cdp` / `--auto-connect`); `--no-default-chrome-args` and `--ignore-default-chrome-arg` are mutually exclusive. Args are threaded into Stagehand `localBrowserLaunchOptions` (`args` / `ignoreDefaultArgs`) and tracked as part of daemon target compatibility so session reuse respects them.

**Usage note:** because `--chrome-arg` / `--ignore-default-chrome-arg` are variadic, pass the URL *before* them (`browse open <url> --local --chrome-arg=--foo`) or terminate flags with `--`. The CLI prints a clear error pointing this out otherwise.

## Files

- `packages/cli/src/lib/driver/flags.ts` — new flag definitions
- `packages/cli/src/lib/driver/mode.ts` — flag parsing, conflict checks, `managedLocalTarget(...)` / `resolveChromeArgs(...)` helpers, chrome-arg target-compatibility comparison (merged on top of the monorepo's `getRemote()` / async `stagehandOptions` refactor)
- `packages/cli/src/lib/driver/command-cli.ts` — `hasChromeArgFlags(flags)` wired into `hasExplicitDriverTarget` / `hasModeOnlyFlag`
- `packages/cli/src/lib/driver/types.ts` — `chromeArgs` / `ignoreDefaultArgs` / `noDefaultChromeArgs` on the managed-local target
- `packages/cli/src/lib/driver/session-manager.ts` — forwards args into `localBrowserLaunchOptions`
- `packages/cli/README.md` — flag table rows + example
- `packages/cli/tests/driver-{commands,foundation}.test.ts` — parsing, conflicts, compatibility, pass-through
- `.changeset/managed-local-chrome-args.md` — `browse` patch

## E2E Test Matrix

| --- | --- | --- |
| `npx turbo run build --filter=browse` | `4 successful, 4 total` — builds `@browserbasehq/stagehand` (with `gen-version` + `build-dom-scripts` prereqs) then `browse`; `tsc` + oclif manifest clean | Proves the merged driver files compile against the monorepo's refactored core. |
| `pnpm test` (in `packages/cli`) | `Test Files 15 passed (15)` / `Tests 212 passed (212)` — incl. the new managed-local chrome-arg foundation tests and driver-command parsing/conflict/compatibility tests | Unit + contract coverage for flag parsing, conflict rules, daemon target compatibility/reuse, and Stagehand option pass-through. |
| `pnpm lint` (in `packages/cli`) | `prettier --check` clean, `eslint` clean, `tsc --noEmit` clean | Format / lint / typecheck green on the ported + merged files. |

Linear: https://linear.app/browserbase/issue/STG-2080

---
## Summary by cubic
Adds managed-local Chrome launch-arg flags to the `browse` CLI so users can add or override Chrome defaults. Fulfills Linear STG-2080 and ensures daemon session reuse only when these options match.

- **New Features**
  - New flags: `--chrome-arg <flag>` (repeatable), `--ignore-default-chrome-arg <flag>` (repeatable), `--no-default-chrome-args`.
  - Managed-local only; rejected with `--remote`, `--cdp`, `--auto-connect`, or implicit remote mode.
  - `--no-default-chrome-args` and `--ignore-default-chrome-arg` are mutually exclusive.
  - Flags flow into `localBrowserLaunchOptions` (`args`, `ignoreDefaultArgs`) for `@browserbasehq/stagehand`.
  - Daemon session reuse now compares these args; chrome-arg flags also mark an explicit driver target.

- **Refactors**
  - Use Node’s `isDeepStrictEqual` to compare `chromeArgs` and `ignoreDefaultArgs` for target compatibility.
  - Restored `browse` patch changeset for the decoupled release workflow.

<sup>Written for commit e3a953169e336bc945c985bdb5c5d02aad440f07. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2201?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cli/src/lib/driver/mode.ts
+++ b/packages/cli/src/lib/driver/mode.ts
@@ -1,3 +1,5 @@
+import { isDeepStrictEqual } from "node:util";
+
 import { fail } from "../errors.js";
 import { getRemote } from "./remote-binding.js";
 import { resolveWsTarget } from "./resolve-ws.js";
@@ -178,12 +180,7 @@ export function targetsCompatible(
 }
 
 function chromeArgsEqual(left?: string[], right?: string[]): boolean {
-  const normalizedLeft = left?.length ? left : [];
-  const normalizedRight = right?.length ? right : [];
-  return (
-    normalizedLeft.length === normalizedRight.length &&
-    normalizedLeft.every((arg, index) => arg === normalizedRight[index])
-  );
+  return isDeepStrictEqual(left ?? [], right ?? []);
 }
 
 function ignoreDefaultArgsEqual(
--- a/packages/cli/src/lib/driver/mode.ts
+++ b/packages/cli/src/lib/driver/mode.ts
@@ -6,11 +6,19 @@ import type { ConnectionTarget } from "./types.js";
 export interface DriverModeFlags {
   "auto-connect"?: boolean;
   cdp?: string;
+  "chrome-arg"?: string[];
   headed?: boolean;
   headless?: boolean;
+  "ignore-default-chrome-arg"?: string[];
   local?: boolean;
+  "no-default-chrome-args"?: boolean;
   remote?: boolean;
   "target-id"?: string;
+}
+
+interface ResolvedChromeArgs {
+  args?: string[];
+  ignoreDefaultArgs?: boolean | string[];
 }
 
 function resolveHeadless(
@@ -25,12 +33,28 @@ function resolveHeadless(
   return true;
 }
 
+export function hasChromeArgFlags(flags: DriverModeFlags): boolean {
+  return chromeArgFlagsInUse(flags).length > 0;
+}
+
+function chromeArgFlagsInUse(flags: DriverModeFlags): string[] {
+  const names: string[] = [];
+  if (flags["chrome-arg"]?.length) names.push("--chrome-arg");
+  if (flags["ignore-default-chrome-arg"]?.length)
+    names.push("--ignore-default-chrome-arg");
+  if (flags["no-default-chrome-args"]) names.push("--no-default-chrome-args");
+  return names;
+}
+
 export async function resolveConnectionTarget(
   flags: DriverModeFlags,
 ): Promise<ConnectionTarget> {
+  const chromeArgFlags = chromeArgFlagsInUse(flags);
+
   if (flags.cdp) {
     failOnConflictingFlags("--cdp", [
       flags["auto-connect"] ? "--auto-connect" : null,
+      ...chromeArgFlags,
       flags.local ? "--local" : null,
       flags.remote ? "--remote" : null,
       flags.headed ? "--headed" : null,
@@ -49,6 +73,7 @@ export async function resolveConnectionTarget(
 
   if (flags["auto-connect"]) {
     failOnConflictingFlags("--auto-connect", [
+      ...chromeArgFlags,
       flags.local ? "--local" : null,
       flags.remote ? "--remote" : null,
       flags.headed ? "--headed" : null,
@@ -63,6 +88,7 @@ export async function resolveConnectionTarget(
 
   if (flags.remote) {
     failOnConflictingFlags("--remote", [
+      ...chromeArgFlags,
       flags.headed ? "--headed" : null,
       flags.headless ? "--headless" : null,
     ]);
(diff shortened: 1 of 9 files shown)


packages/cli/src/lib/driver/mode.ts
```diff
@@ -96,15 +137,62 @@ function failOnConflictingFlags(
...
+      chromeArgsEqual(left.chromeArgs, right.chromeArgs) &&
+      ignoreDefaultArgsEqual(left.ignoreDefaultArgs, right.ignoreDefaultArgs)
+    );
   if (left.kind === "cdp" && right.kind === "cdp") {
     return left.endpoint === right.endpoint && left.targetId === right.targetId;
   }
   return true;
 }
+
+function chromeArgsEqual(left?: string[], right?: string[]): boolean {
```
shrey150, line 182: is there no util function (like lodash) we can use here?

packages/cli/src/lib/driver/mode.ts
```diff
@@ -96,15 +137,62 @@ function failOnConflictingFlags(
...
+  const normalizedLeft = left?.length ? left : [];
+  const normalizedRight = right?.length ? right : [];
+  return (
+    normalizedLeft.length === normalizedRight.length &&
+    normalizedLeft.every((arg, index) => arg === normalizedRight[index])
+  );
+}
+
+function ignoreDefaultArgsEqual(
```
shrey150, line 186: same comment here: is there no util function like lodash we can use here?

packages/cli/src/lib/driver/mode.ts
shrey150: Addressed in [e3a9531](https://github.com/browserbase/stagehand/commit/e3a953169e336bc945c985bdb5c5d02aad440f07) -- replaced the hand-rolled comparison with Node's built-in `isDeepStrictEqual` from `node:util`. Lodash isn't a dependency here (only a pnpm security override at the root), so the builtin avoids adding one.

packages/cli/src/lib/driver/mode.ts
shrey150: Addressed in [e3a9531](https://github.com/browserbase/stagehand/commit/e3a953169e336bc945c985bdb5c5d02aad440f07) -- `ignoreDefaultArgsEqual` now delegates to the same `isDeepStrictEqual`-based helper; only the `true`-vs-array short-circuit remains hand-written since it's a semantic rule (bool `true` only equals `true`), not a deep-equality check.

packages/cli/tests/driver-foundation.test.ts
```diff
@@ -63,6 +63,48 @@ describe("driver foundation", () => {
...
+        local: true,
+      }),
+    ).resolves.toEqual({
+      headless: true,
+      ignoreDefaultArgs: ["--enable-automation"],
+      kind: "managed-local",
+    });
+    await expect(
+      resolveConnectionTarget({
+        "chrome-arg": ["--no-sandbox"],
```
pirate, line 98: I recommend only adding `--no-sandbox` on linux systems that dont have a `$DISPLAY` set, thats a good heuristic for "am I inside docker".

We definitely dont want to disable sandboxing on macOS or linux desktop machines for example

pirate (approved):
Up to you, you can do it this way or `--chrome-args='["--json", "list", "parses", "nicely"]'`.

I'd just double check the sandbox arg logic too.
Will restamp if you decide to change it to json strings.

packages/cli/tests/driver-foundation.test.ts
shrey150: Good callout — verified the logic and the CLI never injects `--no-sandbox` automatically. This test fixture is just exercising user-supplied pass-through: the only way the sandbox gets disabled is if the user explicitly passes `--chrome-arg=--no-sandbox`. Nothing in `resolveChromeArgs()` / `packages/core`'s local launch path (or chrome-launcher's defaults) adds it on any platform.

Agreed on the heuristic if we ever add Docker auto-detection (Linux + no `$DISPLAY`, never macOS/desktop) — considering that as a separate follow-up rather than this PR.

shrey150:
Re: the `--chrome-args='[...]'` JSON suggestion — going to keep the repeatable-flag form. It avoids shell-quoting JSON (error-prone for both humans and agents invoking the CLI), and matches the repeatable-flag convention of the existing surface. Since both shapes were fine by you I'll leave the approval as-is — thanks for the review!

On the sandbox logic: double-checked it — `--no-sandbox` only appears as a user-supplied test fixture; the CLI never adds it automatically on any platform (details in the inline thread).

PR #2203, opened 2026-06-07
 fix: stop injecting --site-per-process into local Chromium launches

### LINKED ISSUE #2193: Hardcoded --site-per-process in local launcher can't be disabled by users
**Before submitting an issue, please:**
- [x] Check the [documentation](https://docs.stagehand.dev/) for relevant information
- [x] Search existing [issues](https://github.com/browserbase/stagehand/issues) to avoid duplicates

## Environment Information

**Stagehand:**
- Language/SDK: TypeScript
- Stagehand version: 3.2.0

**AI Provider:**
- Provider: Anthropic
- Model: claude-haiku-4-5

**Runtime:**
- Node.js: v25.6.1
- OS: macOS (Darwin 25.5.0); also observed on Linux x86_64
- Browser: Chromium 145.0.7632.109

## Issue Description

```
Stagehand's local launcher injects `--site-per-process` into Chromium's argv
from TWO separate places in the v3 path:

  A. dist/esm/lib/v3/v3.js:643 (CJS: dist/cjs/lib/v3/v3.js:682)
     This one IS opt-out-able — v3.js:646-656 honors
     `localBrowserLaunchOptions.ignoreDefaultArgs` to skip or filter
     the defaults array.

  B. dist/esm/lib/v3/launch/local.js:16 (CJS: dist/cjs/lib/v3/launch/local.js:22)
     This one is NOT opt-out-able. `launchLocalChrome` builds its own
     defaults array and prepends `--site-per-process` unconditionally
     before appending the caller's `chromeFlags`. No public option turns
     it off; the only way to disable it is to patch the file.

The result is that even users who do everything right at the v3.js layer
(`ignoreDefaultArgs: ["--site-per-process"]`) still get the flag in argv
from the local.js layer. Both copies need to go.

This causes two concrete problems:

1. `--renderer-process-limit` is silently uncapped.
   Chromium docs note: when site isolation is on, the renderer process pool
   grows roughly one renderer per origin, and `--renderer-process-limit`
   becomes a soft target that the browser routinely exceeds. With site
   isolation off, the limit binds. Stagehand's hardcoded `--site-per-process`
   therefore defeats any user attempt to bound renderer count.

2. `--disable-features=site-per-process` does NOT override `--site-per-process`.
   Chromium's command-line parser gives the explicit switch precedence over
   the feature-list disable, regardless of argv order or count. So passing
   `--disable-features=site-per-process` via `localBrowserLaunchOptions.args`
   has no effect as long as Stagehand also injects the explicit switch
   (from either A or B above — even one copy is enough to win).

For users who reach Stagehand through a wrapper that constructs
`localBrowserLaunchOptions` itself with a fixed shape (e.g.
`@crawlee/stagehand`, which passes only `{ headless, args, executablePath,
proxy, viewport }` and never propagates `ignoreDefaultArgs`), the only
opt-out Stagehand currently offers is inaccessible. Such users have no
non-patching way to disable site isolation.
```

### Steps to Reproduce

1. Start a Stagehand v3 LOCAL session against any multi-origin SPA (OWASP Juice Shop is a convenient public target that can be spin up via docker).
2. Pass `localBrowserLaunchOptions.args: ["--disable-features=site-per-process", "--renderer-process-limit=6"]`.
3. Navigate across a handful of SPA routes that issue cross-origin requests.
4. Inspect `ps -ww` on the Chromium parent process — `--site-per-process` appears in argv despite the user's intent to disable it.
5. Count `--type=renderer` descendant processes — the renderer pool grows past the requested cap.

### Minimal Reproduction Code

```ts
import { Stagehand } from "@browserbasehq/stagehand";

  env: "LOCAL",
  verbose: 0,
  localBrowserLaunchOptions: {
    headless: true,
    args: [
      "--disable-features=site-per-process,IsolateOrigins",
      "--renderer-process-limit=6",
    ],
    // Even uncommenting this won't fully clear --site-per-process from argv:
    // it removes the v3.js copy (v3.js:646-656 honours ignoreDefaultArgs),
    // but the launchLocalChrome copy at local.js:16 is unconditional and
    // still fires.
    // ignoreDefaultArgs: ["--site-per-process"],
  },
});

await stagehand.init();
const page = stagehand.page;

// Visit a few routes on any multi-origin SPA.
const urls = [
  "http://localhost:3000/",
  "http://localhost:3000/#/route-a",
  "http://localhost:3000/#/route-b",
  "http://localhost:3000/#/route-c",
];
for (const url of urls) {
  await page.goto(url, { waitUntil: "networkidle" });
  await new Promise((r) => setTimeout(r, 5_000));
}

// Inspect from a separate shell:
//   ps -ww -o command= $(pgrep -f 'chromium' | head -1)  # has --site-per-process
//   pgrep -af 'type=renderer' | wc -l                    # > requested limit
```

### Error Messages / Log trace

```
# `ps -ww` on the Chromium parent during the repro:
--site-per-process                                       ← from launchLocalChrome (local.js:16)
--site-per-process                                       ← from V3.init defaults (v3.js:643)
--disable-features=...,site-per-process                  ← user's intended disable (ignored)
--renderer-process-limit=6                               ← user's intended cap

# `--type=renderer` descendant count over time on a multi-origin SPA crawl
# (renderer cap = 6 requested):
t=0s    4 renderers
t=30s   6 renderers
t=90s   8 renderers
t=150s  11 renderers
t=240s  14 renderers
```

No exception is raised by Stagehand — the failure is silent. Symptoms surface downstream as memory pressure / OOM kill of the browser process tree under any host or container memory ceiling.

### Screenshots / Videos

```
N/A — issue is observable via `ps` output rather than UI.
```

### Suggested fix

Pick one (in order of intrusiveness):

1. **Drop `--site-per-process` from both default lists entirely.** It is a security feature designed for end-user browsing; it is not required for automation. Users who want it can still opt in via their own `args`. (One-line change at `v3.js:643` and `local.js:16`, plus the matching CJS builds.)

2. **Have `local.js` honour an `ignoreDefaultArgs` option** mirroring v3.js's behaviour, and have `v3.js` forward `lbo.ignoreDefaultArgs` into the call to `launchLocalChrome`. Users could then pass `localBrowserLaunchOptions.ignoreDefaultArgs: ["--site-per-process"]` to neutralize both sites in one place.

3. **Document the current behaviour** as a known limitation and explicitly point users to `localBrowserLaunchOptions.cdpUrl` as the workaround (launch their own Chromium, hand the WS URL to Stagehand — which then takes the `v3.js:598` branch that skips both flag-injection sites).

### Verification

We confirmed option 1 by removing the two hardcoded `--site-per-process` lines from `node_modules/@browserbasehq/stagehand/dist/esm/lib/v3/{v3,launch/local}.js` and re-running the repro.

Result after the patch:
- `ps -ww` on the Chromium parent shows zero explicit `--site-per-process` switches (only the substring inside the user-intended `--disable-features=...,site-per-process`).
- `--renderer-process-limit=N` binds: the renderer count holds at the configured value rather than growing one-per-origin.
- Residual renderer count may sit slightly above N (typically +1 to +3) due to Chromium's spare-renderer cache and service workers, both of which are inherent to Chrome and out of Stagehand's control.
- No functional regression: `page.act()`, navigation, form detection, and other Stagehand features continue to work.

### Related Issues

- Related to: N/A
- Duplicate of: N/A
- Blocks: N/A

optimusbuilder:
# why

Stagehand unconditionally injected `--site-per-process` when launching local Chromium. That explicit switch overrides `--disable-features=site-per-process` and defeats `--renderer-process-limit` on multi-origin workloads, causing renderer pools to grow unbounded and leading to memory pressure/OOM kills under container limits.

The duplicate injection in `v3.ts` was already removed in #2127; this removes the remaining copy in `launch/local.ts`.

# what changed

- Dropped `--site-per-process` from `STAGEHAND_DEFAULT_FLAGS` in `packages/core/lib/v3/launch/local.ts`
- Dropped the same flag from `packages/evals/core/targets/localChrome.ts`
- Added regression tests in `launch-local-ignore-default-args.test.ts`
- Added a patch changeset for `@browserbasehq/stagehand`

# test plan

- [x] `pnpm --filter @browserbasehq/stagehand exec vitest run dist/esm/tests/unit/launch-local-ignore-default-args.test.js` (13/13 passed)
- [x] Live Chrome launch: parent argv has no standalone `--site-per-process`; user disable/cap flags present
- [x] Full multi-origin SPA repro: renderer count stays near configured `--renderer-process-limit`

---
## Summary by cubic
Stop injecting `--site-per-process` into local Chromium launches. This lets user flags like `--disable-features=site-per-process` and `--renderer-process-limit` work and prevents runaway renderer counts.

- **Bug Fixes**
  - Removed the flag from `packages/core/lib/v3/launch/local.ts` and `packages/evals/core/targets/localChrome.ts`.
  - Added regression tests in `packages/core/tests/unit/launch-local-ignore-default-args.test.ts` (includes opt-in via args).
  - Added a patch changeset for `@browserbasehq/stagehand`.

<sup>Written for commit 736539650a7125e2221a7b1b420baaf87f2a1005. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2203?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Fixes #2193

### DIFF
--- a/.changeset/remove-site-per-process-default.md
+++ b/.changeset/remove-site-per-process-default.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+Stop injecting `--site-per-process` into local Chromium launches so user-supplied flags like `--disable-features=site-per-process` and `--renderer-process-limit` take effect.
--- a/packages/core/lib/v3/launch/local.ts
+++ b/packages/core/lib/v3/launch/local.ts
@@ -12,7 +12,6 @@ const STAGEHAND_DEFAULT_FLAGS = [
   "--no-first-run",
   "--no-default-browser-check",
   "--disable-dev-shm-usage",
-  "--site-per-process",
 ];
 
 export async function launchLocalChrome(
--- a/packages/core/tests/unit/launch-local-ignore-default-args.test.ts
+++ b/packages/core/tests/unit/launch-local-ignore-default-args.test.ts
@@ -80,6 +80,16 @@ async function getLaunchArgs(
 }
 
 describe("launchLocalChrome ignoreDefaultArgs", () => {
+  it("does not inject --site-per-process by default", async () => {
+    const args = await getLaunchArgs({});
+    expect(args.chromeFlags).not.toContain("--site-per-process");
+  });
+
+  it("allows users to opt in to --site-per-process via args", async () => {
+    const args = await getLaunchArgs({ args: ["--site-per-process"] });
+    expect(args.chromeFlags).toContain("--site-per-process");
+  });
+
   it("does not set ignoreDefaultFlags when ignoreDefaultArgs is omitted", async () => {
     const args = await getLaunchArgs({});
     expect(args.ignoreDefaultFlags).toBe(false);
--- a/packages/evals/core/targets/localChrome.ts
+++ b/packages/evals/core/targets/localChrome.ts
@@ -118,7 +118,6 @@ export async function launchRunnerProvidedLocalChrome(): Promise<{
     "--no-first-run",
     "--no-default-browser-check",
     "--disable-dev-shm-usage",
-    "--site-per-process",
     `--remote-debugging-port=${port}`,
     `--user-data-dir=${userDataDir}`,
     "about:blank",


seanmcguire12:
hey @optimusbuilder! the issue that you're aiming to address here has already been fixed. in version 3.5.0, you can ignore default args by passing in `ignoreDefaultArgs` like this:

```typescript
  env: "LOCAL",
  model: "openai/gpt-5",
  localBrowserLaunchOptions: {
    ignoreDefaultArgs: ["--site-per-process"],
  },
  verbose: 2,
});
```

for reference, see [this test](https://github.com/browserbase/stagehand/blob/4e6f5ccfe5567c75ce9def00bb4e3a65c5eb1f8c/packages/core/tests/unit/launch-local-ignore-default-args.test.ts#L197)

PR #2227, opened 2026-06-10
fix(core): normalize Act cache URL keys

### LINKED ISSUE #2189: core(cache): ActCache key derivation isn't URL-normalized, causing silent cache misses

RitwijParmar:
Fixes #2189.

This changes ActCache key derivation so query parameter order no longer causes silent cache misses. The raw page URL is still stored in the cache entry for debugging. Only the hash input is normalized.

I also added a legacy raw URL key fallback. That keeps existing cache files reachable after the normalized key starts being used, instead of making old entries quietly unreachable.

What changed

- Added normalizeUrlForCacheKey for conservative query parameter sorting.
- Used the normalized URL when building Act cache keys.
- Preserved the raw pageUrl in ActCacheContext and stored cache entries.
- Added legacyCacheKey fallback reads for old raw URL cache files.
- Added tests for reordered query params, stable URLs, malformed inputs, repeated params, and legacy fallback replay.

Validation

- bunx prettier --check on changed files passed.
- git diff --check passed.
- Direct Bun runtime check for normalizeUrlForCacheKey passed.

Could not run the full Vitest path locally because pnpm is not installed in this checkout, and bunx vitest fails before running tests due a Rolldown native binding code signature error on macOS. Bundling ActCache directly also hits Stagehand's generated version file, so I left the full suite to CI.

---
## Summary by cubic
Normalize Act cache keys by sorting URL query parameters so reordering no longer causes cache misses. Adds a legacy raw-URL fallback to keep existing cache entries accessible (fixes #2189).

- **Bug Fixes**
  - Use a normalized URL only for cache key hashing; keep the raw `pageUrl` in context and stored entries.
  - Fallback to legacy raw-URL keys when normalized reads miss, preserving access to old cache files.
  - Add tests for param reordering, repeated params, malformed inputs, stable URLs, and legacy fallback.

<sup>Written for commit bee1a2109bb2c7277eca3a295946251664be747f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2227?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/core/lib/v3/cache/ActCache.ts
+++ b/packages/core/lib/v3/cache/ActCache.ts
@@ -4,7 +4,11 @@ import type { LLMClient } from "../llm/LLMClient.js";
 import type { Action, ActResult, Logger } from "../types/public/index.js";
 import type { Page } from "../understudy/page.js";
 import { CacheStorage } from "./CacheStorage.js";
-import { safeGetPageUrl, waitForCachedSelector } from "./utils.js";
+import {
+  normalizeUrlForCacheKey,
+  safeGetPageUrl,
+  waitForCachedSelector,
+} from "./utils.js";
 import {
   ActCacheContext,
   ActCacheDeps,
@@ -50,14 +54,21 @@ export class ActCache {
       ? Object.keys(sanitizedVariables).sort()
       : [];
     const pageUrl = await safeGetPageUrl(page);
+    const normalizedPageUrl = normalizeUrlForCacheKey(pageUrl);
     const cacheKey = this.buildActCacheKey(
       sanitizedInstruction,
-      pageUrl,
+      normalizedPageUrl,
       variableKeys,
     );
+    const legacyCacheKey =
+      normalizedPageUrl === pageUrl
+        ? undefined
+        : this.buildActCacheKey(sanitizedInstruction, pageUrl, variableKeys);
+
     return {
       instruction: sanitizedInstruction,
       cacheKey,
+      legacyCacheKey,
       pageUrl,
       variableKeys,
       variables: sanitizedVariables,
@@ -72,7 +83,7 @@ export class ActCache {
   ): Promise<ActResult | null> {
     if (!this.enabled) return null;
 
-    const {
+    let {
       value: entry,
       error,
       path,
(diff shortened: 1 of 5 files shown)


RitwijParmar:
Closing this PR because it has been inactive for several months and is no longer merge-ready. Thanks for the consideration.

PR #2231, opened 2026-06-10
feat(llm): support claude-fable-5 (structured outputs, adaptive thinking, refusal fallback)

miguelg719:
## Why

Claude Fable 5 breaks Stagehand today in three independent ways: it rejects forced tool use (`400 tool_choice forces tool use is not compatible with this model`), which kills both `act`/`extract`/`observe` (forced `json` tool) and the agent's final `done` call; it rejects sampling parameters like `temperature`; and its safety classifiers can decline a turn (`stop_reason: "refusal"`) with no retry story. Separately, the hybrid/DOM agent path has never sent any Anthropic thinking config, so adaptive-thinking models ran with thinking off.

## What changed

- **`@ai-sdk/anthropic` 2.0.57 → 2.0.81** (v5-compatible line, within `^2.0.34`; floor raised). The provider's capability table now knows fable-5/opus-4-7/4-8: the existing `structuredOutputMode: "auto"` resolves to native `output_config.format` instead of the forced json tool, rejected sampling params are stripped, and `max_tokens` is sized from the real 128k limit — zero code changes for the primitives.
- **New `lib/v3/llm/anthropicOptions.ts`** — one capability table for Anthropic agent models (adaptive-thinking set, xhigh-capable set, fable-5 constants, `rejectsForcedToolUse`).
- **Adaptive thinking on the hybrid/DOM agent path** via typed provider options on both `streamText` call sites. Effort defaults to a shared **`medium`** on both agent paths and is overridable per client via `thinkingEffort` in the agent model config (plumbed through `V3AgentHandler` so the typed option now works on the hybrid path too); `ThinkingEffort` gains `xhigh` (clamped to `high` on models that reject it), and `"none"` opts out of thinking on both paths. The CUA client drops its inline model list and effort default for the shared ones.
- **Server-side refusal fallback**: fable-5 requests carry `fallbacks: [{model: "claude-opus-4-8"}]` as a typed provider option on agent and act/extract/observe calls; the provider adds the `server-side-fallback-2026-06-01` beta header automatically and reports attribution via `usage.iterations`.
- **Final `done` call**: consults `rejectsForcedToolUse()` and goes straight to `toolChoice: "auto"` on fable-5 (forced tool choice is incompatible with always-on thinking); a narrow `/tool_choice/i` catch remains as a safety net for unknown models.
- **Registration**: `claude-fable-5` in `AgentProvider`, `AVAILABLE_CUA_MODELS`, and the public-api test.

Not in scope: CUA-mode (raw `@anthropic-ai/sdk` 0.39.0) fallback wiring — the pinned SDK predates the `fallbacks` param; a refusal there surfaces as `stop_reason: "refusal"` without retry.

## Tests

- `pnpm typecheck` and `build:esm` green; 34/34 unit tests (12 new covering the capability helpers: adaptive set, xhigh clamping, effort precedence, fallbacks gating, forced-tool gate).
- Offline request-body capture (mock fetch, 2.0.81): `generateObject` on fable-5 sends `output_config.format = {type: "json_schema", …, additionalProperties: false}` + `structured-outputs-2025-11-13` beta with **no** `tools`/`tool_choice`; `generateText` with the agent options sends `thinking: {"type":"adaptive"}`, `output_config: {"effort":"xhigh"}`, `fallbacks: [{"model":"claude-opus-4-8"}]` and the `effort-2025-11-24,server-side-fallback-2026-06-01` betas.
- Not yet run against the live API from this machine — worth one live pass (extract + DOM/hybrid agent on fable-5) before un-drafting.

### DIFF
--- a/packages/core/lib/v3/llm/anthropicOptions.ts
+++ b/packages/core/lib/v3/llm/anthropicOptions.ts
@@ -28,8 +28,6 @@ import { stripModelProvider } from "../../utils.js";
  */
 export type AnthropicAgentProviderOptions = Record<string, JSONValue>;
 
-export const ANTHROPIC_FABLE_5_MODEL_ID = "claude-fable-5" as const;
-
 // Fallback model used when Fable 5 declines a turn.
 export const ANTHROPIC_FABLE_5_FALLBACK_MODEL_ID = "claude-opus-4-8" as const;
 
@@ -41,7 +39,7 @@ const ADAPTIVE_THINKING_MODEL_BASES = new Set<string>([
   "claude-opus-4-7",
   "claude-opus-4-8",
   "claude-sonnet-4-6",
-  ANTHROPIC_FABLE_5_MODEL_ID,
+  "claude-fable-5",
 ]);
 
 // Models that accept effort "xhigh". Sending xhigh to other models (e.g.
@@ -49,7 +47,7 @@ const ADAPTIVE_THINKING_MODEL_BASES = new Set<string>([
 const XHIGH_CAPABLE_MODEL_BASES = new Set<string>([
   "claude-opus-4-7",
   "claude-opus-4-8",
-  ANTHROPIC_FABLE_5_MODEL_ID,
+  "claude-fable-5",
 ]);
 
 /** True for Anthropic models that support adaptive thinking. */
@@ -58,7 +56,7 @@ export function isAdaptiveThinkingAnthropicModel(modelId: string): boolean {
 }
 
 export function isAnthropicFable5Model(modelId: string): boolean {
-  return stripModelProvider(modelId) === ANTHROPIC_FABLE_5_MODEL_ID;
+  return stripModelProvider(modelId) === "claude-fable-5";
 }
 
 /**
--- a/packages/core/lib/v3/llm/anthropicOptions.ts
+++ b/packages/core/lib/v3/llm/anthropicOptions.ts
@@ -1,0 +1,130 @@
+/**
+ * Anthropic model capabilities + typed provider options for the agent paths.
+ *
+ * Adaptive thinking (`thinking: { type: "adaptive" }` + `effort`) is the
+ * recommended thinking mode on Claude 4.6+ models and the only mode on
+ * Opus 4.7/4.8; on Claude Fable 5 thinking is always on and adaptive is the
+ * only mode. The CUA path (AnthropicCUAClient) sets the raw Messages API
+ * fields directly; the hybrid/DOM path goes through @ai-sdk/anthropic, which
+ * maps the typed provider options below to the same API fields.
+ *
+ * Fable 5 may decline a turn (stop_reason "refusal"). The `fallbacks`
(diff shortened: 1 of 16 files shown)


miguelg719:
@cubic-dev-ai

tkattkat (approved):

packages/core/lib/v3/llm/anthropicOptions.ts
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * Shape accepted by the AI SDK's per-provider `providerOptions` maps. The
+ * helpers below build values as typed `AnthropicProviderOptions` (so field
+ * names stay checked) and return them under this JSON-compatible alias so
+ * call sites need no casts.
+ */
+export type AnthropicAgentProviderOptions = Record<string, JSONValue>;
+
+export const ANTHROPIC_FABLE_5_MODEL_ID = "claude-fable-5" as const;
```
shrey150: atp shouldn't we just refer to it as "claude-fable-5" in code instead of abstracting out into another variable? also makes the array in `ADAPTIVE_THINKING_MODEL_BASES` more cluttered / less readable

packages/core/lib/v3/llm/anthropicOptions.ts
```diff
@@ -0,0 +1,141 @@
...
+  ANTHROPIC_FABLE_5_MODEL_ID,
+]);
+
+/** True for Anthropic models that support adaptive thinking. */
+export function isAdaptiveThinkingAnthropicModel(modelId: string): boolean {
+  return ADAPTIVE_THINKING_MODEL_BASES.has(stripModelProvider(modelId));
+}
+
+export function isAnthropicFable5Model(modelId: string): boolean {
+  return stripModelProvider(modelId) === ANTHROPIC_FABLE_5_MODEL_ID;
```
shrey150: also same here - we could just refer to it as the string literal, and likely even just remove this utility function

packages/core/lib/v3/llm/anthropicOptions.ts
```diff
@@ -0,0 +1,141 @@
...
+}
+
+/**
+ * True for models that reject forced tool use
+ * (`tool_choice: { type: "tool" }`). Forced tool choice is incompatible with
+ * active extended thinking, and on Fable 5 thinking is always on — so the
+ * rejection is a certainty there, not a transient quirk.
+ */
+export function rejectsForcedToolUse(modelId: string): boolean {
+  return isAnthropicFable5Model(modelId);
```
shrey150, line 69: if there were more models / conditions in which to reject forced tool use, I'd keep this - but we could probably just note this with a comment for now - your call

shrey150 (approved):
Pre-approving, left some comments if you choose to address - non-blocking

packages/core/lib/v3/llm/anthropicOptions.ts
miguelg719: leaving, this is the first of the anthropic line of models that will reject

PR #2246, opened 2026-06-12
[STG-2275] feat(cli): default screenshot to file output (--base64 for legacy stdout)

shrey150:
## Summary

Bare `browse screenshot` now writes a file by default — `screenshot-<yyyymmdd-hhmmss>.<type>` in the current directory, never overwriting (atomic collision counter) — and prints the small `{ "saved": "<path>" }` JSON. A new `--base64` flag preserves the legacy stdout contract (`{ "base64": "..." }`); it is mutually exclusive with `--path`. Explicit `--path` behavior is unchanged.

Linear: [STG-2275](https://linear.app/browserbase/issue/STG-2275/default-browse-screenshot-to-file-output-with-base64-legacy-flag)

## Impact if merged

screenshot is one of the two commands in the runaway agent retry loops (the loop population generates 92.3% of all CLI telemetry) and is used by 1,337 users/30d (89.9% success last-7d). Every bare invocation today prints ~22KB of base64 JSON directly into the calling agent's context window — a per-call token tax on exactly the ICP population (claude-code/codex agents driving the CLI; agent-tagged usage is 90%-successful and growing). Defaulting to a file write makes the common case agent-safe at zero cost to `--path` users; `--base64` preserves the old contract for scripts. Stdout contract change for bare invocations is called out in the changeset with a `--base64` migration note (minor bump — the bare-invocation stdout contract changes; `browse` is `<1.0.0` so this is advisory).

## Implementation notes

- **Breaking change (stdout contract):** bare `browse screenshot` now prints `{ "saved": "<path>" }` instead of `{ "base64": "..." }`. **Migration:** pass `--base64` to restore the old output.
- Command-layer only: the driver handler (`runtime.ts`) already supported both branches (`{ saved }` when a path is given, `{ base64 }` otherwise), so the change is confined to `src/commands/screenshot.ts`.
- Default filename respects `--type` (`.jpeg` vs `.png`) and resolves against the invoking shell's cwd (absolute path passed to the driver so a daemon with a different cwd can't misplace the file).
- The default filename is **atomically reserved** via exclusive create (`openSync(path, "wx")`), advancing a `-2`, `-3`, ... counter on `EEXIST` — concurrent same-second invocations can never claim the same file (addresses cubic's race-condition review). If the command fails afterward, the empty placeholder is removed best-effort.
- `--base64` is mutually exclusive with `--path` via oclif's `exclusive` option.
- Changeset: `browse` **minor** (per review) — the bare-invocation stdout contract change is called out in the changeset body with the `--base64` migration note.
- Bundled skill doc: `skills/browse/SKILL.md` screenshot snippet updated in this PR (moved from #2245 per review) to document the new default: bare saves a file, `--path` chooses it, `--base64` is the legacy stdout form.

## E2E Test Matrix

All rows ran against the local build (`pnpm build` in `packages/cli`, invoked as `node bin/run.js`) from a `<scratch dir>`.

| --- | --- | --- |
| `node bin/run.js open https://example.com` | `{ "title": "Example Domain", "url": "https://example.com/", ... }`, exit 0 | Session setup for the runs below; proves the local build is functional end-to-end. |
| `node bin/run.js screenshot` (bare) | `{ "saved": "<scratch dir>/screenshot-20260612-123732.png" }`, exit 0; `ls -la` shows the file at 15,988 bytes; stdout is the small JSON only | Proves the new default: file written to cwd with timestamped name, no base64 on stdout. |
| `node bin/run.js screenshot --base64 \| head -c 200` | `{ "base64": "iVBORw0KGgoAAAANSUhEUgAABQgAAALH..."` (PNG magic in base64), exit 0 | Proves the legacy stdout contract is fully preserved behind `--base64`. |
| `node bin/run.js screenshot --path /tmp/custom.png` | `{ "saved": "/tmp/custom.png" }`, exit 0; file exists at 15,238 bytes | Proves explicit `--path` behavior is unchanged. |
| Two **concurrent** bare runs (same second, shell `&` + `wait`) | `{ "saved": ".../screenshot-20260612-123741.png" }` and `{ "saved": ".../screenshot-20260612-123741-2.png" }`; both files present and non-empty (4,202 bytes each) | Proves the atomic no-overwrite reservation under true same-second concurrency — the exact race cubic flagged. |
| `node bin/run.js screenshot --cdp http://127.0.0.1:1` (forced failure) | `TypeError: fetch failed`, nonzero exit; directory left empty — no placeholder file | Proves failed runs do not leave empty `screenshot-*.png` placeholders behind. |
| `node bin/run.js screenshot --base64 --path /tmp/x.png` | `Error: --path=/tmp/x.png cannot also be provided when using --base64`, exit 2 | Proves the oclif mutual exclusion works. |
| `node bin/run.js stop` | `{ "stopped": true, "session": "default" }`, exit 0 | Clean session teardown. |
| `pnpm test` (packages/cli) | `Test Files 15 passed (15), Tests 213 passed (213)` — re-run after the race fix | Supporting: no regressions in the existing CLI suite (includes the screenshot `--help` surface test). |
| `pnpm lint` (packages/cli) | tsc/eslint/prettier clean | Supporting only. |

### DIFF
--- a/.changeset/screenshot-default-file.md
+++ b/.changeset/screenshot-default-file.md
@@ -1,5 +1,5 @@
 ---
-"browse": patch
+"browse": minor
 ---
 
 `browse screenshot` now writes a file by default instead of printing base64 to stdout. Bare invocations save to `screenshot-<yyyymmdd-hhmmss>.<type>` in the current directory (with a collision counter instead of overwriting) and print `{ "saved": "<path>" }`. A new `--base64` flag preserves the legacy behavior of printing `{ "base64": "..." }` to stdout; it is mutually exclusive with `--path`. `--path` behavior is unchanged.
--- a/.changeset/screenshot-default-file.md
+++ b/.changeset/screenshot-default-file.md
@@ -1,7 +1,7 @@
 ---
-"browse": minor
+"browse": patch
 ---
 
 `browse screenshot` now writes a file by default instead of printing base64 to stdout. Bare invocations save to `screenshot-<yyyymmdd-hhmmss>.<type>` in the current directory (with a collision counter instead of overwriting) and print `{ "saved": "<path>" }`. A new `--base64` flag preserves the legacy behavior of printing `{ "base64": "..." }` to stdout; it is mutually exclusive with `--path`. `--path` behavior is unchanged.
 
-This is a minor (not patch) bump because bare `browse screenshot` changes its stdout contract from `{ base64 }` to `{ saved }` — scripts that parsed the base64 output must now pass `--base64`.
+Note for scripts that parsed the bare-invocation base64 output: pass `--base64` to keep the old stdout contract.
--- a/.changeset/screenshot-default-file.md
+++ b/.changeset/screenshot-default-file.md
@@ -1,0 +1,7 @@
+---
+"browse": minor
+---
+
+`browse screenshot` now writes a file by default instead of printing base64 to stdout. Bare invocations save to `screenshot-<yyyymmdd-hhmmss>.<type>` in the current directory (with a collision counter instead of overwriting) and print `{ "saved": "<path>" }`. A new `--base64` flag preserves the legacy behavior of printing `{ "base64": "..." }` to stdout; it is mutually exclusive with `--path`. `--path` behavior is unchanged.
+
+This is a minor (not patch) bump because bare `browse screenshot` changes its stdout contract from `{ base64 }` to `{ saved }` — scripts that parsed the base64 output must now pass `--base64`.
--- a/packages/cli/skills/browse/SKILL.md
+++ b/packages/cli/skills/browse/SKILL.md
@@ -139,8 +139,9 @@ browse get html "#main"
 browse get value "#email"
 browse get markdown body                   # page/element content as markdown
 browse eval "document.title"              # run JavaScript in the active page
-browse screenshot                         # print base64 JSON
-browse screenshot --path page.png
+browse screenshot                         # saves screenshot-<timestamp>.png, prints { "saved": "<path>" }
+browse screenshot --path page.png         # choose the output path
+browse screenshot --base64                # legacy: print base64 JSON to stdout (avoid in agent loops)
 ```
 
 Interaction:
--- a/packages/cli/src/commands/screenshot.ts
+++ b/packages/cli/src/commands/screenshot.ts
@@ -48,7 +48,7 @@ export default class Screenshot extends BrowseCommand {
     path: Flags.string({
       char: "p",
       description:
-        "Write the screenshot to this file. Defaults to screenshot-<timestamp>.<type> in the current directory.",
+        "Write the screenshot to this file. Defaults to screenshot-<timestamp>.png (or .jpeg with --type jpeg) in the current directory.",
       helpValue: "<path>",
     }),
     quality: Flags.integer({
@@ -63,10 +63,7 @@ export default class Screenshot extends BrowseCommand {
 
   async run(): Promise<void> {
     const { flags } = await this.parse(Screenshot);
-    const defaultPath =
-      flags.path || flags.base64
-        ? undefined
-        : reserveDefaultScreenshotPath(flags.type);
+    const defaultPath = getDefaultPathFromFlags(flags);
     try {
       await runDriverCommandFromFlags(
         "screenshot",
@@ -88,10 +85,30 @@ export default class Screenshot extends BrowseCommand {
   }
 }
 
+// Generous upper bound on filename-collision retries; far beyond any real
+// same-second burst, it just guarantees the loop below always terminates.
+const MAX_RESERVE_ATTEMPTS = 1000;
+
+/**
+ * Resolves the file the screenshot should be written to, or undefined when the
+ * driver should return base64 (explicit --path is handled separately; --base64
+ * opts out of a file entirely).
+ */
+function getDefaultPathFromFlags(flags: {
+  path?: string;
+  base64?: boolean;
+  type?: string;
+}): string | undefined {
+  if (flags.path || flags.base64) return undefined;
+  return reserveDefaultScreenshotPath(flags.type);
+}
+
 /**
  * Picks the next free screenshot-<timestamp>[-<counter>].<type> name in the
- * current directory and atomically reserves it with an exclusive create, so
- * concurrent invocations can never claim (and overwrite) the same file.
+ * current directory and reserves it. "Reserve" = create the file with an
+ * exclusive open (`wx` → O_CREAT|O_EXCL), which atomically fails with EEXIST if
+ * the name already exists, so two concurrent runs can never claim the same
+ * file. On EEXIST we advance the counter and try the next name.
  */
 function reserveDefaultScreenshotPath(type: string | undefined): string {
   const now = new Date();
@@ -100,21 +117,30 @@ function reserveDefaultScreenshotPath(type: string | undefined): string {
     `${now.getFullYear()}${pad(now.getMonth() + 1)}${pad(now.getDate())}` +
     `-${pad(now.getHours())}${pad(now.getMinutes())}${pad(now.getSeconds())}`;
   const extension = type === "jpeg" ? "jpeg" : "png";
-  for (let counter = 1; ; counter += 1) {
+  for (let counter = 1; counter <= MAX_RESERVE_ATTEMPTS; counter += 1) {
     const suffix = counter === 1 ? "" : `-${counter}`;
     const candidate = resolve(`screenshot-${stamp}${suffix}.${extension}`);
     try {
+      // Exclusive create reserves the name; close immediately since the driver
+      // (re)writes the file. The empty placeholder is cleaned up on failure.
       closeSync(openSync(candidate, "wx"));
       return candidate;
     } catch (error) {
       if ((error as NodeJS.ErrnoException).code !== "EEXIST") throw error;
     }
   }
+  throw new Error(
+    `Could not reserve a screenshot filename after ${MAX_RESERVE_ATTEMPTS} attempts; pass --path to choose one.`,
+  );
 }
 
+// Removes the reserved placeholder when the screenshot failed before the driver
+// wrote to it. `path` is always a file we created via openSync above, so the
+// isFile guard is just defensive against an unexpected directory/symlink.
 function removeIfEmpty(path: string): void {
   try {
-    if (statSync(path).size === 0) unlinkSync(path);
+    const stats = statSync(path);
+    if (stats.isFile() && stats.size === 0) unlinkSync(path);
   } catch {
     // Best effort: leave the placeholder behind rather than mask the error.
   }
--- a/packages/cli/src/commands/screenshot.ts
+++ b/packages/cli/src/commands/screenshot.ts
@@ -1,4 +1,4 @@
-import { existsSync } from "node:fs";
+import { closeSync, openSync, statSync, unlinkSync } from "node:fs";
 import { resolve } from "node:path";
 
 import { Flags } from "@oclif/core";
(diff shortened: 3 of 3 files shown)


packages/cli/src/commands/screenshot.ts
```diff
@@ -53,18 +63,35 @@ export default class Screenshot extends BrowseCommand {
...
+
+function defaultScreenshotPath(type: string | undefined): string {
+  const now = new Date();
+  const pad = (value: number) => String(value).padStart(2, "0");
+  const stamp =
+    `${now.getFullYear()}${pad(now.getMonth() + 1)}${pad(now.getDate())}` +
+    `-${pad(now.getHours())}${pad(now.getMinutes())}${pad(now.getSeconds())}`;
+  const extension = type === "jpeg" ? "jpeg" : "png";
+  let candidate = resolve(`screenshot-${stamp}.${extension}`);
+  for (let counter = 2; existsSync(candidate); counter += 1) {
```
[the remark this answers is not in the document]
shrey150: Addressed in [692f2d623](https://github.com/browserbase/stagehand/commit/692f2d623) — replaced the `existsSync` check with an atomic exclusive-create (`openSync(path, "wx")`) reservation, so concurrent invocations can never claim the same default filename; the loop advances the counter on `EEXIST`. If the screenshot command later fails, the empty placeholder is removed (best effort). Verified empirically: two concurrent bare runs in the same second produced `screenshot-20260612-123741.png` and `screenshot-20260612-123741-2.png` (both non-empty), and a forced failure (`--cdp` to a dead endpoint) left no placeholder behind.

.changeset/screenshot-default-file.md
```diff
@@ -0,0 +1,7 @@
+---
+"browse": minor
```
shrey150, line 2: should be patch

shrey150: Done in [1552300ef](https://github.com/browserbase/stagehand/commit/1552300efe3460559a44fee65c3ee814e30c804e) — changeset is now `patch` (kept the migration note for scripts that parsed the bare base64 output). Same commit also picks up the bundled SKILL.md screenshot snippet from #2245, rewritten for the new default per your comment there.

packages/cli/src/commands/screenshot.ts
```diff
@@ -53,18 +63,59 @@ export default class Screenshot extends BrowseCommand {
...
+    }
+  }
+}
+
+/**
+ * Picks the next free screenshot-<timestamp>[-<counter>].<type> name in the
+ * current directory and atomically reserves it with an exclusive create, so
+ * concurrent invocations can never claim (and overwrite) the same file.
+ */
+function reserveDefaultScreenshotPath(type: string | undefined): string {
```
shrey150, line 113: is there any utility to do this instead of us rolling it ourselves? only if it makes sense

shrey150: Looked at the realistic candidates before rolling it — none actually covers this combination (human-readable timestamped name, in cwd, atomically reserved):

- **Node built-ins**: nothing — `fs.mkdtemp` is directories-in-tmpdir only; there's no mkstemp-style API for named patterns.
- **`tempy`**: tempdir-only semantics (we want a user-visible artifact in cwd), and it generates random names — loses the readable timestamp.
- **`unique-filename` (npm's)**: returns `prefix-<hash>` — also loses the timestamp, and it only *generates* a name, no atomic reservation, so the cubic-flagged race would come right back.
- **`write-file-atomic`**: solves atomic *writes* (tmp+rename), not unique *naming* — it would happily overwrite an existing same-named file.

Each one either misses a requirement or still needs the `wx`/EEXIST loop on top, while adding a dependency to a deliberately lean CLI (and another supply-chain surface for Socket to watch). The hand-rolled version is 17 lines of stdlib idiom (`openSync(path, "wx")` + counter on EEXIST), unit-tested incl. the concurrent case — so my call is it doesn't make sense here; keeping it.

.changeset/screenshot-default-file.md
```diff
@@ -0,0 +1,7 @@
+---
+"browse": patch
```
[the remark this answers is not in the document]
shrey150: Maintainer call, explicitly reviewed: @shrey150 requested patch in the thread above (repo convention: changesets default to patch; minor is reserved for stronger reasons than this). The contract change is called out in the changeset body with the `--base64` migration note, the PR title carries the behavior change, and the release notes will surface it. Leaving as patch.

packages/cli/skills/browse/SKILL.md
```diff
@@ -139,8 +139,9 @@ browse get html "#main"
 browse get value "#email"
 browse get markdown body                   # page/element content as markdown
 browse eval "document.title"              # run JavaScript in the active page
-browse screenshot                         # print base64 JSON
-browse screenshot --path page.png
+browse screenshot                         # saves screenshot-<timestamp>.png, prints { "saved": "<path>" }
```
ajmcquilkin, line 142: Do we need to communicate this breaking change to anyone other than through the changeset? Suppose we can rely on the package being `<1.0.0`

packages/cli/src/commands/screenshot.ts
```diff
@@ -38,7 +48,7 @@ export default class Screenshot extends BrowseCommand {
     path: Flags.string({
       char: "p",
       description:
-        "Write the screenshot to a file. Without this flag, base64 is printed.",
+        "Write the screenshot to this file. Defaults to screenshot-<timestamp>.<type> in the current directory.",
```
ajmcquilkin:
```suggestion
        "Write the screenshot to this file. Defaults to 'screenshot-<timestamp>.<filetype>' in the current directory.",
```
What does the filetype default to?

packages/cli/src/commands/screenshot.ts
```diff
@@ -53,18 +63,59 @@ export default class Screenshot extends BrowseCommand {
...
-        caret: flags.caret,
-        clip: parseClip(flags.clip),
-        fullPage: flags["full-page"],
-        path: flags.path,
-        quality: flags.quality,
-        type: flags.type,
-      },
-      flags,
-    );
+    const defaultPath =
```
ajmcquilkin: Nit: should this be extracted to a `getDefaultPathFromFlags` method?

packages/cli/src/commands/screenshot.ts
```diff
@@ -53,18 +63,59 @@ export default class Screenshot extends BrowseCommand {
...
+ * concurrent invocations can never claim (and overwrite) the same file.
+ */
+function reserveDefaultScreenshotPath(type: string | undefined): string {
+  const now = new Date();
+  const pad = (value: number) => String(value).padStart(2, "0");
+  const stamp =
+    `${now.getFullYear()}${pad(now.getMonth() + 1)}${pad(now.getDate())}` +
+    `-${pad(now.getHours())}${pad(now.getMinutes())}${pad(now.getSeconds())}`;
+  const extension = type === "jpeg" ? "jpeg" : "png";
+  for (let counter = 1; ; counter += 1) {
```
ajmcquilkin: This is a small thing, but we should have a max number of attempts here in the off chance that the code below has a bug and doesn't term the loop (can make it absurdly high like 100 or 1000 attempts)

packages/cli/src/commands/screenshot.ts
```diff
@@ -53,18 +63,59 @@ export default class Screenshot extends BrowseCommand {
...
+    const suffix = counter === 1 ? "" : `-${counter}`;
+    const candidate = resolve(`screenshot-${stamp}${suffix}.${extension}`);
+    try {
+      closeSync(openSync(candidate, "wx"));
```
ajmcquilkin, line 126: What does this do? Off rip this seems like "close a file we just opened." If this does more we should detail in a comment

packages/cli/src/commands/screenshot.ts
```diff
@@ -53,18 +63,59 @@ export default class Screenshot extends BrowseCommand {
...
+    } catch (error) {
+      if (defaultPath) removeIfEmpty(defaultPath);
+      throw error;
+    }
+  }
+}
+
+/**
+ * Picks the next free screenshot-<timestamp>[-<counter>].<type> name in the
+ * current directory and atomically reserves it with an exclusive create, so
```
ajmcquilkin: What mechanism does it use to "reserve"? Not understanding from the code

packages/cli/src/commands/screenshot.ts
```diff
@@ -53,18 +63,59 @@ export default class Screenshot extends BrowseCommand {
...
+      return candidate;
+    } catch (error) {
+      if ((error as NodeJS.ErrnoException).code !== "EEXIST") throw error;
+    }
+  }
+}
+
+function removeIfEmpty(path: string): void {
+  try {
+    if (statSync(path).size === 0) unlinkSync(path);
```
ajmcquilkin: Will `path` always be a file? Seems like weird stuff happens when calling `statSync(dir).size` (either `0` or `4096`)

.changeset/screenshot-default-file.md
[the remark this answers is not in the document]
shrey150: I think this should be minor

packages/cli/skills/browse/SKILL.md
shrey150: agreed, I'll bump to minor changeset

shrey150: Done in [81159c517](https://github.com/browserbase/stagehand/commit/81159c517) — bumped the changeset to **minor**. (`<1.0.0` makes it advisory either way, but minor is the honest signal for the stdout contract change.)

packages/cli/src/commands/screenshot.ts
shrey150: Filetype defaults to **png** (or jpeg with `--type jpeg`). Clarified the help text in [81159c517](https://github.com/browserbase/stagehand/commit/81159c517): "Defaults to screenshot-<timestamp>.png (or .jpeg with --type jpeg) in the current directory."

packages/cli/src/commands/screenshot.ts
shrey150: Extracted to `getDefaultPathFromFlags(flags)` in [81159c517](https://github.com/browserbase/stagehand/commit/81159c517).

packages/cli/src/commands/screenshot.ts
shrey150: Added `MAX_RESERVE_ATTEMPTS = 1000` in [81159c517](https://github.com/browserbase/stagehand/commit/81159c517); the loop now throws a clear error ("pass --path to choose one") rather than spinning if the counter logic ever regressed.

packages/cli/src/commands/screenshot.ts
shrey150: Right — it closes the fd we just opened. We don't write through it; the exclusive `openSync(..., "wx")` is only there to *reserve* the name, and the driver (re)writes the file afterward. Added a comment saying exactly that in [81159c517](https://github.com/browserbase/stagehand/commit/81159c517).

packages/cli/src/commands/screenshot.ts
shrey150: The mechanism is `openSync(candidate, "wx")` — `wx` = O_CREAT|O_EXCL, which atomically fails with EEXIST if the name already exists. So whichever concurrent run creates the file first wins the name; the loser catches EEXIST and advances the counter. Spelled this out in a comment in [81159c517](https://github.com/browserbase/stagehand/commit/81159c517).

packages/cli/src/commands/screenshot.ts
shrey150: In practice `path` is always the placeholder file we created via `openSync` above, so it's always a regular file. But you're right that `statSync(dir).size` would be misleading — guarded it with `stats.isFile()` in [81159c517](https://github.com/browserbase/stagehand/commit/81159c517) so a directory/symlink can never be unlinked.

packages/cli/src/commands/screenshot.ts
shrey150: is there no common pattern / utility function from a library for generating unique, sequential screenshot file names? is the locking strictly needed? how would you think through the cost-benefit of finding a package to do this? how do other CLI packages that take screenshots / save unique files do this?

packages/cli/src/commands/screenshot.ts
```diff
@@ -53,18 +63,85 @@ export default class Screenshot extends BrowseCommand {
...
+  }
+  throw new Error(
+    `Could not reserve a screenshot filename after ${MAX_RESERVE_ATTEMPTS} attempts; pass --path to choose one.`,
+  );
+}
+
+// Removes the reserved placeholder when the screenshot failed before the driver
+// wrote to it. `path` is always a file we created via openSync above, so the
+// isFile guard is just defensive against an unexpected directory/symlink.
+function removeIfEmpty(path: string): void {
```
shrey150, line 140: this feels like there's a utility function for it

shrey150: There isn't a stdlib or repo utility for "delete a file if it's empty" — it's inherently stat + unlink, and we don't depend on fs-extra (and even that has no unlink-if-empty helper). So no clean util to lean on here.

But the real question is whether this function needs to exist at all, and it only does as the cleanup half of the atomic `wx` reservation: we pre-create a 0-byte placeholder to win the filename race (the concurrent-overwrite issue cubic flagged), so if the screenshot then fails before the driver writes, we'd leave a stray empty `screenshot-<ts>.png` behind. `removeIfEmpty` deletes that placeholder.

Two ways to simplify if the complexity isn't worth it:
1. **Drop `removeIfEmpty`** and accept that a *failed* screenshot leaves a 0-byte file in cwd (rare, minor litter). −7 lines.
2. **Drop the whole reservation** (placeholder + loop + cleanup, ~20 lines) by making the name collision-proof without touching disk — e.g. a short random suffix `screenshot-<ts>-<rand>.png`. Removes the race entirely, but the filename is less clean.

I'd lean toward keeping it as-is (zero litter, clean names, and it's tested), but happy to take either simplification — your call.

packages/cli/src/commands/screenshot.ts
shrey150: Good push — I dug into this, and the honest conclusion is the atomic locking is *more* rigor than any comparable tool applies, and we can simplify.

**Is there a library / common pattern?** No package fits. The candidates are all tmpdir-random-name generators — `tempy`, `tmp`, `unique-filename` — which produce hashed names in the OS temp dir, not a human-readable timestamped file in cwd. None does "timestamped, sequential, collision-handled name in the working directory"; that's app-specific. So a dep would be net-negative (wrong UX, new supply-chain surface, and it still wouldn't give us the naming we want).

**How do peer tools do it?**
- **macOS `screencapture`**: `Screenshot <date> at <time>.png`, appends ` 2`/` 3` on a same-second collision. Non-atomic (check-then-write).
- **`scrot` / `flameshot` (Linux)**: `strftime` timestamp names; scrot just overwrites on collision. Non-atomic.
- **Playwright / Puppeteer**: the caller supplies the path; the test runner auto-names as `<title>-<index>.png` within a single process (no cross-process locking needed).

The shared pattern is **timestamp + counter, non-atomic** — everyone accepts the tiny TOCTOU window because two captures in the *same second, same directory* is a non-event for an interactive tool, and the worst case is one screenshot overwriting another.

**Is the locking strictly needed?** No. I added the atomic `wx` reservation to satisfy cubic's race flag, but the realistic trigger is two parallel `browse screenshot` calls landing in the same second in the same cwd (agent parallelism) — and even then the cost is one overwritten screenshot, which is exactly what `screencapture`/`scrot` tolerate.

**Recommendation:** simplify to the industry norm — timestamp + counter, non-atomic. That drops the `wx` placeholder, the reservation loop, *and* `removeIfEmpty` (your other comment) — about 20 lines gone, and the "is there a util for removeIfEmpty" question disappears with it. The only behavior change is that in the astronomically-rare same-second-same-cwd collision we might overwrite instead of atomically dodging — a tradeoff every comparable tool makes.

Want me to push that simplification? I'd keep the `--path`/`--base64` surface identical; only the internal default-name logic shrinks.

ajmcquilkin (approved):

PR #2248, opened 2026-06-12
[STG-2277] fix(cli): actionable driver errors, result codes, and init-failure backoff

shrey150:
## Summary

Makes browse driver (browser session) failures actionable, classified, and self-correcting. Today an invalid `BROWSERBASE_API_KEY` surfaces a bare `Error: 401 Unauthorized` with no remediation, the 5s init-failure cache makes every retry instant and identical, and most driver failures reach telemetry as `unexpected`.

Linear: [STG-2277](https://linear.app/browserbase/issue/STG-2277/make-browse-driver-errors-actionable-with-result-codes-and-init)

## Impact if merged

This targets the browse CLI's largest failure mode by volume and by user pain. A small set of installs stuck in get/screenshot retry loops generates the overwhelming majority of all CLI telemetry, including a meaningful share from tagged claude-code and codex agents on current versions — exactly the target ICP (coding agents driving browsers). Root cause (smoke-tested): any `BROWSERBASE_API_KEY` forces remote mode; an invalid key surfaces a bare `Error: 401 Unauthorized` with no remediation, and the 5s init-failure cache makes every retry instant, so agents can't self-correct and loop forever. Separately, many users hit missing_api_key/auth_401, and `open` — whose failures are mostly unclassifiable today (result_code `unexpected`) — gates activation: a failed first command sharply cuts the odds a user ever activates, and we see significantly higher retention among users who succeed early. This PR makes auth/driver failures actionable (agents recover in one turn) and classified (we can finally measure why open fails).

## Implementation notes

- **Remote init classification** (`remote.ts`): new `classifyRemoteInitError()` duck-types the SDK error's `status` — 401 → `remote_auth_401` (invalid-key message with settings link, `--local`, `browse doctor`), 403 → `remote_auth_403` (permissions/plan wording, same escape hatches), other → `remote_session_create_failed` (original message preserved + `browse doctor` pointer). Wired through the `RemoteCapability` interface so local-only builds compile.
- **Chrome-not-found** (`session-manager.ts`): chrome-launcher's `ERR_LAUNCHER_NOT_INSTALLED` / `ERR_LAUNCHER_PATH_NOT_SET` failures in managed-local mode get install/`--cdp`/remote guidance instead of leaking launcher internals.
- **Init-failure backoff**: cached init failures now back off exponentially — `min(5s * 2^(n-1), 5min)` — reset on success and `close()`. After ≥3 consecutive failures the cached message gains a `(failing repeatedly — fix BROWSERBASE_API_KEY, use --local, or run browse doctor)` suffix (deduped on rethrow).
- **Result codes over the daemon protocol**: `ErrorResponseSchema` gains optional `code`/`httpStatus` (backward compatible — old daemons omit them); the daemon's `formatError` surfaces them from typed `DriverError`s; the client rethrows as `CommandFailure` with `resultCode`/`httpStatus` so the existing #2210 telemetry plumbing records them. Client-side fail sites tagged: `daemon_lock_timeout`, `daemon_unresponsive`, `daemon_socket_timeout`, `daemon_spawn_failed`. Already-authored driver errors tagged: `stale_ref` (unknown ref), `no_active_page`.
- **Local-only build contract preserved**: remediation strings that mention `BROWSERBASE_API_KEY` live behind the remote capability (`driverInitHints()`), so the `build:local-only` artifact stays key-free (guarded by the existing `local-only-build.test.ts`, which caught the first draft).

## E2E Test Matrix

| --- | --- | --- |
| `BROWSERBASE_API_KEY=bb_invalid_test <local build> get url` | `Browserbase rejected your BROWSERBASE_API_KEY (401 Unauthorized). A set key makes browse default to remote mode. Check the key at https://browserbase.com/settings, run without one using --local (browse open <url> --local), or diagnose with browse doctor.` exit=1 | Proves the new 401 classification flows daemon → protocol → client → stderr end-to-end against the real Browserbase API. |
| Same command 4x rapidly (cached failure window) | Identical actionable message each time, ~400ms per run (no remote round-trip) | Proves cached failures keep the actionable message and stay instant; does not by itself prove backoff growth. |
| Same command after 6s, then after 11s more (real failures #2, #3) | Message gains ` (failing repeatedly — fix BROWSERBASE_API_KEY, use --local, or run browse doctor)` suffix, exactly once, exit=1 | Proves the ≥3-consecutive-failures hint and suffix dedupe on the live failure path. |
| Valid key: `open https://example.com` → `get title` → `stop` | `"mode": "remote" ... "title": "Example Domain"`, then `{"title": "Example Domain"}`, then `{"stopped": true}` | Proves the remote happy path is unchanged (no regression in outputs or exit codes). |
| `env -u BROWSERBASE_API_KEY <local build> open https://example.com --local` → `get url` | `"mode": "managed-local" ... "url": "https://example.com/"`, then `{"url": "https://example.com/"}` exit=0 | Proves keyless managed-local mode is unaffected. |
| `get text @9-99` on the local session | `Unknown ref "9-99" - run browse snapshot first to populate refs (have 0 refs).` exit=1 | Proves the stale-ref message is unchanged while now carrying `stale_ref` through the protocol (round-trip unit-tested). |
| `browse doctor` with and without key | `Status: ok` in both; `target remote` with key, `target managed-local` without | Proves doctor behavior unchanged. |
| `pnpm build` + `pnpm lint` (prettier, eslint, tsc) | All pass | Supporting only. |
| `pnpm test:cli` | 16 files / 228 tests pass, incl. new `driver-errors.test.ts` (classification, backoff schedule, chrome-not-found detection, protocol round-trip, key-free local-only hints) and the `local-only-build` artifact guard | Supporting; covers mappings and the local-only security contract not exercised by live smokes. |

---
## Summary by cubic
Makes browse driver failures actionable and self-correcting with classified result codes and exponential init backoff. Addresses Linear STG-2277 by giving clear fixes for bad `BROWSERBASE_API_KEY`, missing Chrome/Chromium, and daemon issues, while improving telemetry.

- **New Features**
  - Classify remote init errors into actionable messages with codes: `remote_auth_401`, `remote_auth_403`, `remote_session_create_failed` (with links to settings, `--local`, and `browse doctor`).
  - Add error result codes to the daemon protocol (`code`, `httpStatus`) and propagate to the client for telemetry.
  - Exponential backoff for cached init failures (5s doubling, capped at 1 minute) with a “failing repeatedly” hint after 3 failures.
  - Tag common failures with stable codes: `daemon_lock_timeout`, `daemon_unresponsive`, `daemon_socket_timeout`, `daemon_spawn_failed`, `stale_ref`, `no_active_page`, `no_chrome_found`.
  - Use `http-status-codes` for status mapping and extract chrome-launcher error codes to a constant (no behavior change).

- **Bug Fixes**
  - Chrome-not-found now gives Chromium-first guidance: Linux `apt install chromium`; macOS `brew install --cask google-chrome` or set `CHROME_PATH` for Chromium, plus `--cdp` or remote as options.
  - Keep the local-only build key-free by moving `BROWSERBASE_API_KEY` remediation strings behind the remote capability.

<sup>Written for commit b7a3f7e1fb367426e546836a00a0abb29d343cfc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2248?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/cli/src/lib/driver/remote.ts
```diff
@@ -38,6 +40,51 @@ export function remoteStagehandOptions(): StagehandConstructorOptions {
...
+    code: "remote_session_create_failed",
+    ...(httpStatus !== undefined ? { httpStatus } : {}),
+    message: `Failed to start a remote (Browserbase) session: ${original}\nRun browse doctor to diagnose remote connectivity.`,
+  };
+}
+
+export function driverInitHints(): DriverInitHints {
+  return {
+    chromeNotFound:
+      "Chrome was not found on this machine. Install Chrome (https://www.google.com/chrome/), attach to a running browser with --cdp <port>, or set BROWSERBASE_API_KEY to use a remote browser.",
```
shrey150: shouldn't we tell them to use brew install for installing chrome, or tell them to install chromium?

packages/cli/src/lib/driver/remote.ts
```diff
@@ -38,6 +40,51 @@ export function remoteStagehandOptions(): StagehandConstructorOptions {
...
+ * stable result code. Browserbase SDK errors carry an HTTP `status`.
+ */
+export function classifyRemoteInitError(
+  error: unknown,
+): RemoteInitErrorClassification {
+  const status = (error as { status?: unknown } | null | undefined)?.status;
+  const httpStatus = typeof status === "number" ? status : undefined;
+  const original = error instanceof Error ? error.message : String(error);
+
+  if (httpStatus === 401) {
```
shrey150: didn't another downstream PR handle this [here](https://github.com/browserbase/stagehand/pull/2245)? or is this just a user-facing descriptive error

packages/cli/src/lib/driver/remote.disabled.ts
```diff
@@ -25,6 +27,27 @@ export function remoteStagehandOptions(): StagehandConstructorOptions {
...
+    code: "remote_session_create_failed",
+    message: error instanceof Error ? error.message : String(error),
+  };
+}
+
+export function driverInitHints(): DriverInitHints {
+  // Key-free variants: a local-only artifact must not mention the API key.
+  return {
+    chromeNotFound:
+      "Chrome was not found on this machine. Install Chrome (https://www.google.com/chrome/) or attach to a running browser with --cdp <port>.",
```
shrey150: same here, we should remove google.com/chrome and advocate for chromium right

packages/cli/src/lib/driver/remote.ts
shrey150: Addressed in [7a891ffbd](https://github.com/browserbase/stagehand/commit/7a891ffbd) — chromium-first per platform reality: chrome-launcher detects `chromium`/`chromium-browser` binaries on Linux (the dominant failing cohort: 14.5k linux users attempted open in 30d), but on macOS its finder only matches `*Chrome.app` patterns — Chromium there requires `CHROME_PATH`. New message: "Install one (Linux: apt install chromium · macOS: brew install --cask google-chrome, or Chromium with CHROME_PATH set)…". The google.com/chrome URL is gone.

packages/cli/src/lib/driver/remote.disabled.ts
shrey150: Addressed in [7a891ffbd](https://github.com/browserbase/stagehand/commit/7a891ffbd) — same change applied to the key-free local-only variant (minus the BROWSERBASE_API_KEY clause, which the local-only artifact guard forbids). URL removed, chromium advocated where the finder can actually detect it; 19/19 driver-error + local-only-guard tests green.

packages/cli/src/lib/driver/remote.ts
shrey150: Different layer, deliberately complementary: #2245 is the *skill docs* — it tells agents how to *react* to a 401 (fix key / unset / --local). This string is the runtime error itself, which is what makes the failure self-describing for everyone who hits it **without** the skill installed — which is most of the death-loop population today (the skill nudge in #2200 exists precisely because they don't have it). The docs reference the behavior, not this string, so there's no duplication to drift.

ajmcquilkin: Nit: we use https://www.npmjs.com/package/http-status-codes in `api` for this

packages/cli/src/lib/driver/session-manager.ts
```diff
@@ -21,15 +22,45 @@ export type DriverContext = Stagehand["context"];
...
+/**
+ * Exponential backoff for cached init failures: 5s, 10s, 20s, ... capped at
+ * 5 minutes. Prevents agents stuck in retry loops from hammering init while
+ * still allowing a quick retry after the first failure.
+ */
+export function initFailureBackoffMs(consecutiveFailures: number): number {
+  const attempt = Math.max(1, consecutiveFailures);
+  return Math.min(
+    INIT_FAILURE_RETRY_MS * 2 ** (attempt - 1),
+    INIT_FAILURE_RETRY_MAX_MS,
```
ajmcquilkin, line 49: Nit: imo 300 seconds is too long for a backoff

packages/cli/src/lib/driver/session-manager.ts
```diff
@@ -21,15 +22,45 @@ export type DriverContext = Stagehand["context"];
...
+  );
+}
+
+/** chrome-launcher signals "no Chrome on this machine" via these codes. */
+export function isChromeNotFoundError(error: unknown): boolean {
+  const code = (error as { code?: unknown } | null | undefined)?.code;
+  if (
+    code === "ERR_LAUNCHER_NOT_INSTALLED" ||
```
ajmcquilkin: Can we import these from a constant from `chrome-launcher` as constants? Hard-codes spook me lol

packages/cli/src/lib/driver/session-manager.ts
```diff
@@ -311,3 +359,35 @@ async function safeTitle(page: DriverPage): Promise<string> {
...
+  if (error instanceof DriverError) return error;
+
+  if (target.kind === "remote") {
+    const { code, httpStatus, message } = (
+      await getRemote()
+    ).classifyRemoteInitError(error);
+    return new DriverError(message, { cause: error, code, httpStatus });
+  }
+
+  if (target.kind === "managed-local" && isChromeNotFoundError(error)) {
```
ajmcquilkin, line 386: Same "const-ify this" nit

packages/cli/tests/driver-errors.test.ts
```diff
@@ -0,0 +1,158 @@
...
+  it("handles non-Error values and non-numeric statuses", () => {
+    const classified = classifyRemoteInitError("boom");
+    expect(classified.code).toBe("remote_session_create_failed");
+    expect(classified.message).toContain("boom");
+  });
+});
+
+describe("initFailureBackoffMs", () => {
+  it("doubles from 5s per consecutive failure", () => {
+    expect(initFailureBackoffMs(1)).toBe(5_000);
```
ajmcquilkin: A nice pattern here is a "table test" which could handle most/all of the `initFailureBackoffMs` tests

```typescript
const tests = [
  { failureCount: 1, expectedBackoffMs: 5_000, description: "..." },
  { failureCount: 2, expectedBackoffMs: 10_000, description: "..." },
  { failureCount: 3, expectedBackoffMs: 20_000, description: "..." },
  { failureCount: 4, expectedBackoffMs: 40_000, description: "..." },
  // ...
];

tests.forEach(({ failureCount, expectedBackoffMs, description }) => {
  it(description, (t) => {
    expect(initFailureBackoffMs(failureCount).toEqual(expectedBackoffMs);
  }); 
});
```

ajmcquilkin (approved):

packages/cli/src/lib/driver/remote.ts
shrey150: Done in [03cd23539](https://github.com/browserbase/stagehand/commit/03cd23539) — added `http-status-codes` (^2.3.0, matching server-v3) and switched to `StatusCodes.UNAUTHORIZED`/`StatusCodes.FORBIDDEN`.

packages/cli/src/lib/driver/session-manager.ts
shrey150: Done in [03cd23539](https://github.com/browserbase/stagehand/commit/03cd23539) — lowered the backoff cap from 300s to 60s (sequence is now 5s, 10s, 20s, 40s, 60s, 60s…). Happy to drop to 30s if you'd prefer.

packages/cli/src/lib/driver/session-manager.ts
shrey150: Done in [03cd23539](https://github.com/browserbase/stagehand/commit/03cd23539) — extracted to a named `CHROME_NOT_FOUND_ERROR_CODES` set. Heads up: chrome-launcher only exposes these as a `const enum` (`LaunchErrorCodes` in its internal utils), which can't be imported directly — const enums are erased at build time and `isolatedModules` forbids cross-module access. So the constant documents chrome-launcher as the source rather than importing it; left a comment saying so.

packages/cli/src/lib/driver/session-manager.ts
shrey150: Covered by the same change in [03cd23539](https://github.com/browserbase/stagehand/commit/03cd23539) — there's only one copy of those codes now (the `CHROME_NOT_FOUND_ERROR_CODES` set), used by `isChromeNotFoundError`. The remaining literals here (`target.kind === "managed-local"`, `code: "no_chrome_found"`) are typed-union / result-code-taxonomy values used inline throughout the package (cloud/api.ts, skills/install.ts, etc.), so I left them as-is for consistency rather than const-ifying one site.

packages/cli/tests/driver-errors.test.ts
shrey150: Done in [03cd23539](https://github.com/browserbase/stagehand/commit/03cd23539) — `initFailureBackoffMs` is now an `it.each` table (failure count → expected ms), including the cap and nonsense-count rows.

PR #2249, opened 2026-06-12
[STG-2278] feat(cli): did-you-mean + telemetry for unknown commands

shrey150:
## Summary

Linear: https://linear.app/browserbase/issue/STG-2278/add-did-you-mean-suggestions-and-telemetry-for-unknown-browse-commands

Adds a `command_not_found` oclif hook to the browse CLI that prints a did-you-mean suggestion for unknown commands and emits a new `cli.command_not_found` telemetry event, while preserving oclif's standard "command not found" error and exit code 2.

## Impact if merged

Unknown commands (`browse sessions` / `search` / `contexts` / `auth status` — the old Commander-era syntax that agents were trained on, plus plain typos) currently exit 2 with no suggestion and emit NO telemetry event, so this failure class is invisible by construction. Old-binary telemetry shows the pattern is real (1,310 commander-error events from `sessions.list` alone in 30d; 115 from `search`). It's a failed-first-command class, and a failed first command cuts 7-day retention 12.4x (0.42% vs 5.21%). Did-you-mean turns an agent guess-loop into a one-turn recovery, and the new `cli.command_not_found` event finally lets us size and rank the dead ends. No new dependency — deliberately avoids `@oclif/plugin-not-found`, which prompts interactively (agent-hostile).

## Implementation notes

- **New hook** `src/hooks/command-not-found.ts`, registered in the `oclif.hooks` config. Suggestion order: explicit alias table first (old-CLI syntax → current tree, e.g. `sessions` → `cloud sessions list`, `auth status` → `doctor`, `search` → `cloud search`), then nearest match by Levenshtein over `config.commandIDs` with a distance threshold (the did-you-mean clause is omitted when nothing decent matches). Alias targets are validated against the live command tree at runtime and against `oclif.manifest.json` in tests, so they can't silently drift.
- **Privacy: id + suggestion only, never argv.** oclif's spaced-topic parsing glues unknown leading argv tokens into the attempted id (e.g. `browse opne https://example.com` arrives as `opne:https://example.com`), so the hook sanitizes down to leading command-shaped tokens and reports only the matched prefix (or the first token when nothing matches). The telemetry payload carries exactly `attempted_command` and `suggested_command` — URLs, selectors, queries, and secrets never leave the machine. Covered by a dedicated test asserting argv values are absent from captured payloads.
- **Exit semantics preserved.** A `command_not_found` hook that returns normally makes oclif treat the invocation as handled (exit 0), silently swallowing the failure. The hook therefore re-throws oclif's standard `CLIError("command <id> not found")` after printing the suggestion, keeping stderr output and exit code 2 byte-identical to current behavior.
- **Telemetry can't hang or get lost.** The event reuses the existing PostHog transport (400ms abort timeout, best-effort catch) and is awaited inside the hook before the error is thrown, so it is delivered before process exit but cannot delay it beyond the transport timeout. The `finally`-hook completion path early-returns for unknown commands (prerun never fires), so there is no double counting.
- No new runtime dependency; ~140 LOC of source plus tests.

## E2E Test Matrix

| --- | --- | --- |
| `node bin/run.js sessions` (local build) | stderr: `"browse sessions" is not a browse command. Did you mean "browse cloud sessions list"? Run browse --help for all commands.` then `Error: command sessions not found`; `echo $?` → `2` | Proves alias suggestion + preserved exit code on the highest-volume old-syntax pattern |
| `node bin/run.js auth status` | stderr: `"browse auth status" is not a browse command. Did you mean "browse doctor"? ...`; exit `2` | Proves multi-token alias matching (`auth:status` → `doctor`) |
| `node bin/run.js search "test"` | stderr: `"browse search" is not a browse command. Did you mean "browse cloud search"? ...`; exit `2`; the query token is not shown as part of the attempted command | Proves alias prefix matching strips trailing user args from messaging |
| `node bin/run.js opne https://example.com` | stderr: `"browse opne" is not a browse command. Did you mean "browse open"? ...`; exit `2` | Proves Levenshtein typo fallback; URL excluded from the attempted command |
| `node bin/run.js open https://example.com --local` | JSON result `{"mode": "managed-local", ..., "title": "Example Domain", "url": "https://example.com/"}`; exit `0`; no suggestion output | Proves valid commands are completely unaffected (hook never fires) |
| Live telemetry capture: `BROWSERBASE_TELEMETRY_HOST=<local capture server>` + `node bin/run.js auth status` | Capture server logged `POST /i/v0/e/` with `"event": "cli.command_not_found"`, `"attempted_command": "auth.status"`, `"suggested_command": "doctor"` plus standard env/version props; payload received before CLI exit `2`; no argv content in payload | Proves the event actually sends, flushes before process exit, and carries only id + suggestion |
| `pnpm test` (builds then vitest) | `Test Files 16 passed (16), Tests 229 passed (229)` — includes 13 new unit/integration tests (alias table validity vs manifest, Levenshtein, thresholds, token sanitization, built-CLI suggestion/exit-code/telemetry/privacy) | Full regression sweep; existing telemetry suite still green |
| `pnpm lint` | prettier + eslint + `tsc --noEmit` all pass | Supporting evidence only |

---
## Summary by cubic
Adds did-you-mean suggestions and privacy-safe telemetry for unknown `browse` CLI commands, while keeping the standard error output and exit code 2. Addresses Linear STG-2278 by helping users recover from old syntax and typos; typo matching now uses `fastest-levenshtein`.

- **New Features**
  - Added a `command_not_found` hook that prints a suggestion using an alias table for old syntax, with segment-aligned Levenshtein fallback for typos; omitted when no good match.
  - Sends `cli.command_not_found` telemetry with strict privacy: only the sanitized attempted command id and the suggested command, never raw argv.
  - Preserves default behavior by rethrowing the not-found error (stderr unchanged, exit code 2) and avoids `@oclif/plugin-not-found`.
  - Removed misleading `auth`/`login` → `doctor` suggestions.

<sup>Written for commit bcee6ed466b57053cddcba1a4610068503431d4e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2249?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cli/src/lib/command-suggestions.ts
+++ b/packages/cli/src/lib/command-suggestions.ts
@@ -14,9 +14,6 @@
  * also be topics (e.g. `cloud:contexts`, which prints topic help).
  */
 export const aliasSuggestions: ReadonlyMap<string, string> = new Map([
-  ["auth", "doctor"],
-  ["auth:status", "doctor"],
-  ["login", "doctor"],
   ["sessions", "cloud:sessions:list"],
   ["sessions:list", "cloud:sessions:list"],
   ["sessions:create", "cloud:sessions:create"],
--- a/packages/cli/src/lib/command-suggestions.ts
+++ b/packages/cli/src/lib/command-suggestions.ts
@@ -84,11 +84,38 @@ export function levenshtein(a: string, b: string): number {
   return previous[b.length] ?? 0;
 }
 
-function suggestionThreshold(attempted: string): number {
-  return Math.min(
-    Math.max(2, Math.floor(attempted.length / 3)),
-    maxSuggestionDistance,
-  );
+function tokenThreshold(token: string): number {
+  return Math.max(2, Math.floor(token.length / 3));
+}
+
+/**
+ * Segment-aligned fuzzy distance between a token prefix and a command id.
+ * Only command ids with the same number of segments are considered, and every
+ * token must be within its own edit-distance threshold of the corresponding
+ * segment. This means a trailing token can only ever be retained in the
+ * attempted command when it itself looks like a typo of a real command
+ * segment — free-form user values can never ride along.
+ */
+function prefixDistance(
+  tokens: readonly string[],
+  commandId: string,
+): number | null {
+  const segments = commandId.split(":");
+  if (segments.length !== tokens.length) {
+    return null;
+  }
+
+  let total = 0;
+  for (let i = 0; i < tokens.length; i++) {
+    const token = tokens[i] ?? "";
+    const distance = levenshtein(token, segments[i] ?? "");
+    if (distance > tokenThreshold(token)) {
+      return null;
+    }
+    total += distance;
+  }
+
+  return total > maxSuggestionDistance ? null : total;
 }
 
 /**
@@ -117,14 +144,11 @@ export function suggestCommand(
 
   let best: (CommandSuggestion & { distance: number }) | undefined;
   for (let length = tokens.length; length >= 1; length--) {
-    const attempted = tokens.slice(0, length).join(":");
+    const prefix = tokens.slice(0, length);
     for (const commandId of commandIds) {
-      const distance = levenshtein(attempted, commandId);
-      if (
-        distance <= suggestionThreshold(attempted) &&
-        (!best || distance < best.distance)
-      ) {
-        best = { attempted, suggestion: commandId, distance };
+      const distance = prefixDistance(prefix, commandId);
+      if (distance !== null && (!best || distance < best.distance)) {
+        best = { attempted: prefix.join(":"), suggestion: commandId, distance };
       }
     }
   }
--- a/packages/cli/src/lib/command-suggestions.ts
+++ b/packages/cli/src/lib/command-suggestions.ts
@@ -1,0 +1,137 @@
+/**
+ * Suggestion engine for unknown commands.
+ *
+ * oclif's spaced-topic parsing glues unknown leading argv tokens into the
+ * attempted command id (e.g. `browse opne https://example.com` arrives as
+ * `opne:https://example.com`), so the id may contain user-provided values.
+ * Everything here works on a sanitized token prefix and never returns raw
+ * argv content beyond the tokens that matched a known command shape.
+ */
+
+/**
+ * Old Commander-era syntax (and common agent guesses) mapped to the current
+ * command tree. Keys and values are colon-separated oclif ids; values may
+ * also be topics (e.g. `cloud:contexts`, which prints topic help).
+ */
+export const aliasSuggestions: ReadonlyMap<string, string> = new Map([
+  ["auth", "doctor"],
+  ["auth:status", "doctor"],
+  ["login", "doctor"],
+  ["sessions", "cloud:sessions:list"],
+  ["sessions:list", "cloud:sessions:list"],
+  ["sessions:create", "cloud:sessions:create"],
+  ["sessions:get", "cloud:sessions:get"],
+  ["projects", "cloud:projects:list"],
+  ["projects:list", "cloud:projects:list"],
+  ["contexts", "cloud:contexts"],
+  ["extensions", "cloud:extensions"],
+  ["search", "cloud:search"],
+  ["fetch", "cloud:fetch"],
+  ["goto", "open"],
+  ["navigate", "open"],
+]);
+
+const safeTokenPattern = /^[A-Za-z][A-Za-z0-9_-]*$/;
+const maxCommandTokens = 4;
+const maxSuggestionDistance = 5;
+
+export interface CommandSuggestion {
+  /** Sanitized colon-separated tokens treated as the attempted command. */
+  attempted: string;
+  /** Colon-separated suggested command or topic, when a decent match exists. */
+  suggestion: string | null;
+}
+
+/**
+ * Extracts the leading command-shaped tokens from an attempted id, stopping
+ * at the first token that does not look like a command word (URLs, selectors,
+ * flags, and other argument-like values).
+ */
+export function extractCommandTokens(id: string): string[] {
+  const tokens: string[] = [];
+  for (const token of id.split(":")) {
+    if (tokens.length >= maxCommandTokens || !safeTokenPattern.test(token)) {
+      break;
+    }
+    tokens.push(token.toLowerCase());
+  }
+  return tokens;
+}
+
+export function levenshtein(a: string, b: string): number {
+  if (a === b) {
+    return 0;
+  }
+  if (a.length === 0 || b.length === 0) {
+    return a.length || b.length;
+  }
+
+  let previous = Array.from({ length: b.length + 1 }, (_, index) => index);
+  let current = new Array<number>(b.length + 1).fill(0);
+
+  for (let i = 1; i <= a.length; i++) {
+    current[0] = i;
+    for (let j = 1; j <= b.length; j++) {
+      const deletion = (previous[j] ?? 0) + 1;
+      const insertion = (current[j - 1] ?? 0) + 1;
+      const substitution =
+        (previous[j - 1] ?? 0) + (a[i - 1] === b[j - 1] ? 0 : 1);
+      current[j] = Math.min(deletion, insertion, substitution);
+    }
+    [previous, current] = [current, previous];
+  }
+
+  return previous[b.length] ?? 0;
+}
+
+function suggestionThreshold(attempted: string): number {
+  return Math.min(
+    Math.max(2, Math.floor(attempted.length / 3)),
+    maxSuggestionDistance,
+  );
+}
+
+/**
+ * Computes a suggestion for an unknown command id. Explicit aliases win over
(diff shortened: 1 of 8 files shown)


packages/cli/src/lib/command-suggestions.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  let best: (CommandSuggestion & { distance: number }) | undefined;
+  for (let length = tokens.length; length >= 1; length--) {
+    const attempted = tokens.slice(0, length).join(":");
+    for (const commandId of commandIds) {
+      const distance = levenshtein(attempted, commandId);
+      if (
+        distance <= suggestionThreshold(attempted) &&
+        (!best || distance < best.distance)
```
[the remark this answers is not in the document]
shrey150: Addressed in [340fb4e70](https://github.com/browserbase/stagehand/commit/340fb4e70) — valid catch. Fuzzy matching is now segment-aligned instead of whole-string: a token prefix only matches command ids with the same number of segments, and every token must be within its own edit-distance threshold of the aligned segment. A trailing user token can therefore only be retained when it itself looks like a typo of a real command word (e.g. `browse cloud sessions lst`); free-form values can never ride along (`browse stat s` now reports `attempted_command: "stat"` and suggests `status`). Added a regression test covering exactly this case.

packages/cli/src/lib/command-suggestions.ts
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * Old Commander-era syntax (and common agent guesses) mapped to the current
+ * command tree. Keys and values are colon-separated oclif ids; values may
+ * also be topics (e.g. `cloud:contexts`, which prints topic help).
+ */
+export const aliasSuggestions: ReadonlyMap<string, string> = new Map([
+  ["auth", "doctor"],
+  ["auth:status", "doctor"],
+  ["login", "doctor"],
```
shrey150: I don't think auth + login commands should send you to doctor, they're not related

packages/cli/src/lib/command-suggestions.ts
```diff
@@ -0,0 +1,161 @@
...
+  for (const token of id.split(":")) {
+    if (tokens.length >= maxCommandTokens || !safeTokenPattern.test(token)) {
+      break;
+    }
+    tokens.push(token.toLowerCase());
+  }
+  return tokens;
+}
+
+export function levenshtein(a: string, b: string): number {
```
shrey150: there has to be a util function for this

packages/cli/src/lib/command-suggestions.ts
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * Segment-aligned fuzzy distance between a token prefix and a command id.
+ * Only command ids with the same number of segments are considered, and every
+ * token must be within its own edit-distance threshold of the corresponding
+ * segment. This means a trailing token can only ever be retained in the
+ * attempted command when it itself looks like a typo of a real command
+ * segment — free-form user values can never ride along.
+ */
+function prefixDistance(
```
shrey150, line 72: there has to be a util func for this

shrey150:
Re: [the auth/login alias comment](https://github.com/browserbase/stagehand/pull/2249#discussion_r3406499083) — agreed and addressed in [e8520c4a0](https://github.com/browserbase/stagehand/commit/e8520c4a0e8d4099d7db03ee6a16f032b4e17fdd): dropped all three (`auth`, `auth:status`, `login` → doctor). They now fall through to the plain not-found message (verified live; exit 2 unchanged), and `cli.command_not_found` still records the attempts — which doubles as demand signal for a real `browse login`. Regression test pins that auth/login never map to doctor.

Also pre-addressed the two pending draft comments on the hand-rolled edit distance ("there has to be a util function for this") in [37ae41c74](https://github.com/browserbase/stagehand/commit/37ae41c74): swapped to `fastest-levenshtein` (zero-dep, 21KB, the same lib `@oclif/plugin-not-found` uses, already in the monorepo pnpm store). The segment-aligned per-token threshold logic stays ours — no util provides that. −26 LOC, 230/230 tests green.

(Replying here instead of in-thread: GitHub blocks threaded replies while your review draft is pending.)

packages/cli/src/lib/command-suggestions.ts
shrey150: Addressed in [e8520c4a0](https://github.com/browserbase/stagehand/commit/e8520c4a0e8d4099d7db03ee6a16f032b4e17fdd) — dropped all three (`auth`, `auth:status`, `login` → doctor). They fall through to the plain not-found message now (verified live, exit 2 unchanged), `cli.command_not_found` still records the attempts (demand signal for a real `browse login`), and a regression test pins that auth/login never map to doctor.

packages/cli/src/lib/command-suggestions.ts
shrey150: Agreed — addressed in [37ae41c74](https://github.com/browserbase/stagehand/commit/37ae41c74): swapped the hand-rolled DP for `fastest-levenshtein` (zero-dep, 21KB, the same lib `@oclif/plugin-not-found` uses; already in the monorepo pnpm store). −26 LOC, 230/230 tests green.

packages/cli/src/lib/command-suggestions.ts
shrey150: Same fix — [37ae41c74](https://github.com/browserbase/stagehand/commit/37ae41c74) replaces the edit-distance implementation with `fastest-levenshtein`'s `distance()`. The segment-aligned per-token thresholding above this line stays ours: that's the privacy guard (free-form user tokens can't ride into telemetry), and no util provides it.

packages/cli/tests/cli-command-not-found.test.ts
```diff
@@ -0,0 +1,272 @@
...
+  while (cleanupPaths.length > 0) {
+    const path = cleanupPaths.pop();
+    if (!path) continue;
+    await rm(path, { recursive: true, force: true });
+  }
+});
+
+describe("extractCommandTokens", () => {
+  it("keeps leading command-shaped tokens and lowercases them", () => {
+    expect(extractCommandTokens("auth:status")).toEqual(["auth", "status"]);
```
ajmcquilkin: These could be table tests for simplicity

packages/cli/tests/cli-command-not-found.test.ts
```diff
@@ -0,0 +1,272 @@
...
+    "open",
+    "doctor",
+    "status",
+    "cloud:search",
+    "cloud:sessions:list",
+    "cloud:sessions:create",
+  ];
+
+  it("prefers explicit aliases over fuzzy matches", () => {
+    expect(suggestCommand("sessions", commandIds)).toEqual({
```
ajmcquilkin: These should be table tests too

packages/cli/tests/cli-command-not-found.test.ts
```diff
@@ -0,0 +1,272 @@
...
+        "cloud.sessions.list",
+      );
+      expect(payload?.properties?.source).toBe("cli");
+    } finally {
+      await telemetryServer.close();
+    }
+  });
+
+  it("never sends raw argv values in not-found telemetry", async () => {
+    const telemetryServer = await startTelemetryServer();
```
ajmcquilkin: These are closer to integration tests, might be worth separating them (in my head mocks + interfaces are unit, dummy servers are integration)

ajmcquilkin (approved):

packages/cli/tests/cli-command-not-found.test.ts
shrey150: Done in [bcee6ed46](https://github.com/browserbase/stagehand/commit/bcee6ed46) — `extractCommandTokens` is now an `it.each` table.

packages/cli/tests/cli-command-not-found.test.ts
shrey150: Done in [bcee6ed46](https://github.com/browserbase/stagehand/commit/bcee6ed46) — `suggestCommand` cases are now an `it.each` table of `[name, input, expected]`. Kept the auth/login case as a separate `it` since it's a negative assertion (must *not* map to doctor), and added a comment noting the trailing-token rows double as the telemetry-privacy guard.

packages/cli/tests/cli-command-not-found.test.ts
shrey150: Agreed with the mocks-vs-dummy-servers line. Split in [bcee6ed46](https://github.com/browserbase/stagehand/commit/bcee6ed46): the built-CLI tests (real command_not_found hook + telemetry against a dummy server) moved to `cli-command-not-found.integration.test.ts`; the pure-function unit tests stay in `cli-command-not-found.test.ts`. Both still run under `test:cli`.

PR #2250, opened 2026-06-12
[STG-2279] fix(cli): quote win32 npx spawn + bound skills installer stages

shrey150:
## Summary

Linear: https://linear.app/browserbase/issue/STG-2279/fix-windows-skills-add-npx-quoting-and-bound-installer-timeouts

Fixes `browse skills add` on Windows (cmd.exe spawn quoting) and bounds the two unbounded skills-installer stages (the `npx skills add` child and the catalog/file fetches).

## Impact if merged

skills.add is effectively broken for a meaningful share of Windows users — every default `C:\Program Files\nodejs` Node install fails, because the npx child is spawned through cmd.exe with an unquoted path. Windows devs are a smaller slice of the CLI base, but skill installers are the single highest-value cohort in our telemetry: significantly higher engagement and multi-day retention than non-installers, and skills.find→add is an agent-facing funnel (a large share of finders attempt an install within the hour). This also bounds the two unbounded installer stages (npx child: 180s; catalog fetches: 10s) — today they can hang forever, feeding the slow-failure retry loops that dominate telemetry volume.

## Implementation notes

**Root cause.** `findExecutable` resolves `npx` via PATH+PATHEXT to `npx.cmd` on Windows, and `spawnPassthrough` spawns it with `shell: true` (required for `.cmd`/`.bat` shims). Node's `shell: true` joins command+args **unquoted** into `cmd.exe /d /s /c "..."`, so `C:\Program Files\nodejs\npx.cmd` splits at the space and cmd executes `C:\Program` → `'C:\Program' is not recognized` → exit 1. Install-path args under `C:\Users\<First Last>\...` break the same way.

**Why not `shell: false`.** Spawning a `.cmd` directly with `shell: false` throws `EINVAL` on all current Node versions — the CVE-2024-27980 hardening (Node 18.20.x / 20.12.x / 21.7.x+) forbids spawning batch files without a shell because cmd.exe argument splitting cannot be made injection-safe generically. So the shell path is mandatory for `.cmd` shims, and the args must be quoted for cmd.

**Quoting semantics.** `quoteForCmdShell` wraps tokens containing whitespace, quotes, or cmd metacharacters (`^ & | < >`) in double quotes, doubling embedded quotes. Node wraps the joined string in outer quotes after `/d /s /c`; with `/s`, cmd strips only those outer quotes and executes the inner, correctly-quoted line:

```
before:  cmd.exe /d /s /c "C:\Program Files\nodejs\npx.cmd --yes skills add C:\Users\First Last\..."
after:   cmd.exe /d /s /c ""C:\Program Files\nodejs\npx.cmd" --yes skills add "C:\Users\First Last\...""
```

**Alternative considered.** Resolving `npx-cli.js` next to `npx.cmd` and spawning `process.execPath` with `shell: false` would avoid cmd quoting entirely, but the `npx.cmd` → `npx-cli.js` relative layout differs across npm versions and Node distribution channels (nvm-windows, Volta, Scoop shims, fnm), so it trades a well-understood quoting rule for fragile path archaeology. The quoting approach is smaller and matches what cross-platform tools (e.g. `cross-spawn`) do.

**Bounding the installer stages.**
- `spawnPassthrough` now enforces a 180s deadline: SIGTERM, then SIGKILL after 5s if the child ignores it. A timed-out install fails with a clear message and a distinct `skill_install_timeout` result code through the existing `fail`/`resultCode` plumbing from #2210.
- The catalog file-list fetch, the direct-Blob HEAD probe, and skill-file downloads now use `AbortSignal.timeout(10s)`. An aborted catalog fetch is classified exactly like a network failure (`unavailable`), preserving the existing fallback semantics.
- Both deadlines are env-overridable (`BROWSE_SKILLS_INSTALL_TIMEOUT_MS`, `BROWSE_SKILLS_FETCH_TIMEOUT_MS`), following the module's existing `BROWSE_SKILLS_*` override pattern; this is also what makes the deadlines provable end-to-end in tests.

## E2E Test Matrix

All commands ran against the locally built CLI (`<local build>/bin/run.js`) on macOS (darwin/arm64).

| --- | --- | --- |
| **Windows execution** (one-time `windows-latest` before/after run: [actions/runs/27448084696](https://github.com/browserbase/stagehand/actions/runs/27448084696)) | System Node at `C:\Program Files\nodejs` (no setup-node), `where npx` → `C:\Program Files\nodejs\npx.cmd`. **main:** `browse skills install` → exit 1, verbatim `'C:\Program' is not recognized as an internal or external command`. **PR head d2e909899:** exit 0, `Installed 1 skill`, `~\.agents\skills\browse\SKILL.md` present; win32 vitest gate (quoting + shim + spawnPassthrough timeout) 10/10 passed. | Closes the Windows gap with a real before/after on the same runner layout (win25-vs2026): the exact predicted failure reproduces on main and the PR build installs end-to-end. Full evidence in the [validation comment](https://github.com/browserbase/stagehand/pull/2250#issuecomment-4696289105). |
| `browse skills find flights` (real catalog) | exit 0; returned `google.com/search-flights-ts4g1f` with full metadata | Proves catalog discovery is unaffected. |
| `browse skills add google.com/search-flights-ts4g1f` (real catalog, real `npx`) | exit 0; `Downloaded 2 skill files to <config dir>`; `npx skills add` installed the skill ("Installed 1 skill ... Done!") | Proves the darwin install path (quoting branch not taken) still works end-to-end with the new deadline code in place — no regression. |
| Quoting before/after for `C:\Program Files\nodejs\npx.cmd` (unit tests + helper output) | before: `C:\Program Files\nodejs\npx.cmd --yes skills add C:\Users\First Last\...` (unquoted → cmd runs `C:\Program`); after: `"C:\Program Files\nodejs\npx.cmd" --yes skills add "C:\Users\First Last\..."` | Reproduces the bug shape and asserts the exact corrected command line, incl. embedded-quote doubling, `& \| ^ < >` metachars, and the empty token. Static proof only — see Windows row. |
| Hung `npx` stub (`exec /bin/sleep 600`) + `BROWSE_SKILLS_INSTALL_TIMEOUT_MS=2000` → `browse skills install` | exit 1 after 2s elapsed (timed): `Skill install timed out after 2s waiting for \`npx skills add\`...` | Proves the deadline kills a hung child and surfaces the timeout failure (`skill_install_timeout` flows through the same `fail` plumbing verified in #2210). Also covered by `spawnPassthrough` unit tests (timeout + non-timeout control). |
| Hung catalog server (accepts, never responds) at **default timeouts** → `browse skills add google.com/search-flights-ts4g1f` with stubbed `npx` | exit 0 after 21s (10s API fetch abort + 10s Blob HEAD abort); npx stub invoked with `--yes skills add browserbase/browse.sh --skill google.com/search-flights-ts4g1f` | Proves a hung catalog aborts at the 10s default and the catalog-unavailable fallback semantics are preserved. Previously this hung forever. Also covered by a fast CLI-level test with `BROWSE_SKILLS_FETCH_TIMEOUT_MS=500`. |
| `npx vitest run` (packages/cli) | 15 files, 224 tests passed (incl. 13 in skills-install.test.ts) | Full CLI suite green; supporting evidence only. |
| `pnpm lint` (packages/cli) | exit 0 (prettier + eslint + tsc) | Supporting evidence only. |

**Windows gap closed:** a one-time `windows-latest` before/after run ([actions/runs/27448084696](https://github.com/browserbase/stagehand/actions/runs/27448084696), details in the [validation comment](https://github.com/browserbase/stagehand/pull/2250#issuecomment-4696289105)) reproduced the exact `'C:\Program' is not recognized` failure on main and verified `browse skills install` succeeds end-to-end on this PR's build with the default `C:\Program Files\nodejs` system Node. Full Windows vitest: 205/224 passed; all 17 failures are pre-existing POSIX test-harness assumptions (`#!/bin/sh` npx stubs etc.), identical by construction on main.

---
## Summary by cubic
Fixes Windows failures in `browse skills add` by quoting the `npx` command when spawned via cmd.exe and bounding installer/fetch stages to prevent hangs. Also runs the full CLI test suite on Linux and Windows via a matrix job. Addresses Linear: STG-2279.

- **Bug Fixes**
  - Quote command and args when spawning `.cmd`/`.bat` through the shell, so `C:\Program Files\nodejs\npx.cmd` and paths with spaces work.
  - Add a 180s deadline to `npx skills add` (SIGTERM, then SIGKILL) and a 10s abort for catalog/file fetches; both overridable via `BROWSE_SKILLS_INSTALL_TIMEOUT_MS` and `BROWSE_SKILLS_FETCH_TIMEOUT_MS`; install timeouts surface `skill_install_timeout`.
  - Run the full CLI suite on `ubuntu-latest` and `windows-latest` via a matrix; POSIX-only tests are guarded via `itPosix`/`describePosix` so Windows gets full coverage without brittle CI filters.

<sup>Written for commit 9fe60b7964d2dc9f9ccc4e03f6f982bcc2cb7f75. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2250?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -351,13 +351,12 @@ jobs:
           use-prebuilt-artifacts: "true"
           restore-turbo-cache: "false"
 
-      # Targeted Windows gate: the platform-specific spawn/quoting tests that
-      # guard the cmd.exe npx invocation. The rest of the CLI suite still has
-      # POSIX-only harness assumptions (#!/bin/sh stubs), so it stays on the
-      # ubuntu job until the harness is made cross-platform.
-      - name: Run win32 CLI tests
-        shell: bash
-        run: pnpm --filter=browse exec vitest run tests/skills-install.test.ts -t "quoteForCmdShell|spawnPassthrough|uses a shell for Windows command shims"
+      # Runs the full CLI suite on Windows. POSIX-only tests opt out at the test
+      # site via the itPosix/describePosix helpers (tests/helpers/platform.ts),
+      # so coverage is on by default — a new test runs here unless it explicitly
+      # skips. No test-name knowledge lives in this workflow.
+      - name: Run CLI Tests
+        run: pnpm exec turbo run test:cli --filter=browse
 
   run-evals-unit-tests:
     name: Evals Unit Tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -319,10 +319,14 @@ jobs:
           retention-days: 1
 
   run-cli-tests:
-    name: CLI Tests
-    runs-on: ubuntu-latest
+    name: CLI Tests (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
     needs: [run-build, determine-changes]
     if: needs.determine-changes.outputs.cli == 'true'
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
@@ -333,28 +337,10 @@ jobs:
           use-prebuilt-artifacts: "true"
           restore-turbo-cache: "false"
 
-      - name: Run CLI Tests
-        run: pnpm exec turbo run test:cli --filter=browse
-
-  run-cli-tests-win32:
-    name: CLI Tests (win32)
-    runs-on: windows-latest
-    needs: [run-build, determine-changes]
-    if: needs.determine-changes.outputs.cli == 'true'
-    steps:
-      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
-        with:
-          fetch-depth: 1
-
-      - uses: ./.github/actions/setup-node-pnpm-turbo
-        with:
-          use-prebuilt-artifacts: "true"
-          restore-turbo-cache: "false"
-
-      # Runs the full CLI suite on Windows. POSIX-only tests opt out at the test
+      # Runs the full CLI suite on each OS. POSIX-only tests opt out at the test
       # site via the itPosix/describePosix helpers (tests/helpers/platform.ts),
-      # so coverage is on by default — a new test runs here unless it explicitly
-      # skips. No test-name knowledge lives in this workflow.
+      # so coverage is on by default — a new test runs on every OS unless it
+      # explicitly skips. No test-name knowledge lives in this workflow.
       - name: Run CLI Tests
         run: pnpm exec turbo run test:cli --filter=browse
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -336,6 +336,29 @@ jobs:
       - name: Run CLI Tests
         run: pnpm exec turbo run test:cli --filter=browse
 
+  run-cli-tests-win32:
+    name: CLI Tests (win32)
+    runs-on: windows-latest
+    needs: [run-build, determine-changes]
+    if: needs.determine-changes.outputs.cli == 'true'
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          fetch-depth: 1
+
+      - uses: ./.github/actions/setup-node-pnpm-turbo
+        with:
+          use-prebuilt-artifacts: "true"
+          restore-turbo-cache: "false"
+
+      # Targeted Windows gate: the platform-specific spawn/quoting tests that
+      # guard the cmd.exe npx invocation. The rest of the CLI suite still has
+      # POSIX-only harness assumptions (#!/bin/sh stubs), so it stays on the
+      # ubuntu job until the harness is made cross-platform.
+      - name: Run win32 CLI tests
+        shell: bash
+        run: pnpm --filter=browse exec vitest run tests/skills-install.test.ts -t "quoteForCmdShell|spawnPassthrough|uses a shell for Windows command shims"
+
   run-evals-unit-tests:
     name: Evals Unit Tests
     runs-on: ubuntu-latest
--- a/.changeset/win32-skills-add-quoting.md
+++ b/.changeset/win32-skills-add-quoting.md
@@ -1,0 +1,9 @@
+---
+"browse": patch
+---
+
+Fix `browse skills add` on Windows and bound the unbounded installer stages.
+
+- Quote the `npx` command and arguments when spawning through cmd.exe (`shell: true` for `.cmd`/`.bat` shims), so the default `C:\Program Files\nodejs\npx.cmd` path and install paths with spaces (e.g. `C:\Users\First Last\...`) no longer split at the space and fail with "'C:\Program' is not recognized".
+- Kill the `npx skills add` child after a 180s deadline (SIGTERM, then SIGKILL) and fail with a clear message and a distinct `skill_install_timeout` telemetry result code instead of hanging forever.
+- Bound the catalog and skill-file fetches with a 10s abort timeout, preserving the existing catalog-unavailable fallback semantics when a fetch hangs.
--- a/packages/cli/src/lib/skills/install.ts
+++ b/packages/cli/src/lib/skills/install.ts
@@ -63,6 +63,26 @@ type SkillFilesResult =
     };
 
 const maxCapturedOutputBytes = 2048;
+const defaultInstallTimeoutMs = 180_000;
+const defaultFetchTimeoutMs = 10_000;
+
+function envTimeoutMs(name: string, fallback: number): number {
+  const value = Number(process.env[name]);
+  return Number.isFinite(value) && value > 0 ? value : fallback;
+}
+
+function installTimeoutMs(): number {
+  return envTimeoutMs(
+    "BROWSE_SKILLS_INSTALL_TIMEOUT_MS",
+    defaultInstallTimeoutMs,
+  );
+}
+
+function fetchTimeoutSignal(): AbortSignal {
+  return AbortSignal.timeout(
+    envTimeoutMs("BROWSE_SKILLS_FETCH_TIMEOUT_MS", defaultFetchTimeoutMs),
+  );
+}
 
 export function parseSkillId(rawSkillId: string): ParsedSkillId {
   const parts = rawSkillId.split("/");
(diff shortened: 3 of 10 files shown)


shrey150:
## Win32 validation (one-time windows-latest run)

Closes the Windows verification gap called out in the E2E matrix. A throwaway workflow (`win32-skills-validation.yml`, on a now-deleted throwaway branch — PR diff untouched) ran a before/after on a GitHub-hosted `windows-latest` runner, deliberately **without** `actions/setup-node` so the CLI resolves the system Node at the space-containing default install path — the exact bug trigger.

**Run (green):** https://github.com/browserbase/stagehand/actions/runs/27448084696
- Repro on main (expected failure): https://github.com/browserbase/stagehand/actions/runs/27448084696/job/81137436168
- Verify PR head (expected success): https://github.com/browserbase/stagehand/actions/runs/27448084696/job/81137436160
- First iteration (same before/after results, full-suite step details): https://github.com/browserbase/stagehand/actions/runs/27447860478

### Runner layout (both jobs)

```
ImageOS=win25-vs2026 ImageVersion=20260608.135.2
node: v22.22.3  npm: 10.9.8
where.exe npx:
  C:\Program Files\nodejs\npx
  C:\Program Files\nodejs\npx.cmd
```

`npx` resolves under `C:\Program Files\nodejs` — the space-containing path from the telemetry cohort.

### Before — `origin/main`: `node packages/cli/bin/run.js skills install`

```
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.
Could not install skill: 'C:\Program' is not recognized as an internal or external command,
operable program or batch file.
```

Exit code **1** — verbatim, the exact failure shape predicted in the PR description.

### After — PR head `d2e909899`: same command, same runner layout

```
◇  Found 1 skill
●  Installing to all 71 agents
◇  Installed 1 skill
   ✓ ~\.agents\skills\browse
└  Done!
```

Exit code **0**, and the job asserted the installed file exists: `C:\Users\runneradmin\.agents\skills\browse\SKILL.md`. (The one `✗ browse → PromptScript: PromptScript does not support global skill installation` line is upstream `npx skills` behavior for that agent target, identical on every platform — not Windows-related.)

### Windows vitest results (packages/cli, PR head)

| Suite | Result | Notes |
| --- | --- | --- |
| Win32-relevant gate: `vitest run tests/skills-install.test.ts -t "quoteForCmdShell\|spawnPassthrough\|uses a shell for Windows command shims"` | **10 passed, 3 skipped** | All quoting tests, shim detection, and both `spawnPassthrough` timeout tests (incl. "kills a hung child after the deadline", 522ms) pass natively on Windows — this also covers the install-timeout behavior, so no separate `BROWSE_SKILLS_INSTALL_TIMEOUT_MS` smoke was needed. |
| Full CLI suite: `vitest run` | 205 passed, 17 failed, 2 skipped (224) | All 17 failures are pre-existing POSIX assumptions in the test harness, not product code: 9 skills/skills-install tests stub `npx` as an extension-less `#!/bin/sh` script (unresolvable via PATHEXT on Windows → "`npx` is not installed") or `exec /bin/sleep`; the other 8 (cli-functions-contract ×3, cli-templates ×3, driver-foundation ×1, package-manifest ×1) have similar host-tooling/path assumptions. Same-by-construction on main and this PR. |

### Net

Same runner, same Node layout, same command: **main fails with the exact predicted `'C:\Program' is not recognized` error; this PR installs the skill end-to-end.** The Windows row in the E2E matrix is updated to point at this run.

ajmcquilkin (commented):
I have no idea if this works on Windows or not, I would want to see platform matrix E2E tests in CI to validate that this install works

.github/workflows/ci.yml
```diff
@@ -336,6 +336,29 @@ jobs:
...
+          use-prebuilt-artifacts: "true"
+          restore-turbo-cache: "false"
+
+      # Targeted Windows gate: the platform-specific spawn/quoting tests that
+      # guard the cmd.exe npx invocation. The rest of the CLI suite still has
+      # POSIX-only harness assumptions (#!/bin/sh stubs), so it stays on the
+      # ubuntu job until the harness is made cross-platform.
+      - name: Run win32 CLI tests
+        shell: bash
+        run: pnpm --filter=browse exec vitest run tests/skills-install.test.ts -t "quoteForCmdShell|spawnPassthrough|uses a shell for Windows command shims"
```
shrey150: only worry: is this too brittle of a filter? is there a better way to indicate win32 specific tests?

shrey150:
**Alternative considered: `cross-spawn`.** The canonical library for this class of Windows spawn bug is [`cross-spawn`](https://github.com/moxystudio/node-cross-spawn), and it's worth saying why we didn't reach for it here.

We chose the local `quoteForCmdShell` quoter instead because:
- `cross-spawn` isn't a drop-in — it does its own executable resolution (PATH/PATHEXT/shebang), which overlaps the `findExecutable` logic this PR already relies on. Adopting it means reworking/removing that resolution path too, i.e. a larger change than this targeted fix, touching code that's currently working and now verified on Windows.
- Swapping the spawn path would invalidate the one-time `windows-latest` before/after validation (the proof would no longer apply to the shipped code).
- The quoter is unit-tested for the inputs in play (default path, spaces, embedded-quote doubling, `& | ^ < >`, empty token) and those tests run natively on the win32 CI leg.

The main long-term argument for `cross-spawn` is security (the cmd.exe command-injection edge — the "BatBadBut"/CVE-2024-1874 class), but our exposure is low: the spawn inputs are a resolved npx path, fixed flags, a constrained skill ID, and a local install path — none arbitrary. If we later spawn more child processes cross-platform, or skill IDs become less trusted, migrating spawn → `cross-spawn` (which would also remove the custom quoting) is the better shape. Not blocking this fix on it.

.github/workflows/ci.yml
```diff
@@ -319,10 +319,14 @@ jobs:
...
-    name: CLI Tests
-    runs-on: ubuntu-latest
+    name: CLI Tests (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
     needs: [run-build, determine-changes]
     if: needs.determine-changes.outputs.cli == 'true'
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
```
ajmcquilkin, line 329: Nit: long-term let's do macos too

ajmcquilkin (approved):
Approving with passing tests

.github/workflows/ci.yml
shrey150: Agreed — tracked as a backlog follow-up: [STG-2403](https://linear.app/browserbase/issue/STG-2403/add-macos-latest-to-cli-test-matrix). Keeping it out of this PR since macOS shares the POSIX/no-shell spawn path with the Linux leg (so it adds little coverage for this win32 fix) and macOS runners bill at 10x minutes — nice-to-have, not urgent.

PR #2251, opened 2026-06-12
[STG-2280] feat(cli): emit skill_id telemetry property

shrey150:
## Summary

Adds two properties to the existing `cli.command_completed` telemetry event:

- **`skill_id`** — the validated, catalog-public skill id (e.g. `google.com/search-flights-ts4g1f`, or `bundled/browse` for `browse skills install`), attached for `skills add`/`install` on success **and** every downstream failure path (`skill_not_found`, `skill_install_failed`, `npx_missing`, ...). `null` on all other commands.

## Implementation notes

- **Privacy rule:** `skill_id` is set only **after** `parseSkillId` validation succeeds (ids are catalog-public and regex-constrained). The raw, unparsed argument — which can contain arbitrary user input or local paths — is never attached. A test asserts the raw argument does not appear anywhere in any payload.
- `skill_id` flows through the existing `RunTelemetryState` (same channel as `result_code`/`http_status`), so it is captured by the one `cli.command_completed` emit point regardless of how the command ends.

## E2E Test Matrix

Run against the locally built `browse` (`node <local build>/bin/run.js`) with a local HTTP capture server (`BROWSERBASE_TELEMETRY_HOST=http://127.0.0.1:<port>`) logging full POST bodies, and a fresh `BROWSERBASE_TELEMETRY_INSTALL_ID_FILE` dir. All payloads below are real captured `cli.command_completed` events.

| --- | --- | --- |
| `browse skills add google.com/search-flights-ts4g1f` (real catalog + real `npx skills add`, exit 0) | Captured event: `{"command_path":"skills.add", "success":true, "result_code":"ok", "skill_id":"google.com/search-flights-ts4g1f"}` | Proves the validated id is attached end-to-end on a real successful install against the live catalog. |
| `browse skills find flights` (real catalog) | Exit 0, returned `google.com/search-flights-ts4g1f`; captured event has `skill_id:null` | Supporting: `skill_id` stays null for non-install skills commands. |
| `pnpm --filter browse lint` (prettier + eslint + tsc) | clean | Supporting: format/lint/typecheck pass. |

## Changeset

---
## Summary by cubic
Adds a `skill_id` property to `cli.command_completed` telemetry for `browse skills add` and `install`. It records the parsed, catalog‑public id on success and all failure paths, is `null` for other commands, and never includes the raw argument.

<sup>Written for commit 41a8c56a37f1cbe3246560dbcf07b553c9a05876. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2251?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Why no first_success
The original revision also stamped a `first_success` property via a per-install marker file. Dropped in `2c526dac5` per maintainer decision: PostHog's `first_time_for_user` math on (`cli.command_completed`, `success=true`, driver commands) derives the same activation signal at query time — verified live (118→456 weekly first-activations over the last 4 weeks; ~2%→37% normalized per new install). Details: https://github.com/browserbase/stagehand/pull/2251#issuecomment-4695908087. `skill_id` remains — that data exists in no event today.

### DIFF
--- a/.changeset/telemetry-skill-id-first-success.md
+++ b/.changeset/telemetry-skill-id-first-success.md
@@ -2,7 +2,6 @@
 "browse": patch
 ---
 
-Emit `skill_id` and `first_success` properties on `cli.command_completed` telemetry.
+Emit a `skill_id` property on `cli.command_completed` telemetry.
 
-- `skill_id`: the validated, catalog-public skill id (e.g. `yelp.com/extract-reviews`, or `bundled/browse` for `skills install`) is attached to the completion event for `browse skills add`/`install`, covering both successful installs and every downstream failure path (`skill_not_found`, `skill_install_failed`, ...). Only the parsed, regex-validated id is ever attached — never the raw argument.
-- `first_success: true` is emitted exactly once per anonymous install: the first time a browser-driver command completes successfully, tracked via a marker file stored alongside the existing anonymous install id. Best-effort like all CLI telemetry — it never affects command behavior.
+The validated, catalog-public skill id (e.g. `yelp.com/extract-reviews`, or `bundled/browse` for `skills install`) is attached to the completion event for `browse skills add`/`install`, covering both successful installs and every downstream failure path (`skill_not_found`, `skill_install_failed`, ...). Only the parsed, regex-validated id is ever attached — never the raw argument.
--- a/.changeset/telemetry-skill-id-first-success.md
+++ b/.changeset/telemetry-skill-id-first-success.md
@@ -1,0 +1,8 @@
+---
+"browse": patch
+---
+
+Emit `skill_id` and `first_success` properties on `cli.command_completed` telemetry.
+
+- `skill_id`: the validated, catalog-public skill id (e.g. `yelp.com/extract-reviews`, or `bundled/browse` for `skills install`) is attached to the completion event for `browse skills add`/`install`, covering both successful installs and every downstream failure path (`skill_not_found`, `skill_install_failed`, ...). Only the parsed, regex-validated id is ever attached — never the raw argument.
+- `first_success: true` is emitted exactly once per anonymous install: the first time a browser-driver command completes successfully, tracked via a marker file stored alongside the existing anonymous install id. Best-effort like all CLI telemetry — it never affects command behavior.
--- a/packages/cli/src/lib/skills/install.ts
+++ b/packages/cli/src/lib/skills/install.ts
@@ -13,6 +13,7 @@ import { delimiter, dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
 
 import { fail } from "../errors.js";
+import { setRunTelemetryCompletion } from "../run-telemetry.js";
 import { defaultSkillsApiBaseUrl, isRecord, responseDetail } from "./shared.js";
 
 const defaultBlobBaseUrl =
@@ -106,6 +107,10 @@ export function isBlobSkillId(skillId: ParsedSkillId): boolean {
 
 export async function installSkill(rawSkillId: string): Promise<void> {
   const skillId = parseSkillId(rawSkillId);
+  // Attach only the validated, catalog-public id to telemetry — never the raw
+  // argument, which can contain arbitrary user input. Setting it right after
+  // parsing covers success and every downstream failure path.
+  setRunTelemetryCompletion({ skillId: skillId.id });
   const npxPath = await findExecutable("npx");
   if (!npxPath) {
     fail(
@@ -151,6 +156,7 @@ export async function installSkill(rawSkillId: string): Promise<void> {
 }
 
 export async function installBundledCliSkill(): Promise<void> {
+  setRunTelemetryCompletion({ skillId: "bundled/browse" });
   const npxPath = await findExecutable("npx");
   if (!npxPath) {
     fail(
--- a/packages/cli/src/lib/skills/install.ts
+++ b/packages/cli/src/lib/skills/install.ts
@@ -106,11 +106,11 @@ export function isBlobSkillId(skillId: ParsedSkillId): boolean {
 }
 
 export async function installSkill(rawSkillId: string): Promise<void> {
-  const skillId = parseSkillId(rawSkillId);
+  const skill = parseSkillId(rawSkillId);
   // Attach only the validated, catalog-public id to telemetry — never the raw
   // argument, which can contain arbitrary user input. Setting it right after
   // parsing covers success and every downstream failure path.
-  setRunTelemetryCompletion({ skillId: skillId.id });
+  setRunTelemetryCompletion({ skillId: skill.id });
   const npxPath = await findExecutable("npx");
   if (!npxPath) {
     fail(
@@ -120,9 +120,9 @@ export async function installSkill(rawSkillId: string): Promise<void> {
     );
   }
 
-  const files = await fetchSkillFiles(skillId);
+  const files = await fetchSkillFiles(skill);
   if (files.status === "found") {
-    const result = await downloadBlobSkill(skillId, files.files);
+    const result = await downloadBlobSkill(skill, files.files);
     process.stdout.write(
       `Downloaded ${result.fileCount} skill file${result.fileCount === 1 ? "" : "s"} to ${result.installPath}\n`,
     );
@@ -137,7 +137,7 @@ export async function installSkill(rawSkillId: string): Promise<void> {
 
   if (files.status === "not_found") {
     fail(
-      `Skill "${skillId.id}" not found in the catalog. Run \`browse skills find ${skillId.domain}\` to discover available skills, or \`browse skills list\` to browse them.`,
+      `Skill "${skill.id}" not found in the catalog. Run \`browse skills find ${skill.domain}\` to discover available skills, or \`browse skills list\` to browse them.`,
       1,
       { resultCode: "skill_not_found" },
     );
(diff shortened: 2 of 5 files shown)


packages/cli/src/lib/telemetry.ts
```diff
@@ -255,7 +258,69 @@ function commandCompletedProperties(
...
+  try {
+    await access(markerPath);
+    return false;
+  } catch {
+    // No marker yet: this is the first successful driver command.
+  }
+
+  try {
+    await mkdir(dirname(markerPath), { recursive: true });
+    await writeFile(markerPath, `${new Date().toISOString()}\n`, "utf8");
```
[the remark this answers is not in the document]
shrey150: Addressed in [c87ee42a0](https://github.com/browserbase/stagehand/commit/c87ee42a0) — replaced the access()-then-write sequence with a single exclusive create (`flag: "wx"`), so exactly one concurrent run can win the race. EEXIST (or any other persistence failure) now returns not-first instead of reporting first_success, which also removes the prior overcount-on-unwritable-config-dir behavior.

.changeset/telemetry-skill-id-first-success.md
```diff
@@ -0,0 +1,8 @@
+---
+"browse": patch
+---
+
+Emit `skill_id` and `first_success` properties on `cli.command_completed` telemetry.
+
+- `skill_id`: the validated, catalog-public skill id (e.g. `yelp.com/extract-reviews`, or `bundled/browse` for `skills install`) is attached to the completion event for `browse skills add`/`install`, covering both successful installs and every downstream failure path (`skill_not_found`, `skill_install_failed`, ...). Only the parsed, regex-validated id is ever attached — never the raw argument.
+- `first_success: true` is emitted exactly once per anonymous install: the first time a browser-driver command completes successfully, tracked via a marker file stored alongside the existing anonymous install id. Best-effort like all CLI telemetry — it never affects command behavior.
```
shrey150: why is this metric important? is it something we need to support first-class?

shrey150:
Re: the `first_success` design question — what it is, whether the backend can derive it, and how much it matters. I tested the backend path live before answering.

**What it is:** a boolean stamped on exactly one `cli.command_completed` event per install — the first time a browser-driver command (open/get/click/snapshot/…) succeeds. It's the event-level marker for "this install reached a working browser and did something" — the activation proxy from the June 11 health report (28.5% of real users). Implementation here: ~40 LOC + a marker file beside the install id.

**Can the backend derive it? Yes — verified live just now.** PostHog's native `first_time_for_user` math on (`cli.command_completed`, `success=true`, driver-command list) works on our anonymous events (person-on-events mode). Real numbers from project 370621, weekly first-activations: **118 → 158 → 280 → 456** over the last 4 weeks — non-zero, and the rising curve matches the known June activation recovery. So the headline dashboard tile ("installs activating per week", breakdowns by version/platform/`skill_present`) needs **zero CLI code**.

**What backend derivation costs:** `first_time_for_user` is a math mode, not a filter — you can't use "was this the install's first success?" as a condition in *other* insights. Time-to-activate distributions, activated-cohort behavior, and the #2247 readout (skill-present installs whose first commands are cloud → do they later reach first driver success) all become multi-CTE HogQL instead of property filters. All doable — the health report did exactly this — just slower per question, and the definition lives in each query instead of being pinned in code. Counterpoint in backend's favor: a query-time definition can be changed retroactively; a client-stamped property is frozen at emission.

**How important is it?** The *metric* is the north star — #2247 is literally parked on it. The client-side *property* is a convenience/robustness layer, not the only way to get the metric.

**Recommendation, given the live proof:** consistent with the de-scope philosophy on #2200 — **drop `first_success` from this PR and keep `skill_id`** (which is *not* backend-derivable; the data doesn't exist in any event today). I'll build the activation views as saved PostHog insights (the `first_time_for_user` tile + a saved HogQL funnel for the cross-cutting questions) so nothing is hand-rebuilt per readout. That cuts this PR roughly in half (~35 LOC of telemetry plumbing, no marker file). Say the word and I'll amend — or if you'd rather keep the composable event property, that's defensible too; it's the same trade as ever: 40 LOC in the CLI vs. HogQL at query time.

shrey150:
Amended in [2c526dac5](https://github.com/browserbase/stagehand/commit/2c526dac5) per the decision above — `first_success` and its marker file are gone; the PR is now `skill_id`-only (**+81 lines total** incl. tests, down from ~160). Title, body, changeset, and Linear updated. 215/215 tests green. The activation signal lives backend-side via `first_time_for_user` (recipe + caveats saved to the team's notes).

shrey150:
**Design note for the record — why `skill_id` stays client-side even though the backend sees skill requests** (raised in review: the Browse.sh API behind `browse skills list`/`add` already logs which skills are fetched).

Verified against the code: `skills find/list` hits `GET api/skills` and `skills add` hits the skill-files API with the parsed id — so the server does see **demand**. What it cannot see is **outcome**:

1. The dominant install failures happen *after* the file-list API call succeeds — the external `npx skills` chain (45% of adds fail overall; 3.2% success on Windows; timeout tails). Server-side, a failed and a successful install are indistinguishable: both fetched files. `skill_id` on `cli.command_completed` exists for the join to `success`/`result_code` — *which skills fail to install, and why*.
2. No join key: the CLI's catalog/file fetches are bare `fetch(url)` with no install id or client header, so API logs can't be connected to the PostHog funnel (find→add conversion per skill, retries, engagement by installed skill).
3. Coverage holes: `skills install` (bundled — no API call), the GitHub-clone fallback (bypasses the file API), and `invalid_skill_id` failures (die before any request).

Cost asymmetry vs the dropped `first_success`: that was ~40 LOC + filesystem state *and* fully backend-derivable (`first_time_for_user`) → cut. `skill_id` is 8 lines, zero state, and only its demand half exists server-side.

Complementary uses of the API logs: authoritative skill *popularity* (including non-CLI catalog clients) and cross-checking CLI-reported totals. If true server-side joinability is ever wanted, the path is adding `User-Agent: browse-cli/<ver>` + install-id headers to those fetches — a separate privacy/design decision tracked from the June 8 telemetry investigation.

.changeset/telemetry-skill-id-first-success.md
shrey150: This question got a full data answer on the conversation tab ([here](https://github.com/browserbase/stagehand/pull/2251#issuecomment-4695908087)) and changed the PR: activation IS the north-star metric (28.5% of real users; #2247 is parked on it) but it does **not** need first-class CLI support — PostHog's `first_time_for_user` math derives it at query time (verified live: 118→456 weekly first-activations, ~2%→37% normalized per new install). So `first_success` was dropped in [2c526dac5](https://github.com/browserbase/stagehand/commit/2c526dac5); the PR is now skill_id-only.

packages/cli/src/lib/telemetry.ts
```diff
@@ -255,6 +255,7 @@ function commandCompletedProperties(
       failureTelemetry?.requestHadHttpResponse ??
```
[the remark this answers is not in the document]
shrey150: Intentional, per maintainer decision in this PR's review — not a regression. The activation metric is derived backend-side instead: PostHog `first_time_for_user` on (cli.command_completed, success=true, driver commands) reproduces the one-time first-driver-success signal at query time, verified live against real data ([evidence + trade-offs](https://github.com/browserbase/stagehand/pull/2251#issuecomment-4695908087), [decision](https://github.com/browserbase/stagehand/pull/2251#issuecomment-4696006420)). The PR title/body/changeset were all updated to skill_id-only in the same commit.

packages/cli/tests/cli-telemetry.test.ts
```diff
@@ -220,6 +221,71 @@ describe("CLI telemetry", () => {
...
+          },
+        },
+      );
+
+      expect(result.exitCode).toBe(1);
+      expect(result.stderr).toContain("not found in the catalog");
+
+      const payloads = telemetryPayloads(telemetryServer);
+      const completedPayload = findPayload(payloads, "cli.command_completed");
+      expect(completedPayload.properties.command_path).toBe("skills.add");
```
ajmcquilkin: Nit: `expect(completedPayload.properties).toMatchObject({ command_path: "skills.add", skill_id: "...", result_code: "skill_not_found" })`

in my experience it's better to test full objects (easier to detect breaking object shape changes)

packages/cli/src/lib/skills/install.ts
```diff
@@ -106,6 +107,10 @@ export function isBlobSkillId(skillId: ParsedSkillId): boolean {
 
 export async function installSkill(rawSkillId: string): Promise<void> {
   const skillId = parseSkillId(rawSkillId);
```
ajmcquilkin: Nit: `skillId.id` is unintuitive, perhaps `skill.id`

ajmcquilkin (approved):

packages/cli/tests/cli-telemetry.test.ts
shrey150: Done in [41a8c56a3](https://github.com/browserbase/stagehand/commit/41a8c56a3) — switched to `toMatchObject` over the full shape (command_path, skill_id, result_code, success: false).

packages/cli/src/lib/skills/install.ts
shrey150: Renamed the local to `skill` in [41a8c56a3](https://github.com/browserbase/stagehand/commit/41a8c56a3) (`skill.id`, `skill.domain`). The telemetry field stays `skillId` since that's the RunTelemetryState key.

PR #2269, opened 2026-06-24
fix(agent): don't fail a completed run when the forced "done" call errors

akeimach:
# why

Edit: pulling in the description from @filip-michalsky on https://github.com/browserbase/stagehand/pull/2270

Root-cause fix for STG-2335: a non-CUA agent.execute() that successful
run as { success: false }, with a red Invalidprompt: messages must be a
ModelMessage[] error logged after all the work already completed.
This replaces the symptom patch (wrap the finalization in try/catch and
force state.completed = true) with a fix for the actual defect.
Root cause
After the main agent loop finishes, ensureDone() runs a forced "done"
finalization (handleDoneToolCall) that re-submits the accumulated run
history into a fresh generateText call to produce the structured
su-validates accumulated tool results, but this re-submissiondoes.
When a custom tool returns an object with an optional field left
undefined — e.g. captureField returning { matchedExpected:
undefined }when no expectedText is passed — that undefined lands insid
The AI SDK's ModelMessage validation (standardizePrompt)rejects it,
because its JSON-value schema disallows undefined (only
null/string/number/boolean/object/array). The finalization throws,
flipping the result to { success: false } even though every action succe

▎ Note: the original "reasoning traces" hypothesis was rule parts come
back with a valid text: "" and pass validation.The undefined tool-result
field is the trigger.

# what changed

sanitizeMessagesForResubmission() deep-strips undefined from the run
history before the forced "done" call, keeping all real content. It only
traverses plain objects/arrays, so class instances (URL, tyata, Date)
pass through untouched.

# test plan

- 4 unit tests in agent-finalization-resilience.test.ts agaces
InvalidPromptError with an undefined tool-result field →fixed by
sanitize → real content (reasoning/tool-call/text) preserved → class
instances untouched. All pass.
- End-to-end repro (openai/gpt-5.5 + custom tool, mirrors Pon main
(success=false, red error), succeeds on this branch(success=true,
completed=true, no error).

---
## Summary by cubic
Prevents non-CUA `agent.execute()` from reporting a completed run as failed by sanitizing the run history before the forced "done" call and making finalization best-effort. Fixes STG-2335 for reasoning models like `openai/gpt-5.x` by stripping nested `undefined` values that break SDK prompt validation.

- **Bug Fixes**
  - Deep-strip `undefined` from re-submitted messages via `sanitizeMessagesForResubmission`; traverse only plain objects/arrays to preserve real content and class instances; apply in `handleDoneToolCall` and null-guard `result.toolCalls`.
  - If the forced "done" call throws, log a warning and synthesize completion from the run instead of failing it.
  - Add unit tests for the InvalidPromptError repro (including `providerOptions` in `gpt-5.x`), sanitizer behavior, class-instance pass-through, and the finalization-failure fallback.

<sup>Written for commit 4d1c904e5842c1a1959751910dd7aefdfdc5ff5f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2269?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


miguelg719 (approved):

.changeset/curvy-pillows-attack.md
```diff
@@ -0,0 +1,7 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+Fix non-CUA `agent.execute()` reporting a successful run as failed. After the agent finished all of its work, the forced "done" finalization step (`handleDoneToolCall`) re-submitted the accumulated run history to the model. When a custom tool returned an object with an optional field left `undefined` (e.g. `{ matchedExpected: undefined }`), that `undefined` ended up inside a tool-result `output.value`, which the AI SDK's prompt validation rejects (its JSON-value schema disallows `undefined`), throwing `Invalid prompt: messages must be a ModelMessage[]`. This surfaced as a red error and flipped the result to `{ success: false }` even though every action had already completed (STG-2335).
```
seanmcguire12: @akeimach can we tighten this up a bit? try to keep implementation details out of the changeset, & just give a sentence on user visible behaviour

akeimach: ty!!

PR #2270, opened 2026-06-24
fix(agent): strip reasoning parts before forced "done" re-submission

filip-michalsky:
# why

Root-cause fix for **STG-2335**: a non-CUA `agent.execute()` that uses a custom tool reports a **successful** run as `{ success: false }`, with a red `Invalid prompt: messages must be a ModelMessage[]` error logged **after** all the work already completed.

After the main agent loop finishes, `ensureDone()` runs a forced "done" finalization (`handleDoneToolCall`) that re-submits the accumulated run history into a fresh `generateText` call. The main loop never re-validates accumulated tool results — but this re-submission does. When a custom tool returns an optional field left `undefined` (e.g. `captureField` returning `{ matchedExpected: undefined }` when no `expectedText` is passed), that `undefined` lands inside a tool-result `output.value`, and the AI SDK's `ModelMessage` validation (`standardizePrompt`) rejects it — its JSON-value schema disallows `undefined`. Finalization throws, flipping the result to `{ success: false }` even though every action succeeded.

PR #2269 patched the symptom (wrap finalization in `try/catch` and force `state.completed = true`). This PR fixes the actual defect instead.

> [!NOTE]
> The original "reasoning traces" hypothesis was ruled out empirically — reasoning parts come back with a valid `text: ""` and pass validation. The `undefined` tool-result field is the real trigger.

# what changed

- Added `sanitizeMessagesForResubmission()` in `handleDoneToolCall.ts`, which deep-strips `undefined` from the run history before the forced "done" call. Only plain objects/arrays are traversed, so class instances (`URL`, typed arrays for binary image data, `Date`, …) pass through untouched.
- Reverted PR #2269's force-`completed=true` fallback in `v3AgentHandler.ts`, so genuine finalization failures fail loudly again instead of being masked.
- Added unit coverage in `agent-finalization-resilience.test.ts`.
- Added changeset.

# test plan

- [x] Unit — 4 tests against the real `ai` SDK: reproduces `InvalidPromptError` with an undefined tool-result field → fixed by `sanitizeMessagesForResubmission` → real content (reasoning / tool-call / text) preserved → class instances untouched. All pass.
- [x] End-to-end — `openai/gpt-5.5` + custom tool

  | Branch | Result |
  | --- | --- |
  | `main` | ❌ `success=false`, red `Invalid prompt` error |
  | this branch | ✅ `success=true`, `completed=true`, no error |


akeimach (approved):
fabulous

PR #2280, opened 2026-06-25
[STG-2407] fix(cli): honor BROWSERBASE_API_KEY passed to an already-running daemon

shrey150:
## TL;DR

`BROWSERBASE_API_KEY=xxx browse open <url> --remote` was **silently ignored** when the driver daemon was already running, so the CLI kept printing `Missing BROWSERBASE_API_KEY`. The daemon froze a copy of `process.env` at spawn time and never saw the late key. Fix: the client now **forwards the key with every command**, and the daemon **threads it straight into the Stagehand session at init** — no restart, no `browse stop`, warm sessions untouched.

Closes STG-2407. Reported via the partner AX update (partner-2027dev, Jun 22); confirmed still live on `main`.

---

## Symptom

```console
$ browse open https://example.com --remote          # no key set → starts the daemon, fails
Missing BROWSERBASE_API_KEY

$ BROWSERBASE_API_KEY=bb_live_… browse open https://example.com --remote   # key set, SAME daemon
Missing BROWSERBASE_API_KEY                          # ❌ ignored — only `browse stop` + retry worked
```

This burns 3–4 retries per session and blocks the documented recover-after-interruption flow.

## Root cause

The CLI is a thin client that talks to a long-lived background **daemon** (which holds the warm browser session). The key never reached that daemon:

1. **Frozen env.** The daemon is spawned `detached` with `env: process.env` captured **once** at spawn time (`daemon/client.ts`). A key exported/inlined in a *later* shell never propagates to it.
2. **Key read daemon-side.** The remote session is created inside the daemon, which read `process.env.BROWSERBASE_API_KEY` *there* (`remote.ts` → `session-manager.ts`).
3. **Protocol carried no credentials.** Requests had no way to deliver a key to an already-running daemon (`daemon/protocol.ts`). Since `--remote` is explicit, the client never blocked on the key — so it happily started a doomed key-less daemon.
4. **Stale backoff.** A cached init-failure backoff (5s→60s) replayed the *old* "missing key" error even on an immediate retry.

## The fix

**Make the client the source of truth for the key; never trust the daemon's frozen env.**

```
 client (fresh env each call)                      daemon (long-lived)
 ────────────────────────────                      ───────────────────
 collectForwardedEnv()       ──{API_KEY}──▶        stash on session manager
   reads caller's env, over the owner-only socket    │
                                                     ▼  at init only:
                                                   remoteStagehandOptions(forwardedEnv)
                                                     └▶ new Stagehand({ apiKey })
                                                          (key's only home = live session;
                                                           never written back to process.env)
```

- **`daemon/forwarded-env.ts`** (new) — `collectForwardedEnv()` reads the caller's env; `forwardedEnvSignature()` is a secret-free `sha256` fingerprint used only to detect key changes.
- **`protocol.ts` / `client.ts`** — every `open`/`command` request carries the caller's `forwardedEnv`.
- **`server.ts` / `session-manager.ts`** — the daemon threads the forwarded key **into the Stagehand constructor at init**. It is **never written into the daemon's `process.env`**. When the fingerprint changes on a *cold* session, the stale init backoff is cleared so the retry runs immediately.
- **Warm sessions are untouched** — an already-initialized session returns early and keeps its browser; the key only matters at init.

> **Only the API key is forwarded.** The Browserbase backend infers the project from the key, so `BROWSERBASE_PROJECT_ID` isn't needed for session creation — verified end-to-end with no project id set anywhere. (A multi-project key pinning a non-default project via `BROWSERBASE_PROJECT_ID` is a rare edge; that still resolves from the daemon's own env, exactly as before.)

### Naming + drift guard (per review)

The mechanism is named generically as **forwarded env vars**, not "credentials" (`ForwardedEnv`, `collectForwardedEnv` / `applyForwardedEnv` / `forwardedEnvKeys`, request field `forwardedEnv`, module `daemon/forwarded-env.ts`). It stays a **curated allowlist** (today just `BROWSERBASE_API_KEY`) rather than the whole `process.env`, because: (1) `Object.assign`-ing the full env into an already-running daemon silently no-ops for anything its modules read at import; (2) the caller's env also holds the daemon's own operational vars (`PATH`/`HOME`/`BROWSE_DAEMON_DIR`), so forwarding it wholesale risks clobbering the daemon and can't represent an unset; (3) the driver path has no AI ops, so no model keys are ever read — the API key is the only env-delivered session input. A new test, `tests/daemon-forwarded-env-drift.test.ts`, fails if any new daemon-path `process.env` read is left uncategorized (forward vs daemon-local), so the allowlist can't silently drift.

<details>
<summary>Why thread into the constructor instead of writing <code>process.env</code>?</summary>

The key is read exactly once per session (at init); afterward the live `Stagehand` instance holds it. Writing it into the daemon's global env would leave a stray secret with no reader. Threading keeps the credential scoped to the session, and the change-detector is hashed so the raw key isn't kept in a second field. There's no perf cost — forwarding ~100 bytes is free next to cloud session creation; the only thing cached for speed is the warm session, which is independent of the key value.

**Rejected alternatives:** writing the key into the daemon's env (stray secret, one-way env accumulation); auto-restarting the daemon (kills warm sessions, racy); a mere actionable error (the ask is for it to *work*, not just guide).
</details>

## Security contract (local-only build)

`BROWSERBASE_API_KEY` must not appear in the CDP-only artifact. The forwardable-key *list* is capability-gated: `forwardedEnvKeys()` returns the key in the full build (`remote.ts`) and `[]` in `remote.disabled.ts`. `collectForwardedEnv` / `forwardedEnvSignature` iterate the received object's own keys, so they stay key-name-free. Net: the literal lives only in `dist/lib/driver/remote.js`, and `tests/local-only-build.test.ts` still passes.

## Testing

Local full build (`pnpm build`) of the code under review, real Browserbase key, exercising the exact repro against an already-running key-less daemon.

| Step | Command / flow | Result | Proves |
| --- | --- | --- | --- |
| 1 | Key-less `open … --remote` (spawns daemon) | `Missing BROWSERBASE_API_KEY` (daemon stays up) | Reproduces the stranded key-less daemon |
| 2 | Inline `BROWSERBASE_API_KEY=… open … --remote`, **same** daemon, **no project id anywhere** | ✅ `SUCCESS` — `"title": "Example Domain"`, `"url": "https://example.com/"` | The inline key reaches the running daemon (the fix); project inferred from the key |
| 3 | Warm reuse: `open https://www.iana.org --remote` | ✅ `SUCCESS` — `"Internet Assigned Numbers Authority"`, same `targetId` | Warm-session fast path preserved |
| 4 | `vitest` driver-foundation + remote-disabled + local-only + drift | ✅ 45 pass (4 files) | Unit coverage; asserts the key is **not** written to `process.env` |
| 5 | `tsc -p tsconfig.local-only.json` + local-only-build test | ✅ typecheck clean | Security contract held (no key name in CDP-only build) |
| 6 | `pnpm lint` | ✅ format + eslint + tsc clean | No regressions |
| 7 | Drift guard: inject an uncategorized `process.env.X` on the daemon path | ✅ test goes red with "forward vs daemon-local" guidance, green after revert | The guard is non-vacuous — a new daemon-path env read can't slip through |

## Follow-ups (not in this PR)

- **`browse open` ECONNREFUSED** (also AX-flagged): `sendDriverRequest` (`client.ts`) has no connect-retry, so a transient `ECONNREFUSED`/`ENOENT` (stale socket / daemon mid-shutdown) propagates raw. Couldn't reproduce under load — tracking separately rather than shipping unverified.
- **Fail-fast `--remote` guard** (defense in depth): make explicit `--remote` resolve the key client-side like `autoSelectRemoteTarget` already does, so a key-less first call fails fast instead of spawning a doomed daemon. Forwarding alone fixes the reported bug; the guard would just improve the first-call error.

### DIFF
--- a/packages/cli/src/lib/driver/daemon/forwarded-env.ts
+++ b/packages/cli/src/lib/driver/daemon/forwarded-env.ts
@@ -3,12 +3,12 @@ import { createHash } from "node:crypto";
 import { getRemote } from "../remote-binding.js";
 
 /**
- * Credentials the client forwards to a running daemon so that an inline or
+ * Env vars the client forwards to a running daemon so that an inline or
  * exported API key set *after* the daemon started is still honored.
  *
  * The daemon is a detached background process whose `process.env` is captured
  * once at spawn time. A key set on a *later* client invocation never reaches
- * that process on its own, so the client ships the relevant credentials over
+ * that process on its own, so the client ships the relevant env vars over
  * the (localhost, owner-only) driver socket with every command. The daemon
  * threads them straight into the Stagehand constructor at init — it never
  * writes them back into its own `process.env`, so the key's only home is the
@@ -21,36 +21,36 @@ import { getRemote } from "../remote-binding.js";
  * `remote.ts`. The signature side iterates the received object's own keys, so
  * it needs no key list and stays key-name-free.
  */
-export type ForwardedCredentials = Record<string, string>;
+export type ForwardedEnv = Record<string, string>;
 
-/** Collect the forwardable credentials that are set in the caller's env. */
-export async function collectClientCredentials(
+/** Collect the forwardable env vars that are set in the caller's env. */
+export async function collectForwardedEnv(
   env: NodeJS.ProcessEnv = process.env,
-): Promise<ForwardedCredentials | undefined> {
-  const keys = (await getRemote()).forwardedCredentialKeys();
-  const credentials: ForwardedCredentials = {};
+): Promise<ForwardedEnv | undefined> {
+  const keys = (await getRemote()).forwardedEnvKeys();
+  const forwardedEnv: ForwardedEnv = {};
   for (const key of keys) {
     const value = env[key];
     if (typeof value === "string" && value.length > 0) {
-      credentials[key] = value;
+      forwardedEnv[key] = value;
     }
   }
-  return Object.keys(credentials).length > 0 ? credentials : undefined;
+  return Object.keys(forwardedEnv).length > 0 ? forwardedEnv : undefined;
 }
 
 /**
- * Stable, secret-free fingerprint of a forwarded credential set, used only to
- * detect whether the caller's credentials changed between requests (so a cold
+ * Stable, secret-free fingerprint of a forwarded env set, used only to
+ * detect whether the caller's forwarded env changed between requests (so a cold
  * session can bust its cached init-failure backoff and retry with the new key).
  * Hashing keeps the raw key out of any retained field. Iterates the received
  * object's own keys — the client already filtered to the forwardable set — so
  * this carries no literal key names. Returns "" for an empty/absent set.
  */
-export function credentialSignature(
-  credentials: ForwardedCredentials | undefined,
+export function forwardedEnvSignature(
+  forwardedEnv: ForwardedEnv | undefined,
 ): string {
-  if (!credentials) return "";
-  const entries = Object.entries(credentials)
+  if (!forwardedEnv) return "";
+  const entries = Object.entries(forwardedEnv)
     .filter(([, value]) => typeof value === "string" && value.length > 0)
     .sort(([a], [b]) => (a < b ? -1 : a > b ? 1 : 0));
   if (entries.length === 0) return "";
--- a/packages/cli/src/lib/driver/daemon/server.ts
+++ b/packages/cli/src/lib/driver/daemon/server.ts
@@ -132,6 +132,7 @@ async function handleLine(
 
   try {
     if (request.type === "open") {
+      manager.applyForwardedCredentials(request.credentials);
       await writeResponse(socket, {
         data: await manager.execute("open", {
           timeoutMs: request.timeoutMs,
@@ -145,6 +146,7 @@ async function handleLine(
     }
 
     if (request.type === "command") {
+      manager.applyForwardedCredentials(request.credentials);
       await writeResponse(socket, {
         data: await manager.execute(request.command, request.params),
         id: request.id,
--- a/packages/cli/src/lib/driver/daemon/server.ts
+++ b/packages/cli/src/lib/driver/daemon/server.ts
@@ -132,7 +132,7 @@ async function handleLine(
 
   try {
     if (request.type === "open") {
-      manager.applyForwardedCredentials(request.credentials);
+      manager.applyForwardedEnv(request.forwardedEnv);
       await writeResponse(socket, {
         data: await manager.execute("open", {
           timeoutMs: request.timeoutMs,
@@ -146,7 +146,7 @@ async function handleLine(
     }
 
     if (request.type === "command") {
-      manager.applyForwardedCredentials(request.credentials);
+      manager.applyForwardedEnv(request.forwardedEnv);
       await writeResponse(socket, {
         data: await manager.execute(request.command, request.params),
         id: request.id,
--- a/packages/cli/src/lib/driver/session-manager.ts
+++ b/packages/cli/src/lib/driver/session-manager.ts
@@ -8,9 +8,9 @@ import {
 import { executeDriverCommand } from "./commands/registry.js";
 import type { DriverCommandName } from "./commands/types.js";
 import {
-  credentialSignature,
-  type ForwardedCredentials,
-} from "./daemon/credentials.js";
+  forwardedEnvSignature,
+  type ForwardedEnv,
+} from "./daemon/forwarded-env.js";
 import { DriverError } from "./errors.js";
 import { discoverLocalCdp } from "./local-cdp-discovery.js";
 import { NetworkCapture } from "./network-capture.js";
@@ -71,8 +71,8 @@ export class DriverSessionManager {
 
   private consecutiveInitFailures = 0;
   private context: DriverContext | null = null;
-  private lastCredentialSignature: string | null = null;
-  private pendingCredentials: ForwardedCredentials | undefined;
+  private lastForwardedEnvSignature: string | null = null;
+  private pendingEnv: ForwardedEnv | undefined;
   private initFailure: InitFailure | null = null;
   private initPromise: Promise<void> | null = null;
   private refMaps: RefMaps = emptyRefMaps();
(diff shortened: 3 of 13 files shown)


packages/cli/src/lib/driver/session-manager.ts
```diff
@@ -90,6 +95,31 @@ export class DriverSessionManager {
...
+   * untouched. When the credentials change *before* a successful init (the
+   * common case: a first key-less `open` failed, then a key is supplied), clear
+   * the cached init failure and backoff so the retry runs immediately with the
+   * new key instead of replaying the stale missing-key error.
+   */
+  applyForwardedCredentials(
+    credentials: ForwardedCredentials | undefined,
+  ): void {
+    const signature = applyForwardedCredentials(credentials);
+    if (signature === this.credentialSignature) return;
```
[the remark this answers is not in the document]
shrey150: **Triage (not auto-fixing — design judgment needed):**

Cubic identified by cubic is right that the sync is one-way:  only sets present keys, never deletes. However this is intentional design: the PR solves the case where a key is *added* inline after the daemon started (the AX-reported bug). Propagating key *removals* is a separate, rarely-needed behavior that would require the client to send a "delete this key" signal (or the full forwardable key list so the daemon can diff). Implementing it here would: (1) require the daemon to know which keys are forwardable (violating the security contract — key names must stay out of the apply-side), or (2) require the client to always send the full set including absent keys.

**Recommendation for Shrey:** Leave one-way sync as-is for this PR. If there's a real user-facing bug from key removal (unset env mid-session), track it as a follow-up. The credential-signature mechanism will detect the change and clear backoff on the next keyed attempt regardless.

packages/cli/src/lib/driver/daemon/forwarded-env.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/** Collect the forwardable credentials that are set in the caller's env. */
+export async function collectClientCredentials(
+  env: NodeJS.ProcessEnv = process.env,
+): Promise<ForwardedCredentials | undefined> {
+  const keys = (await getRemote()).forwardedCredentialKeys();
+  const credentials: ForwardedCredentials = {};
+  for (const key of keys) {
+    const value = env[key];
+    if (typeof value === "string" && value.length > 0) {
```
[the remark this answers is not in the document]
shrey150: **Triage (not auto-fixing — design judgment needed):**

Cubic (identified by cubic) has a point about the in-flight init race: if `init()` is running when credentials change, `applyForwardedCredentials` clears `initFailure`/`consecutiveInitFailures`, but the in-flight `initPromise` is still pending. When that in-flight init *fails* (e.g. with the old key), it increments `consecutiveInitFailures` back to 1 and sets a new 5s backoff — so the backoff is not fully avoided in this race window.

However: (1) This is narrow — credentials must change *during* an in-flight init. (2) The post-race backoff starts from 0 (only 5s) since `consecutiveInitFailures` was reset. (3) The main goal — clear cached failure and immediately retry when a key is supplied after a key-less failure — works correctly for the common sequential case.

**Recommendation for Shrey:** Real edge case but low severity (5s delay in a narrow race vs. the 5–60s delay the PR fixes). Safe to leave as a follow-up, or fix here if you want it fully airtight.

packages/cli/src/lib/driver/session-manager.ts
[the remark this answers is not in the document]
shrey150: Decision: accept as a low-severity follow-up — resolving. The fix clears the stale 5–60s init backoff for the common sequential path (the actual reported bug). The flagged window is only when credentials change *during* an in-flight `init()`; there the failing in-flight promise re-arms a single 5s backoff that self-heals on the next retry — far short of the 5–60s this PR eliminates. Not worth complicating the credential-apply path for that edge; noting it as a minor follow-up.

packages/cli/src/lib/driver/daemon/forwarded-env.ts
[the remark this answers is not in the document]
shrey150: Decision: intentional / won't-fix — resolving. This PR targets key *addition* to a running daemon (the reported bug), and switching keys also works since apply overwrites. The only gap is *removal* — a client that unsets a key won't make the daemon forget the prior value. Propagating removals cleanly would require the daemon to know the forwardable key list (breaking the local-only/CDP build's key-name-free contract enforced by `local-only-build.test.ts`) or a per-command "absent keys" signal. Narrow edge (unsetting a key mid-daemon-life), low impact — tracking as a follow-up if it ever bites.

packages/cli/src/lib/driver/daemon/server.ts
```diff
@@ -145,6 +146,7 @@ async function handleLine(
     }
 
     if (request.type === "command") {
+      manager.applyForwardedCredentials(request.credentials);
```
pirate: should this be generic like `manager.applyUpdatedEnvVars(request.env)` or something?

packages/cli/tests/driver-foundation.test.ts
```diff
@@ -933,6 +937,102 @@ describe("driver foundation", () => {
...
     );
   });
+
+  it("collects forwardable credentials from the caller's env", async () => {
+    // Only the API key is forwarded; the project is inferred from the key.
+    await expect(
+      collectClientCredentials({
+        BROWSERBASE_API_KEY: "client-key",
+        BROWSERBASE_PROJECT_ID: "client-project",
+        UNRELATED: "ignored",
```
pirate, line 947: any particular reason to ignore caller env vars? wouldnt users expect all env vars to apply if they dont know about the internal daemon architecture?

pirate (approved):

packages/cli/src/lib/driver/daemon/server.ts
shrey150: yeah I like that naming better

shrey150: Done — renamed the whole mechanism `credentials → forwardedEnv`: type `ForwardedEnv`, `collectForwardedEnv` / `applyForwardedEnv` / `forwardedEnvKeys`, request field `forwardedEnv`, and the module is now `daemon/forwarded-env.ts`. One deliberate diff from `request.env`: the payload stays a curated allowlist (today just `BROWSERBASE_API_KEY`), not the whole environment — reasoning in the thread below / PR body. Addressed in afbd3a9e2.

packages/cli/tests/driver-foundation.test.ts
shrey150: For the record — implemented in afbd3a9e2: kept a curated allowlist (just `BROWSERBASE_API_KEY` today) rather than forwarding the full env. Three reasons, per our chat: (1) `Object.assign`-ing the whole env into an already-running daemon silently no-ops for anything read at import (proxy/TLS/etc.) — false transparency; (2) the caller env also carries the daemon's own operational vars (`PATH`/`HOME`/`BROWSE_DAEMON_DIR`), so wholesale forwarding risks clobbering the daemon and can't represent an unset; (3) the driver path has no AI ops (the one `stagehand.act` is deterministic selector/method), so no model keys are ever read. Added `tests/daemon-forwarded-env-drift.test.ts` so any new daemon-path env read fails CI until consciously categorized.

Issue #2272, opened 2026-06-24
Add Domain whitelisting for a stagehand session


garrettgruss-realpage:
I want to enforce a guardrail to terminate a session if an agent clicks on a link that is outside of an allowed list of domains. See claude proposal:

When running an autonomous agent.execute() loop, the agent can freely navigate to any URL. The goto tool calls page.goto(url) with no validation ([goto.ts:25](/packages/core/lib/v3/agent/tools/goto.ts#L25)), as does the CUA handler ([v3CuaAgentHandler.ts:573](/packages/core/lib/v3/handlers/v3CuaAgentHandler.ts#L573)). There is currently no built-in way to constrain an agent to a set of approved domains, or to terminate a session when it "wanders" off-domain (via a tool call, an HTTP redirect, or JS-driven window.location changes).

For regulated / security-sensitive automation, letting an LLM-driven agent navigate to arbitrary domains is a real risk (data exfiltration to an unexpected host, following a malicious link/redirect, credential entry on a look-alike domain). Today the only mitigation is to roll your own check.

Describe the solution you'd like

A first-class config option to declare allowed domains and a policy for what happens on violation. Proposed shape:

const stagehand = new Stagehand({
  // ...
  navigationGuardrail: {
    allowedDomains: ["example.com", "*.example.com"], // exact + wildcard subdomains
    onViolation: "abort", // "abort" | "block" | "warn"
  },
});
"abort" — terminate the run, surfacing the existing AgentAbortError with a descriptive reason.
"block" — cancel the navigation/keep the agent on the current page and feed an error back to the model.
"warn" — log only.
The most robust enforcement seam is Page.onFrameNavigated ([page.ts:456](packages/core/lib/v3/understudy/page.ts#L456)), since it catches all navigation — agent goto calls, redirects, and client-side location changes — not just tool invocations. CDP request interception (Fetch.enable / Network.setBlockedURLs, not currently used in the codebase) would enable true preventive "block" behavior.

Describe alternatives you've considered

The hooks exist to approximate this today, but each has gaps:

signal + onStepFinish ([agent.ts:305](packages/core/lib/v3/types/public/agent.ts#L305), [agent.ts:126](packages/core/lib/v3/types/public/agent.ts#L126)): check page.url() after each step and call controller.abort() if off-domain. Works, but is reactive — the off-domain page has already loaded, and the check only runs at step boundaries, so one off-domain action can slip through before the abort. It also misses mid-step redirects.

[...]

PR #2283, opened 2026-06-27
[feat]: add domain allowlist

seanmcguire12:
# why

users need a way to restrict browser sessions to known domains. addresses #2272 

# what changed

- users can now call `stagehand.context.setDomainPolicy({ allowedDomains })` to block HTTP(S) requests that do not match the allowed domains
- `DomainPolicy` now supports both `allowedDomains` & `blockedDomains`
- blocked domains take precedence over allowed domains when both are provided
- allowlist policies use broad HTTP(S) `Fetch` request patterns because allowlists require checking every HTTP(S) request
- blocklist-only policies keep the existing optimized behavior of only pausing requests that match generated blocklist patterns
- domain policy matching also returns a decision with the block reason, so logs can distinguish `allowedDomains` blocks from `blockedDomains` blocks

### behavioural notes

- `setDomainPolicy({ allowedDomains: [...] })` applies to active context sessions & future pages/targets
- exact domains like `example.com` allow only that hostname
- wildcard domains like `*.example.com` allow subdomains, but not the apex domain
- requests to domains outside `allowedDomains` are failed with `BlockedByClient`
- if a request matches `blockedDomains`, it is blocked even if it also matches `allowedDomains`
- non-HTTP(S) urls continue instead of being blocked by the domain policy
- `setDomainPolicy()` replaces the current policy; callers should pass both `allowedDomains` & `blockedDomains` in the same call when they want both
- clearing with `setDomainPolicy(null)`, `{ allowedDomains: [] }`, or an otherwise empty policy disables the policy when no blocked domains are present

# test plan

- `packages/core/tests/unit/domain-policy.test.ts` validates allowlist normalization, exact/wildcard matching, invalid allowed domain rejection, block-over-allow precedence, non-HTTP(S) continuation & broad HTTP(S) `Fetch.RequestPattern` generation
- `packages/core/tests/unit/context-domain-policy.test.ts` validates `context.setDomainPolicy()` enables broad HTTP(S) interception for allowlists, preserves precise blocklist-only interception, returns normalized policy state, continues allowed paused requests & fails disallowed paused requests
- `packages/core/tests/integration/context-domain-policy.spec.ts` validates allowed-domain requests are not blocked, non-matching requests fail with `ERR_BLOCKED_BY_CLIENT`, blocked domains take precedence over allowed domains & allowlist policies apply to pages created after the policy is set

---
## Summary by cubic
Add a context-wide domain policy to control which HTTP(S) requests Stagehand can access. You can allow specific domains or block known hosts, with strict enforcement across existing and new pages.

- **New Features**
  - Added `context.setDomainPolicy({ allowedDomains?: string[], blockedDomains?: string[] })` and `context.getDomainPolicy()`. Domain-only patterns; exact or `*.example.com` (wildcards do not match the apex). HTTP(S) only. Applies to current and future pages. Clear with `null` or empty arrays; `getDomainPolicy()` returns only set fields.
  - Enforcement: `blockedDomains` take precedence over `allowedDomains`. Allowlists use broad HTTP(S) `Fetch` interception; blocklist-only policies keep precise patterns. Logs include the block reason (`allowedDomains` vs `blockedDomains`).
  - Errors & validation: Added `StagehandSetDomainPolicyError` when enforcement can’t be guaranteed during target attach. Added runtime validation; malformed inputs (omitted policy, non-array fields, non-string or non-domain patterns) throw `StagehandInvalidArgumentError`.

<sup>Written for commit 60bafc3cefd2d2ba257dbeb8acfda1246e148b7b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2283?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/itchy-actors-kneel.md
+++ b/.changeset/itchy-actors-kneel.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": minor
+---
+
+add `context.setDomainPolicy({ allowedDomains: ["allowed.domain"] })` which allows users to define a set of domains that are accessible to stagehand
--- a/packages/core/lib/v3/types/public/context.ts
+++ b/packages/core/lib/v3/types/public/context.ts
@@ -35,6 +35,9 @@ export interface ClearCookieOptions {
 
 /** Context-wide network domain policy. */
 export interface DomainPolicy {
+  /** Domain-only allow patterns, e.g. "example.com" or "*.example.com". */
+  allowedDomains?: string[];
+
   /** Domain-only block patterns, e.g. "example.com" or "*.example.com". */
   blockedDomains?: string[];
 }
--- a/packages/core/lib/v3/understudy/context.ts
+++ b/packages/core/lib/v3/understudy/context.ts
@@ -32,7 +32,10 @@ import {
   CookieParam,
   DomainPolicy,
 } from "../types/public/context.js";
-import { normalizeDomainPolicy, shouldBlockUrl } from "./domainPolicy.js";
+import {
+  getDomainPolicyDecision,
+  normalizeDomainPolicy,
+} from "./domainPolicy.js";
 import type { NormalizedDomainPolicy } from "./domainPolicy.js";
 
 type TargetId = string;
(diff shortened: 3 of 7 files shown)


seanmcguire12:
@cubic-dev-ai

akeimach (approved):
fantastic

PR #2296, opened 2026-06-30
[STG-2450] feat(cli): browse snapshot lean by default, add --full for ref maps

shrey150:
## Summary

`browse snapshot` now prints the formatted accessibility tree only by default, omitting the `xpathMap`/`urlMap` ref maps that were previously included on **every** call.

- **`browse snapshot`** (default): formatted tree only (~14KB on a content-heavy page).
- **`browse snapshot --full`**: tree + `xpathMap` + `urlMap`.
- **`browse snapshot --compact`**: deprecated no-op alias of the default (prints a stderr-only, TTY-gated deprecation notice).
- **`browse refs`**: prints the cached maps on demand.

Ref-based element commands (`click`, `fill`, `select`, …) are **unaffected** — the maps are still captured and cached server-side, so refs resolve exactly as before.

## Why

The default snapshot emitted **~241KB (~60K tokens)** on a content-heavy page, of which **~217KB (90%) is the `xpathMap`** (a ref→XPath entry for every node) + `urlMap`. The tree an agent actually reasons over is only ~13.5KB.

Because refs resolve from a **server-side cache** (not from stdout), the printed maps were dead weight in the context of any agent consuming `browse snapshot` output. Making the tree lean by default cuts the default payload ~17× with no loss to element interaction; `--full` and `browse refs` recover the maps when needed.

### Behavior + size (local build, live session on capeair.com)

Verified against `packages/cli/bin/run.js` (the local build), not the published `browse`.

| --- | --- | --- |
| `browse snapshot` (default) | `13,845` bytes, keys = `tree` only, `hasMaps=false` | Lean default; ~17× smaller than before |
| `browse snapshot --full` | `241,104` bytes, keys = `tree,urlMap,xpathMap` (1,640 entries) | `--full` returns the full tree + maps |
| `browse snapshot --compact` (piped) | `13,868` bytes, `hasMaps=false`, **0 bytes stderr** | Deprecated alias == default; no noise on non-TTY |
| `browse refs` | `count=1640` | Maps still captured + retrievable on demand |
| `browse click @0-1588` after a **lean** snapshot | `{"clicked": true}` → navigated to `/about_us/` | Ref interaction unaffected by dropping maps from stdout |
| default tree vs `--full` tree | identical (341 lines both) | Lean mode omits **only** the maps; it does not prune tree content |
| `pnpm --dir packages/cli build` (`tsc`) + evals `tsc --noEmit` | success / clean | Typechecks |
| `driver-commands` unit test | 15/15 pass | Locks in: default omits maps, `--full` includes them, maps cached in both modes |

### Agent task-success A/B (lean vs full)

Controlled A/B — same task / model (Sonnet) / session per pair; **the only variable is snapshot mode** (`--full` = maps vs default = lean), against the local build.

| Task | Arm | Outcome | Snapshots | Agent tokens |
| --- | --- | --- | --- | --- |
| capeair (regional booking) | full (maps) | ✅ SUCCESS | 9 | 109.8K |
| capeair | **lean** | ✅ SUCCESS | 8 | 93.5K |
| Google Flights | full (maps) | ✅ SUCCESS | 15 | 85.3K |
| Google Flights | **lean** | ✅ SUCCESS | 15 | 86.4K |

**4/4 success — lean-by-default held task success with zero capability loss.** (A standard-benchmark A/B via the evals package is in progress and will be added here.)

## What's in the diff

- `packages/cli/src/commands/snapshot.ts` — lean default (`compact` decoupled from map omission); add `--full`; deprecate `--compact`.
- `packages/cli/src/lib/driver/commands/snapshot.ts` — driver handler emits the full tree and includes ref maps only when `full` is requested.
- `packages/cli/skills/browse/SKILL.md` — document lean default + `--full`.
- `packages/cli/tests/driver-commands.test.ts` — test the default/`--full`/caching behavior.
- `packages/evals/core/tools/browse_cli.ts` — derive `refCount` from the tree when maps are absent.
- `.changeset/lean-browse-snapshot.md` — `browse` patch.

Linear: [STG-2450](https://linear.app/browserbase/issue/STG-2450)

### DIFF
--- a/packages/cli/src/commands/snapshot.ts
+++ b/packages/cli/src/commands/snapshot.ts
@@ -8,7 +8,7 @@ import {
 
 export default class Snapshot extends BrowseCommand {
   static override description =
-    "Print the active page accessibility snapshot and cache refs for element commands. Pass --full to also include the ref maps (xpathMap, urlMap), or run `browse refs` to print them.";
+    "Print the active page accessibility snapshot and cache refs for element commands. Pass --full to also include the ref maps (xpathMap, urlMap).";
 
   static override examples = [
     "browse snapshot",
--- a/packages/cli/src/commands/snapshot.ts
+++ b/packages/cli/src/commands/snapshot.ts
@@ -8,19 +8,24 @@ import {
 
 export default class Snapshot extends BrowseCommand {
   static override description =
-    "Print the active page accessibility snapshot and cache refs for element commands.";
+    "Print the active page accessibility snapshot and cache refs for element commands. Lean by default (formatted tree only); pass --full to also include the ref maps (xpathMap, urlMap). The ref maps are always cached for element commands regardless, and can be printed on demand with `browse refs`.";
 
   static override examples = [
     "browse snapshot",
-    "browse snapshot --compact",
+    "browse snapshot --full",
     "browse snapshot --filter submit",
     "browse snapshot --max-depth 4",
   ];
 
   static override flags = {
     ...driverCommandFlags,
+    full: Flags.boolean({
+      description:
+        "Include the ref maps (xpathMap, urlMap) in addition to the formatted tree. Restores the pre-lean default output.",
+    }),
     compact: Flags.boolean({
-      description: "Print only the formatted tree, without ref maps.",
+      description:
+        "Deprecated: snapshots are lean by default, so this flag has no effect. Use --full to include the ref maps.",
     }),
     filter: Flags.string({
       description:
@@ -35,10 +40,20 @@ export default class Snapshot extends BrowseCommand {
 
   async run(): Promise<void> {
     const { flags } = await this.parse(Snapshot);
+    // `--compact` is now the default. Keep accepting it as a no-op alias so
+    // existing callers don't break, but nudge interactive users toward --full.
+    // Warn on stderr only, gated on a TTY, so agents parsing stdout JSON (or
+    // capturing stderr) get no extra noise.
+    if (flags.compact && process.stderr.isTTY) {
+      this.warn(
+        "`--compact` is deprecated: `browse snapshot` is lean by default. Use `--full` to include the ref maps.",
+      );
+    }
     await runDriverCommandFromFlags(
       "snapshot",
       {
-        compact: flags.compact,
+        // Lean (no ref maps) unless --full is requested.
+        compact: !flags.full,
         filter: flags.filter,
         maxDepth: flags["max-depth"],
       },
--- a/packages/cli/src/commands/snapshot.ts
+++ b/packages/cli/src/commands/snapshot.ts
@@ -8,7 +8,7 @@ import {
 
 export default class Snapshot extends BrowseCommand {
   static override description =
-    "Print the active page accessibility snapshot and cache refs for element commands. Lean by default (formatted tree only); pass --full to also include the ref maps (xpathMap, urlMap). The ref maps are always cached for element commands regardless, and can be printed on demand with `browse refs`.";
+    "Print the active page accessibility snapshot and cache refs for element commands. Pass --full to also include the ref maps (xpathMap, urlMap), or run `browse refs` to print them.";
 
   static override examples = [
     "browse snapshot",
@@ -20,12 +20,11 @@ export default class Snapshot extends BrowseCommand {
   static override flags = {
     ...driverCommandFlags,
     full: Flags.boolean({
-      description:
-        "Include the ref maps (xpathMap, urlMap) in addition to the formatted tree. Restores the pre-lean default output.",
+      description: "Also include the ref maps (xpathMap, urlMap).",
     }),
     compact: Flags.boolean({
       description:
-        "Deprecated: snapshots are lean by default, so this flag has no effect. Use --full to include the ref maps.",
+        "Deprecated and has no effect; use --full to include the ref maps.",
     }),
     filter: Flags.string({
       description:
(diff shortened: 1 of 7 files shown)


packages/cli/src/commands/snapshot.ts
```diff
@@ -8,19 +8,24 @@ import {
 
 export default class Snapshot extends BrowseCommand {
   static override description =
-    "Print the active page accessibility snapshot and cache refs for element commands.";
+    "Print the active page accessibility snapshot and cache refs for element commands. Lean by default (formatted tree only); pass --full to also include the ref maps (xpathMap, urlMap). The ref maps are always cached for element commands regardless, and can be printed on demand with `browse refs`.";
```
shrey150: description seems a little verbose / historical, want to simplify

packages/cli/src/commands/snapshot.ts
```diff
@@ -35,10 +40,20 @@ export default class Snapshot extends BrowseCommand {
 
   async run(): Promise<void> {
     const { flags } = await this.parse(Snapshot);
+    // `--compact` is now the default. Keep accepting it as a no-op alias so
```
shrey150: unnecessary comment I feel

packages/cli/src/commands/snapshot.ts
shrey150: Addressed in [4060e0d](https://github.com/browserbase/stagehand/commit/4060e0dc36cb065845d2ea969b83300ec635b973) — simplified to present-tense behavior and dropped the historical framing ("pre-lean default" / "now the default").

packages/cli/src/commands/snapshot.ts
shrey150: Addressed in [4060e0d](https://github.com/browserbase/stagehand/commit/4060e0dc36cb065845d2ea969b83300ec635b973) — removed the comments here and the block in `run()`.

packages/cli/src/commands/snapshot.ts
```diff
@@ -35,10 +40,20 @@ export default class Snapshot extends BrowseCommand {
...
+      this.warn(
+        "`--compact` is deprecated: `browse snapshot` is lean by default. Use `--full` to include the ref maps.",
+      );
+    }
     await runDriverCommandFromFlags(
       "snapshot",
       {
-        compact: flags.compact,
+        // Lean (no ref maps) unless --full is requested.
+        compact: !flags.full,
```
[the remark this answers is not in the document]
shrey150: Addressed in [4060e0d](https://github.com/browserbase/stagehand/commit/4060e0dc36cb065845d2ea969b83300ec635b973) — decoupled. The default now emits the full, **unpruned** tree and only omits the ref maps (verified `default tree === --full tree`, 341 lines in both). `--full` re-adds xpathMap/urlMap. The driver handler keys off `full` instead of overloading `compact`, so the old `compactLines` pruning no longer applies to the default path.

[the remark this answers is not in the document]
shrey150: Addressed in [4060e0d](https://github.com/browserbase/stagehand/commit/4060e0dc36cb065845d2ea969b83300ec635b973) — added a `driver-commands` test: default omits ref maps, `--full` includes them, and the maps are cached for element commands in both modes. (Behavior also covered end-to-end by a live smoke run + the lean-vs-full agent A/B in the PR body.)

packages/cli/tests/driver-commands.test.ts
```diff
@@ -335,6 +336,36 @@ describe("driver commands", () => {
...
+    expect(lean).not.toHaveProperty("xpathMap");
+    expect(lean).not.toHaveProperty("urlMap");
+
+    const full = await snapshotHandlers.snapshot!(manager, { full: true });
+    expect(full).toMatchObject({
+      urlMap: snap.urlMap,
+      xpathMap: snap.xpathMap,
+    });
+
+    expect(setRefMaps).toHaveBeenCalledWith({
```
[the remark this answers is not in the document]
shrey150: Addressed in [ab29d17](https://github.com/browserbase/stagehand/commit/ab29d17331cde9d7f83d91516a4d0080af36b50b) — now asserts `setRefMaps` is called (with the maps) on the lean call specifically via `toHaveBeenCalledTimes(1)` + `toHaveBeenLastCalledWith`, then `toHaveBeenCalledTimes(2)` after `--full`. Moving caching into the `--full` path would now fail the test.

packages/cli/src/commands/snapshot.ts
```diff
@@ -8,19 +8,23 @@ import {
 
 export default class Snapshot extends BrowseCommand {
   static override description =
-    "Print the active page accessibility snapshot and cache refs for element commands.";
+    "Print the active page accessibility snapshot and cache refs for element commands. Pass --full to also include the ref maps (xpathMap, urlMap), or run `browse refs` to print them.";
```
seanmcguire12: does `browse refs` still exist?

shrey150: oops we merged the `refs` removal before this, will remove from the help command

seanmcguire12:
@shrey150 dont forget to update the readme. it still says `--compact`

packages/cli/src/commands/snapshot.ts
shrey150: Removed in [5c20c2d](https://github.com/browserbase/stagehand/commit/5c20c2ddb16ab75657630f10d6fdf6dc6f3626c9) — dropped the `browse refs` mention from the `snapshot` help description.

shrey150:
Good catch — fixed in [8754565](https://github.com/browserbase/stagehand/commit/8754565d595bb836e4d5b55f1aeb95e5dba67111): README quick-start + snapshot examples now use `browse snapshot` (lean by default) and `--full` instead of the deprecated `--compact`. `--compact` is gone from all tracked docs (SKILL.md was already correct).

seanmcguire12 (approved):

PR #2304, opened 2026-07-02
fix(agent): enable prompt caching for the Anthropic CUA system prompt

aditya-silna:
## Why

In `AnthropicCUAClient.getAction()`, the system prompt was sent as a plain string:

```ts
requestParams.system = this.userProvidedInstructions;
```

A CUA agent loop re-sends the identical tools + system prefix on every step, so without a cache breakpoint the full prefix is re-processed at full input price each step.

## What changed

The system prompt is now sent as a content block array with `cache_control`:

```ts
requestParams.system = [
  {
    type: "text",
    text: this.userProvidedInstructions,
    cache_control: { type: "ephemeral" },
  },
];
```

Since the Anthropic API renders the prompt as `tools` → `system` → `messages`, this single breakpoint caches the computer-use tool definition and the system prompt together across steps. Cache reads bill at ~0.1× input price (writes at 1.25×), so this pays for itself from the second step of any run. If the prefix is below the model's minimum cacheable size, the marker is silently ignored — no error, no premium.

Caching is transparent to model behavior; outputs are unchanged.

## Testing

- `tsc --noEmit` on `packages/core` passes.

---
## Summary by cubic
Enable prompt caching for Anthropic CUA by sending the system prompt as a content block array with `cache_control: { type: "ephemeral" }` in getAction(). This caches the tools + system prefix across agent steps to cut input cost and latency, and adds a unit test to verify the `system` payload shape (cached array when instructions exist, omitted otherwise).

<sup>Written for commit b6feabf8c90dc91802a1b35c17cd0bb29a744112. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2304?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/core/lib/v3/agent/AnthropicCUAClient.ts
+++ b/packages/core/lib/v3/agent/AnthropicCUAClient.ts
@@ -554,9 +554,16 @@ export class AnthropicCUAClient extends AgentClient {
         ];
       }
 
-      // Add system parameter if provided
+      // Add system parameter if provided, with cache control so the
+      // tools + system prefix is cached across agent steps
       if (this.userProvidedInstructions) {
-        requestParams.system = this.userProvidedInstructions;
+        requestParams.system = [
+          {
+            type: "text",
+            text: this.userProvidedInstructions,
+            cache_control: { type: "ephemeral" },
+          },
+        ];
       }
 
       // Add thinking parameter if available
--- a/packages/core/tests/unit/anthropic-cua-system-prompt-caching.test.ts
+++ b/packages/core/tests/unit/anthropic-cua-system-prompt-caching.test.ts
@@ -1,0 +1,82 @@
+import { describe, expect, it, vi, beforeEach } from "vitest";
+import { AnthropicCUAClient } from "../../lib/v3/agent/AnthropicCUAClient.js";
+import Anthropic from "@anthropic-ai/sdk";
+
+// Mock the Anthropic SDK's beta.messages.create method
+vi.mock("@anthropic-ai/sdk", () => {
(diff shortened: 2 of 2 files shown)


packages/core/lib/v3/agent/AnthropicCUAClient.ts
```diff
@@ -554,9 +554,16 @@ export class AnthropicCUAClient extends AgentClient {
         ];
       }
 
-      // Add system parameter if provided
+      // Add system parameter if provided, with cache control so the
+      // tools + system prefix is cached across agent steps
       if (this.userProvidedInstructions) {
```
sviat-barbutsa, line 559: Hey, one follow-up case: this only creates a cache breakpoint when `userProvidedInstructions` is present.

In the default/no-custom-instructions path, the request still sends the computer-use tool definition, but there is no `system` block and no `cache_control` marker anywhere in the request. 
So, actually, the repeated tool prefix still will not be cached across agent steps.

Could you, please, add coverage for the no-instructions path too, and either put `cache_control` on the last tool definition or make it explicit that caching is only intended when custom instructions are provided?

sviat-barbutsa (commented):
Hi, everything is great and please, just check one small follow-up comment.

PR #2305, opened 2026-07-02
[STG-2463] fix: silence AI SDK "system message in messages" warning on agent.execute()

shrey150:
## What & why

Every hybrid/DOM `agent.execute()` call prints this AI SDK v5 warning to the console:

> AI SDK Warning: System messages in the prompt or messages fields can be a security risk because they may enable prompt injection attacks. Use the system option instead when possible. Set allowSystemInMessages to true to suppress this warning, or false to throw an error.

It's cosmetic (a warning, not an error), but it fires on essentially every agent run, and the "prompt injection" wording reads alarmingly enough to generate support tickets — it did (a customer reported it).

### Root cause
The agent loop builds its system prompt as a **system-role message inside the `messages` array** (`prependSystemMessage()` in `v3AgentHandler.ts`) rather than the top-level `system` param. This is deliberate: it lets the system prompt carry Anthropic **ephemeral cache-control** via `providerOptions` (see the function's own comment). AI SDK v5 warns whenever it sees a system message in `messages`, since it can't distinguish a trusted, library-authored system prompt from untrusted input.

The committed lockfile pinned `ai@5.0.133`, which predates this warning — so the repo build never saw it, but the `^5.0.133` range means real installs resolve to newer 5.0.x (e.g. `5.0.209`) that **do** warn. That's why users hit it and CI didn't.

### Fix
- Pass `allowSystemInMessages: true` to the `generateText` / `streamText` calls in the agent loop. This is the AI-SDK-sanctioned opt-out for an intentional system-in-messages structure, and it **keeps prompt caching intact** (vs. moving to top-level `system:`, which cannot carry per-message `providerOptions` and would regress Anthropic cache-control).
- Bump the `ai` floor `^5.0.133 → ^5.0.185`, where `allowSystemInMessages` is a typed option.
- Add an explicit return type (`LanguageModelV2`) to `getAISDKLanguageModel`. The `ai` bump pulls a newer `@ai-sdk/provider`, and without the annotation tsc emits `TS2742` ("inferred type … not portable").

## E2E Test Matrix

| --- | --- | --- |
| **BEFORE** — published `@browserbasehq/stagehand@3.6.0` (resolves `ai@5.0.209`), hybrid `agent.execute()` on a real Browserbase session, console captured | `AI SDK system-in-messages warning emitted: YES (1)` — full "…security risk…prompt injection…" text | Reproduces the customer's exact symptom on the version they run. |
| **AFTER** — local patched build (this branch, same `ai@5.0.209`), identical hybrid `agent.execute()` flow, same model, real Browserbase session | `AI SDK system-in-messages warning emitted: NO (0)`; run exits cleanly | Proves the fix removes the warning without breaking the agent run. Same harness as BEFORE → clean A/B. |
| `pnpm turbo run build --filter @browserbasehq/stagehand` | `Tasks: 3 successful, 3 total` (fails with `TS2742` before the `LLMProvider` annotation) | Typecheck + emit green; confirms the `ai` bump + annotation compile. |
| `prettier --check` + `eslint` on changed files | `All matched files use Prettier code style!`, eslint clean | Style/lint gates pass. |

A/B model: `anthropic/claude-haiku-4-5-20251001`. Both runs used identical `mode: "hybrid"` + `systemPrompt` config (the customer's shape).

Closes STG-2463.

---
## Summary by cubic
Silences the AI SDK v5 “system message in messages” warning during hybrid/DOM `agent.execute()` runs while preserving Anthropic cache-control. Addresses STG-2463.

- **Bug Fixes**
  - Pass `allowSystemInMessages: true` to `generateText`/`streamText` so the agent’s system prompt in `messages` does not warn and caching stays intact.

- **Dependencies**
  - Bump `ai` to `^5.0.185` (lock resolves to `5.0.209`).
  - Add explicit `LanguageModelV2` return type in `getAISDKLanguageModel` to avoid `TS2742` with newer `@ai-sdk/provider`.

<sup>Written for commit bfc10af9413bd083e239422dde63de9957f5ff33. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2305?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/aisdk-allow-system-in-messages.md
+++ b/.changeset/aisdk-allow-system-in-messages.md
@@ -2,6 +2,4 @@
 "@browserbasehq/stagehand": patch
 ---
 
-Silence the AI SDK "System messages in the prompt or messages fields can be a security risk" warning emitted on every hybrid/DOM `agent.execute()` call.
-
-The agent loop intentionally supplies its system prompt as a system-role message (rather than the top-level `system` param) so it can carry Anthropic ephemeral cache-control via `providerOptions`. AI SDK v5 warns whenever it sees a system message inside `messages`, since it can't tell a trusted system prompt from untrusted input. Because the prompt is Stagehand's own, we now pass `allowSystemInMessages: true` to the `generateText`/`streamText` calls, which keeps prompt caching intact and removes the noisy warning. Bumps the `ai` floor to `^5.0.185`, where this option is available.
+Remove the noisy AI SDK "system message in messages" warning logged on every hybrid/DOM `agent.execute()` call.
--- a/.changeset/aisdk-allow-system-in-messages.md
+++ b/.changeset/aisdk-allow-system-in-messages.md
@@ -1,0 +1,7 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+Silence the AI SDK "System messages in the prompt or messages fields can be a security risk" warning emitted on every hybrid/DOM `agent.execute()` call.
+
+The agent loop intentionally supplies its system prompt as a system-role message (rather than the top-level `system` param) so it can carry Anthropic ephemeral cache-control via `providerOptions`. AI SDK v5 warns whenever it sees a system message inside `messages`, since it can't tell a trusted system prompt from untrusted input. Because the prompt is Stagehand's own, we now pass `allowSystemInMessages: true` to the `generateText`/`streamText` calls, which keeps prompt caching intact and removes the noisy warning. Bumps the `ai` floor to `^5.0.185`, where this option is available.
--- a/packages/core/lib/v3/handlers/v3AgentHandler.ts
+++ b/packages/core/lib/v3/handlers/v3AgentHandler.ts
@@ -458,6 +458,7 @@ export class V3AgentHandler {
       const result = await this.llmClient.generateText({
         model: wrappedModel,
         messages: prependSystemMessage(systemPrompt, messages),
+        allowSystemInMessages: true,
         tools: allTools,
         stopWhen: (result) => this.handleStop(result, maxSteps),
         toolChoice: "auto",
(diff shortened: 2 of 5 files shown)


.changeset/aisdk-allow-system-in-messages.md
```diff
@@ -0,0 +1,7 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+Silence the AI SDK "System messages in the prompt or messages fields can be a security risk" warning emitted on every hybrid/DOM `agent.execute()` call.
+
+The agent loop intentionally supplies its system prompt as a system-role message (rather than the top-level `system` param) so it can carry Anthropic ephemeral cache-control via `providerOptions`. AI SDK v5 warns whenever it sees a system message inside `messages`, since it can't tell a trusted system prompt from untrusted input. Because the prompt is Stagehand's own, we now pass `allowSystemInMessages: true` to the `generateText`/`streamText` calls, which keeps prompt caching intact and removes the noisy warning. Bumps the `ai` floor to `^5.0.185`, where this option is available.
```
seanmcguire12: lets remove stuff around implementation details etc. just looking for a sentence or two explaining any customer facing changes. this one could just be something like "remove noisy aisdk warning log"

shrey150: Done — trimmed to a single customer-facing line in [\`bfc10af\`](https://github.com/browserbase/stagehand/commit/bfc10af94): "Remove the noisy AI SDK \"system message in messages\" warning logged on every hybrid/DOM \`agent.execute()\` call." Dropped all the implementation detail.

seanmcguire12 (approved):

PR #2309, opened 2026-07-02
fix(cli): surface no-match failures in browse click/fill

yawbtng:
## why

`browse click` and `browse fill` report success on a selector that matches **nothing**. Reproduced on `browse@0.9.1` against a real page:

```
$ browse click "#this-element-does-not-exist-xyz"
{ "clicked": true }
$ echo $?
0

$ browse fill "#this-element-does-not-exist-xyz" "hello"
{ "filled": true, "pressedEnter": false }
$ echo $?
0
```

Every other element command errors correctly on the same no-match selector:

```
$ browse is visible "#missing"   →  exit 1  "Could not find an element for the given xPath(s): #missing"
$ browse get text "#missing"     →  exit 1  "Could not find an element ..."
$ browse select "#missing" ...   →  exit 1  (throws via deepLocator)
$ browse upload "#missing" ...    →  exit 1  (throws via deepLocator)
```

`click`/`fill` are the only outliers. They run through `stagehand.act()`, which reports a missing element via `success:false` **rather than throwing** (`actHandler.ts`), and the handlers discard that result — so a script or agent driving the CLI proceeds on a false premise.

## what changed

`packages/cli/src/lib/driver/commands/elements.ts` — `click` and `fill` now check the `act()` result and `throw new Error(result.message)` when `success` is false, so they exit non-zero and print the same "Could not find an element" error as their siblings. A failed `fill` no longer presses Enter afterward.

## test plan

Adds `tests/element-command-failures.test.ts` (5 cases, mock manager per the existing `driver-commands.test.ts` pattern): click/fill throw on `success:false`, resolve on success, and a failed fill never presses Enter. Adjacent suites (`driver-commands`, `driver-foundation`) still green; `tsc -p tsconfig.json` clean.

The pre-fix behavior above was reproduced live; the fix is covered by the unit tests (a full end-to-end run needs a driver session).

---
## Summary by cubic
Fixes the `browse` CLI so `browse click` and `browse fill` fail with a clear error and non-zero exit when the selector matches no element, aligning with other element commands.

- **Bug Fixes**
  - Throw the `act()` error message when `success:false` instead of reporting success.
  - `fill` no longer presses Enter after a failed action.

<sup>Written for commit aee59f50d01ab54ba93d2742f1814ee5bda9efe5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2309?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/fix-cli-click-fill-nomatch.md
+++ b/.changeset/fix-cli-click-fill-nomatch.md
@@ -1,0 +1,5 @@
+---
+"browse": patch
+---
+
+`browse click` and `browse fill` now exit non-zero and report the error when their selector matches no element, instead of printing `{ "clicked": true }` / `{ "filled": true }` and exiting 0. Both commands run through `act()`, which reports a missing element via `success:false` rather than throwing, so the failure was previously swallowed — unlike `select`/`upload`, which already error via `deepLocator`. This makes the false-positive success visible to scripts and agents.
--- a/packages/cli/src/lib/driver/commands/elements.ts
+++ b/packages/cli/src/lib/driver/commands/elements.ts
@@ -8,12 +8,17 @@ export const elementsHandlers: DriverCommandHandlers = {
       .object({ selector: z.string().min(1) })
       .parse(params);
     const stagehand = await manager.stagehandInstance();
-    await stagehand.act({
+    // act() reports a failed action (e.g. a selector that matches nothing) by
+    // returning success:false rather than throwing, so surface it as an error
+    // instead of reporting { clicked: true } — matching select/upload, which
+    // throw via deepLocator when the element is missing.
+    const result = await stagehand.act({
       arguments: [],
       description: "click element",
       method: "click",
       selector: manager.resolveSelector(selector),
     } as never);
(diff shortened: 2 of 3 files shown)


yawbtng:
Obsolete as of the v4 CLI migration. The driver now calls `page.locator(selector).click()` instead of `stagehand.act()`, and `locator.click()` / `fill()` resolve the node first and throw `Could not find an element for the given xPath(s)` when nothing matches — so the unconditional `return { clicked: true }` is no longer reachable on a no-match. Closing; the false-success behavior this fixed is gone.

Note `select` still returns `{ selected }` without comparing it to the requested values, which is a separate case since a matched `<select>` with unmatched options never reaches that guard.

PR #2347, opened 2026-07-10
feat(core): support OpenAI Chat Completions via openaiEndpointFormat

miguelg719:
## Summary
- add `openaiEndpointFormat` to public model configuration
- select OpenAI Chat Completions when `openaiEndpointFormat` is `chat`
- preserve Responses API as the default
- omit the Responses-only `store` option for Chat providers, including finalization calls
- add a patch changeset

## Usage

```ts
  model: {
    modelName: "openai/databricks-claude-opus-4-6",
    apiKey: process.env.DATABRICKS_TOKEN,
    baseURL: "https://example.databricks.com/serving-endpoints",
    openaiEndpointFormat: "chat",
  },
});
```

The option names the *wire format* used against the OpenAI-compatible endpoint at `baseURL` — deliberately not `apiMode` (too generic) or `openaiEndpoint` (reads like it holds a URL, and sits right next to `baseURL`).

## Databricks verification
- basic Chat Completions request: 200
- request with `store: false` reproduced the 400: extra inputs are not permitted
- two-turn tool call and tool-result replay without `store`: 200 on both turns

## Testing
- focused llm-provider and agent suites: 19 passed
- TypeScript typecheck
- ESLint on touched TypeScript files
- Prettier check on touched files

### DIFF
--- a/.changeset/openai-chat-api-mode.md
+++ b/.changeset/openai-chat-api-mode.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+Allow OpenAI-compatible models to select the Chat Completions API with `apiMode: "chat"` in model configuration.
--- a/.changeset/openai-chat-api-mode.md
+++ b/.changeset/openai-chat-api-mode.md
@@ -2,4 +2,4 @@
 "@browserbasehq/stagehand": patch
 ---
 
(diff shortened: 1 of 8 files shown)


akeimach (approved):

akeimach (approved):
Even better than before

miguelg719:
the `e2e/local/chrome-newtab-page-tracking` test keeps failing, will address in a follow up pr

PR #2394, opened 2026-07-23
fix(core): cancel timed-out actions

GautamSharma99:
## Summary

  Stagehand’s timeout handling previously relied on `Promise.race()`. Although the caller received a `TimeoutError`, the underlying browser
  action, navigation, HTTP request, or model inference continued running.

  This could produce late side effects after an agent had already handled the timeout or started a retry. For example, a timed-out click or form
  submission could complete afterward, causing duplicate or out-of-order actions. Model requests could also continue consuming resources after
  their result was no longer usable.

  This PR changes timeouts from response-only deadlines into cooperative cancellation boundaries.

  ## What changed

  ### Introduce timeout-scoped cancellation

  `withTimeout()` now:

  - creates an `AbortController` for each active timeout;
  - aborts the controller with the appropriate timeout error when the deadline expires;
  - accepts lazy operation factories so the operation starts inside the cancellation context;
  - combines nested timeout signals with caller-provided signals;
  - preserves existing timeout error types and messages;
  - rejects immediately when a parent operation is already aborted.

  Cancellation context is propagated through `AsyncLocalStorage`, avoiding invasive `AbortSignal` parameters across every internal method while
  still allowing lower-level browser and provider integrations to inspect the active signal.

  The existing promise-based `withTimeout()` signature remains supported for compatibility, while cancellation-aware call sites now use the lazy
  factory form.

  ### Cancel agent tools and public handlers

  Timeout signals now cover:

  - agent tool execution;
  - `act()`;
  - `extract()`;
  - `observe()`;
  - deterministic cached actions;
  - screenshots;
  - local handler execution;
  - Stagehand API-backed handler calls.

  The AI SDK’s tool execution `abortSignal` is combined with the configured per-tool timeout, so cancellation from either the overall agent
  execution or the individual tool deadline reaches the underlying operation.

  Handler timeout guards also inspect the active signal at checkpoints, preventing processing from resuming after an aborted call settles.

  ### Cancel model inference

  The active signal is forwarded to supported provider requests:

  - OpenAI;
  - Anthropic;
  - Google;
  - Groq;
  - Cerebras;
  - AI SDK language models.

  This allows in-flight inference requests to terminate instead of continuing after Stagehand has discarded their result.

  The `LLM_MAX_MS` timeout path now starts inference within the cancellation context as well.

  ### Prevent late browser commands

  CDP command dispatch checks the active cancellation signal before sending a command. If a timed-out operation resumes later, it cannot issue
  additional clicks, keyboard input, DOM mutations, or other browser commands.

  Navigation methods additionally:

  - race lifecycle and response waits against cancellation;
  - send a best-effort `Page.stopLoading` when aborted;
  - dispose navigation watchers and abort listeners during cleanup.

  Commands already delivered to Chrome cannot be recalled. This PR therefore stops active navigation where possible and quarantines the
  remainder of an aborted operation so it cannot dispatch subsequent commands.

  ### Clean up partially completed input sequences

  Cancellation can occur between the two halves of a compound input operation. This PR adds best-effort cleanup for:

  - mouse press/release during drag and click-and-hold;
  - key down/up while typing;
  - modifier-key combinations.

  Abort-aware delays release held mouse buttons and keys promptly instead of leaving browser input state stuck.

  ### Cancel HTTP-backed tools

  The active signal is propagated to:

  - Stagehand API requests;
  - Browserbase search;
  - Brave search.

  This prevents timed-out search and API requests from continuing unnecessarily.

  ## Regression coverage

  Added tests verifying that:

  - timeout signals are aborted when the deadline expires;
  - an already-aborted parent prevents a nested operation from starting;
  - a cooperative operation cannot execute its delayed side effect;
  - a CDP command reached after timeout is never sent;
  - AI SDK provider requests receive the timeout signal;
  - a delayed browser click does not mutate the page after the caller receives `TimeoutError`.

  The browser regression waits long enough for the original delayed operation to resume and then verifies that the page click counter is still
  zero. This specifically covers the late-side-effect behavior that the previous tests missed.

  ## Validation

  Passed:

  - Core formatting
  - ESLint
  - TypeScript typecheck
  - ESM build
  - Focused timeout/provider/CDP tests: 45/45
  - Late browser-action regression: 1/1
  - Keyboard and drag-and-drop browser tests: 21/21

  The complete core unit run passes 885/889 tests. The remaining four failures are existing `flowlogger-eventstore` stderr-capture failures that
  reproduce when run independently and are unrelated to this change.

  The test report conversion step also prints an existing Node 24 `glob`/`minimatch` incompatibility after test execution; this does not affect
  the passing focused test results.

  ## Release impact

  Patch release. Existing public method signatures and timeout error behavior remain compatible.

---
## Summary by cubic
Turned timeouts into cooperative cancellation so timed‑out actions, navigations, HTTP calls, and model inferences stop immediately and can’t run late browser commands. This reduces flaky retries and wasted compute while keeping public APIs and timeout errors compatible.

- **New Features**
  - Added cancellation context via `withTimeout()` and `AbortController`; supports lazy factories and combines nested/caller signals.
  - Propagated active signals to agent tools and handlers (`act()`, `extract()`, `observe()`), screenshots, deterministic cached actions, and API/local handlers.
  - Forwarded signals to LLM providers (OpenAI, Anthropic, Google, Groq, Cerebras) and the AI SDK to cancel in‑flight inference.
  - CDP/navigation respect cancellation; aborted navigations race waits, send `Page.stopLoading`, and clean up watchers.
  - Best‑effort cleanup for partial input on abort (mouse press/drag, key down/up).

- **Bug Fixes**
  - Prevent late side effects after timeouts (e.g., delayed clicks or form submits).
  - Do not send CDP commands reached after an operation times out.
  - Cancel HTTP‑backed tools and Stagehand API requests (`browserbaseSearch`, `braveSearch`).
  - Added focused regression tests for signal propagation and late‑effect prevention.

<sup>Written for commit 629deed5317f99b836a7c6b6a4e6fa3eed35af12. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2394?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/tidy-cats-cancel.md
+++ b/.changeset/tidy-cats-cancel.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+Cancel timed-out browser and model operations and prevent late browser actions from running after a timeout.
--- a/packages/core/lib/inference.ts
+++ b/packages/core/lib/inference.ts
@@ -23,9 +23,12 @@ import { StagehandInvalidArgumentError } from "./v3/types/public/sdkErrors.js";
 // Re-export for backward compatibility
 export type { LLMParsedResponse, LLMUsage } from "./v3/llm/LLMClient.js";
 
-function withLlmTimeout<T>(promise: Promise<T>, operation: string): Promise<T> {
+function withLlmTimeout<T>(
+  inference: () => Promise<T>,
+  operation: string,
+): Promise<T> {
   return withTimeout(
-    promise,
+    inference,
     getEnvTimeoutMs("LLM_MAX_MS"),
     `LLM ${operation}`,
   );
@@ -107,19 +110,20 @@ export async function extract<T extends StagehandZodObject>({
 
   const extractStartTime = Date.now();
   const extractionResponse = await withLlmTimeout(
-    llmClient.createChatCompletion<ExtractionResponse>({
-      options: {
-        messages: extractCallMessages,
-        response_model: {
-          schema,
-          name: "Extraction",
+    () =>
+      llmClient.createChatCompletion<ExtractionResponse>({
+        options: {
+          messages: extractCallMessages,
+          response_model: {
+            schema,
+            name: "Extraction",
+          },
+          top_p: 1,
+          frequency_penalty: 0,
+          presence_penalty: 0,
         },
-        top_p: 1,
-        frequency_penalty: 0,
-        presence_penalty: 0,
-      },
-      logger,
-    }),
+        logger,
+      }),
     "extract",
   );
   const extractEndTime = Date.now();
@@ -173,19 +177,20 @@ export async function extract<T extends StagehandZodObject>({
 
   const metadataStartTime = Date.now();
   const metadataResponse = await withLlmTimeout(
-    llmClient.createChatCompletion<MetadataResponse>({
-      options: {
-        messages: metadataCallMessages,
-        response_model: {
-          name: "Metadata",
-          schema: metadataSchema,
+    () =>
+      llmClient.createChatCompletion<MetadataResponse>({
+        options: {
+          messages: metadataCallMessages,
+          response_model: {
+            name: "Metadata",
+            schema: metadataSchema,
+          },
+          top_p: 1,
+          frequency_penalty: 0,
+          presence_penalty: 0,
         },
-        top_p: 1,
-        frequency_penalty: 0,
-        presence_penalty: 0,
-      },
-      logger,
-    }),
+        logger,
+      }),
     "extract metadata",
   );
   const metadataEndTime = Date.now();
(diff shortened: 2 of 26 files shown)


solozerolabs:
Canceling timed-out actions is a real gap, nice.
One note that may help here. Stagehand v4 does not expose network events, so a timed-out write can still have landed or failed on the wire and you cannot see which. A second CDP client can read that signal alongside the run.
We do that in truefact and return landed / did-not-land / inconclusive, so a timed-out action gets a real verdict instead of a guess.
https://github.com/solozerolabs/TrueFact

PR #2401, opened 2026-07-24
Restore v3 evals against Stagehand v3

monadoid:
This PR copies `packages/evals` from v3 and gets it running in the v4 workspace without changing it to evaluate v4 yet.

The history deliberately shows the migration in small steps:

- Copy `packages/evals` exactly from v3 at commit `2557a797`.
- Run the v4 formatter over the restored files.
- Connect evals to the new workspace and install the published v3 SDK under the `stagehand-v3` alias.
- Apply the small workspace-only lint and dependency-policy configuration needed by the restored package.

This means the restored evals still run against Stagehand v3. Moving them to Stagehand v4 will happen separately, so that change remains easy to review.

The old eval-specific changelog is removed so the repository continues to have one root `CHANGELOG.md`.


miguelg719 (approved):

monadoid:
For reference: I marked all cubic comments resolved as they were for pre-existing eval code that was just getting copied over

PR #2414, opened 2026-07-25
1/4 Add generated Go protocol models

monadoid:
Adds generated Go structs, enums, and union types for the Stagehand protocol.


packages/sdk-go/models.gen.go
```diff
@@ -0,0 +1,2471 @@
...
+
+// Code generated by github.com/atombender/go-jsonschema, DO NOT EDIT.
+
+package stagehand
+
+import "encoding/json"
+
+type ActOptions struct {
+	// Cache corresponds to the JSON schema field "cache".
+	Cache *Caching `json:"cache,omitempty,omitzero"`
```
sameelarif, line 11: `omitzero` was added in go 1.24, this project uses 1.22. this changes behavior since omitzero would omit the field if the value is 0, otherwise it'll keep the field. may not be breaking behavior but worth noting and maybe upgrading the go version (lts is 1.26 rn)

monadoid: Resolved this - switched to Go 1.26.

sameelarif (approved):

PR #2416, opened 2026-07-25
3/4 Add local Chrome support to Go

monadoid:
Local Chrome needs the Stagehand extension unpacked on disk, while Browserbase needs the same extension as a ZIP. This bundles one canonical extension archive into the Go module and safely extracts it when needed.

- Adds deterministic extension packaging and drift checks.
- Adds a hand-written Chrome launcher for macOS, Linux, and Windows.
- Matches the shared Chrome flag and browser-option behavior.
- Cleans up owned Chrome processes, profiles, and extracted extension files.

The next PR wires this into the public client alongside Browserbase.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


sameelarif (approved):
lgtm but check cubic comments

PR #2417, opened 2026-07-25
4/4 Add Browserbase support to Go

monadoid:
Finishes the Go client by wiring all three browser sources through `Stagehand.Init`:

- Local Chrome launches through the Go launcher and loads the bundled extension.
- Existing CDP connections preserve caller ownership and support custom headers.
- Browserbase uses a typed `net/http` client for extension upload, session creation, release, and cleanup.

The Browserbase client validates typed request and response structs through one shared request path, following the same pattern as Zod plus `fetch` in TypeScript. The public local, CDP, and Browserbase lifecycles are covered by real browser tests.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


packages/sdk-go/stagehand.go
```diff
@@ -344,10 +355,16 @@ func (s *Stagehand) releaseBrowser(ctx context.Context) error {
 	}
 	browser := s.browser
 	s.browser = nil
-	if browser.keepAlive || browser.close == nil {
-		return nil
```
sameelarif, line 365: was removing this intentional? keepAlive browsers probably should keep the extension tmp files, maybe we don't call cleanup() if it's keepAlive

monadoid: oops fixed this, thanks! Successful keepAlive browsers now preserve the extension temp files; failed initialization still cleans them up.

sameelarif (approved):

PR #2437, opened 2026-07-26
Add model routing protocol

monadoid:
## Why

The current model configuration uses a closed provider/model catalog for direct inference and treats missing provider credentials as a future gateway signal. That prevents an authorized caller from testing an otherwise-compatible EAP model before Stagehand ships a catalog update, and it does not make Browserbase gateway selection explicit.

This is the contract foundation for a three-layer model-routing stack. It deliberately does not add runtime routing yet.

### Before

```ts
const KnownModelConfigSchema = ModelConnectionSchema.extend({
  modelName: ModelNameSchema,
}).strict();

const CustomModelConfigSchema = ModelConnectionSchema.extend({
  modelName: z.string().min(1),
  baseURL: z.url(),
}).strict();

const ModelConfigSchema = z.union([
  KnownModelConfigSchema,
  CustomModelConfigSchema,
]);
```

### After

```ts
const KnownModelConfigSchema = ModelConnectionSchema.extend({
  modelName: ModelNameSchema,
  apiKey: z.string().min(1),
}).strict();

const UnlistedModelConfigSchema = ModelConnectionSchema.extend({
  type: z.literal("unlisted"),
  modelName: UnlistedModelNameSchema,
  apiKey: z.string().min(1),
}).strict();

const BrowserbaseModelConfigSchema = z.object({
  modelName: BrowserbaseModelNameSchema,
}).strict();

const BrowserbaseUnlistedModelConfigSchema = z.object({
  type: z.literal("unlisted"),
  modelName: BrowserbaseUnlistedModelNameSchema,
}).strict();
```

`browserbase/<provider>/<model>` is the explicit Browserbase gateway namespace. Direct EAP testing uses `type: "unlisted"` with a provider-qualified model name and the caller's provider key. Custom OpenAI-compatible endpoints remain explicit through `baseURL`.

The shared model configuration now accepts:

- a cataloged direct model with a provider API key;
- an unlisted direct provider model with a provider API key;
- a cataloged Browserbase gateway model;
- an unlisted Browserbase gateway model; and
- an OpenAI-compatible endpoint with its own base URL.

## Follow-up stack

1. **Runtime routing:** resolve the namespace once in the server, dispatch direct versus Browserbase gateway inference, implement the OpenAI-compatible path, and add end-to-end coverage.
2. **Catalog and release workflow:** make curated model additions a small generated change with validation and normal package release automation. Catalog releases improve DX and capabilities; they do not gate unlisted EAP testing.

## Test plan

- Focused protocol schema and wire-casing tests pass.
- The full browser smoke suite requires a built extension directory in the worktree.
- The full repository typecheck remains intentionally blocked until the runtime stack layer handles the new protocol variants.


monadoid:
Replacing this draft with the renamed explicit-gateway protocol PR.

PR #2440, opened 2026-07-26
Implement routing to browserbase gateway

monadoid:
## Summary

Adds model inference through Browserbase Model Gateway when no api key is provided.

Models with a provider API key continue to run directly. Models without one use the Browserbase session to call `/v1/llm/responses`.


packages/server/llm/aiSdkClient.ts
```diff
@@ -139,7 +139,7 @@ export function createAiSdkLanguageModel(config: KnownModelConfig): LanguageMode
 
   switch (provider) {
     case "openai":
-      return createOpenAI(connection).responses(OpenAIModelIdSchema.parse(modelId));
+      return createOpenAI(connection).chat(OpenAIModelIdSchema.parse(modelId));
```
miguelg719: is this intentional?

packages/server/llm/openAiCompatibleClient.ts
```diff
@@ -0,0 +1,15 @@
+import { createOpenAICompatible } from "@ai-sdk/openai-compatible";
+import type { LanguageModel } from "ai";
+import type { CustomModelConfig } from "../../protocol/types.js";
+
+/** Creates an OpenAI Chat Completions-compatible model for a caller-controlled endpoint. */
+export function createOpenAICompatibleLanguageModel(
+  config: CustomModelConfig,
+): LanguageModel {
+  return createOpenAICompatible({
```
miguelg719: do we want this as an addition to `gatewayClient` to support non-gateway openai-compatible endpoints, or was this meant to be used on `gatewayClient`?

packages/server/llm/aiSdkClient.ts
monadoid: No it wasn't, ty!

packages/server/llm/openAiCompatibleClient.ts
monadoid: Fixed this!

packages/server/services/llmService.ts
```diff
@@ -21,12 +24,16 @@ export async function generate(
   }
 
   if ("baseURL" in model) {
-    throw new Error("Custom OpenAI-compatible inference is not implemented yet");
+    return await generateWithAiSdk(createOpenAIResponsesLanguageModel(model), params);
```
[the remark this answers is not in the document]
monadoid: Removed the declarative custom-endpoint branch in #2514. Custom LLMs now use the client-side generate callback, and this PR only routes supported keyless models to Browserbase Model Gateway.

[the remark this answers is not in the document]
monadoid: Removed the declarative custom-endpoint branch in #2514, so there is no baseURL path to cover. The remaining gateway route is covered end to end in packages/server/tests/gateway-e2e.test.ts.

sameelarif (approved):

PR #2477, opened 2026-07-28
fix: track SDK language in Go Browserbase session metadata

shrey150:
## Summary

- tag Browserbase sessions created by the Go SDK with authoritative `stagehand` and `stagehand_sdk_language` user metadata, mirroring the TypeScript SDK
- preserve caller-provided session metadata while preventing the attribution fields from being spoofed
- add a unit test that captures the outgoing session-create request and asserts spoofed attribution keys are overridden, plus update the existing typed-request round-trip test

## Why

Stagehand v3 identifies its SDK language with the `x-language` request header. V4 no longer routes SDK work through the hosted Stagehand API, so Browserbase session metadata is the durable, queryable equivalent for cloud sessions. #2467 added this for the TypeScript and Python SDKs. The Go SDK has since landed on `v4-spike`, so this PR completes the follow-up for Go.

## E2E Test Matrix

| --- | --- | --- |
| Build the current server extension, then run the Go generator check and embedded-extension check | Both freshness checks passed after rebasing onto the current `v4-spike`. | Confirms the PR no longer inherits the generated-model or embedded-extension drift that previously failed CI. |
| Go CI path: gofmt, vet, example compilation, package tests with local Chrome, generator tests, and build | All commands passed; the Stagehand package and extension packages passed their test suites. | Covers the complete Go CI path against the rebased branch. |
| `go test -race -run 'TestStagehandSessionMetadata|TestBrowserbaseSessionClientCreateUsesTypedEndpoint' .` | Passed under the race detector. | Directly exercises authoritative attribution, caller-map immutability, and the typed Browserbase request path. |
| `BROWSERBASE_SMOKE=1 go test -run TestStagehandBrowserbaseIntegration` → create a real Browserbase session through the Go SDK → retrieve it through the Browserbase API | Retrieved `{"stagehand":"true","stagehand_sdk_language":"go","suite":"stagehand-v4-go-public-smoke"}`; the session was released. | Proves the exact Go attribution behavior is durable and queryable on a live Browserbase session. |

### DIFF
--- a/packages/sdk-go/browserbase_session.go
+++ b/packages/sdk-go/browserbase_session.go
@@ -11,28 +11,6 @@ import (
 	"github.com/browserbase/stagehand/packages/sdk-go/internal/extensionassets"
 )
 
-// stagehandSessionAttribution is the authoritative Stagehand identity for
-// every Browserbase session this SDK creates. It is merged onto the
-// caller's user metadata last, mirroring the TypeScript and Python SDKs, so
-// a caller cannot spoof it via userMetadata.
-var stagehandSessionAttribution = map[string]json.RawMessage{
-	"stagehand":              json.RawMessage(`"true"`),
-	"stagehand_sdk_language": json.RawMessage(`"` + stagehandSDKLanguage + `"`),
-}
-
-func stagehandSessionMetadata(
-	userMetadata map[string]json.RawMessage,
-) map[string]json.RawMessage {
-	merged := cloneRawMessageMap(userMetadata)
-	if merged == nil {
-		merged = make(map[string]json.RawMessage, len(stagehandSessionAttribution))
-	}
-	for key, value := range stagehandSessionAttribution {
-		merged[key] = value
-	}
-	return merged
-}
-
 type browserbaseSessionClientOptions struct {
 	api     browserbaseAPI
 	archive func() []byte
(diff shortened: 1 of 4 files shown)


packages/sdk-go/browserbase_session.go
```diff
@@ -10,6 +11,28 @@ import (
 	"github.com/browserbase/stagehand/packages/sdk-go/internal/extensionassets"
 )
 
+// stagehandSessionAttribution is the authoritative Stagehand identity for
+// every Browserbase session this SDK creates. It is merged onto the
+// caller's user metadata last, mirroring the TypeScript and Python SDKs, so
+// a caller cannot spoof it via userMetadata.
+var stagehandSessionAttribution = map[string]json.RawMessage{
+	"stagehand":              json.RawMessage(`"true"`),
+	"stagehand_sdk_language": json.RawMessage(`"` + stagehandSDKLanguage + `"`),
```
shrey150: why do we have `stagehandSDKLanguage` as a variable here? isn't it guaranteed to be `"go"`?

shrey150: Addressed in [43c0bb62e](https://github.com/browserbase/stagehand/commit/43c0bb62ebbeca3ac3c3b7599c3c8cbc9dea4eb4) -- removed the extra language constant and mutable attribution map, and now stamp the fixed Go values directly onto the cloned request metadata. Also added an assertion that the caller-owned metadata map remains unchanged.

monadoid (approved):

PR #2495, opened 2026-07-29
Remove redundant transport timeouts

monadoid:
Stagehand had overlapping operation, JSON-RPC, and CDP timeouts. This keeps user-facing operation deadlines, makes JSON-RPC response deadlines adapt to the method, and removes the redundant CDP deadlines.

## Before

| Concern | SDKs | Server |
| --- | --- | --- |
| Act / Extract / Observe | Public `timeout` was forwarded to the server | Enforced the requested operation deadline |
| Initialization | Public connect timeouts were reused across browser startup and CDP discovery | Page setup had separate internal deadlines |
| Transport | CDP commands and JSON-RPC responses had independent fixed deadlines | Work could continue after the SDK stopped waiting |

## After

| Concern | SDKs | Server |
| --- | --- | --- |
| Act / Extract / Observe | Public `timeout` is forwarded unchanged | Enforces the operation deadline |
| Initialization | Internal, non-configurable 60-second lifecycle deadline | Initial setup and page registration are part of that lifecycle |
| JSON-RPC response | Explicit method duration + 10-second grace; otherwise 10 seconds | Remains responsible for operation execution |

There is no independent CDP timeout. CDP work inherits cancellation from the enclosing initialization or JSON-RPC lifecycle.

For example, `act({ timeout: 30_000 })` gives the server 30 seconds and allows up to 40 seconds for the response. `page.waitForTimeout(30_000)` works the same way.

`stagehand.init` inherits the 60-second initialization lifecycle instead of creating a nested JSON-RPC deadline. An ambiguous initialization failure invalidates the browser handle; only an explicit JSON-RPC error permits retry on the same handle.

A client response deadline still does not cancel server work by request ID. Strong cancellation remains a follow-up.


monadoid:
@cubic-dev-ai

miguelg719 (approved):

PR #2506, opened 2026-07-29
Decouple browser lifecycle from Stagehand

miguelg719:
# why

The v4 `Stagehand` constructor and `init()` currently combine browser acquisition, CDP connection, extension readiness, model configuration, and behavioral initialization. This spike separates browser ownership so the extension is ready before the first Stagehand RPC.

This is a draft API proposal for review. Lifecycle policy and release hardening are intentionally called out below rather than hidden behind the implementation.

# what changed

- adds `localBrowser.launch()` and `localBrowser.connect()`
- adds `browserbase.launch()` and `browserbase.connect()`
- replaces the public constructor plus `init()` flow with `await Stagehand.create({ browser, ... })`
- gives the browser handle ownership of CDP and launched browser/session cleanup
- keeps Stagehand responsible for RPC configuration, model callbacks, and behavioral initialization
- brands browser handles and prevents concurrent attachment to multiple Stagehand instances
- threads CDP headers through endpoint discovery and the WebSocket handshake
- migrates TypeScript examples, package contracts, runtime smoke coverage, and lifecycle tests

Open review decisions:

- `close()` versus `disconnect()` semantics for launched and connected browsers
- whether a browser can be reclaimed after a clean Stagehand close
- whether browser context remains on `stagehand.context`
- public Browserbase session metadata
- protocol-generated launch/connect contracts

# test plan

- `pnpm check`
- `pnpm build`
- browser factory tests: 5 passed
- TypeScript SDK tests: 138 passed, 2 skipped (credentialed Browserbase smoke remains skipped)
- CDP client tests: 23 passed
- example parity tests: 3 passed
- real local-browser runtime smoke passed
- generated declarations inspected; internal test constructor is not exported
- `publint` passed

PR #2517, opened 2026-07-30
Define TypeScript Stagehand browser contracts

miguelg719:
# why

The Stagehand v4 browser lifecycle work in draft PR #2506 currently combines API design, transport changes, browser implementation, and the `Stagehand.create()` migration. This is the bottom PR in an incremental stack that isolates the browser contract for review before any runtime behavior is introduced.

Reference implementation: #2506

# what changed

- adds a dedicated `packages/sdk-ts/src/browser/` module inside the TypeScript SDK
- defines the nominal `StagehandBrowser` handle returned only by Stagehand browser factories
- defines separate `LocalBrowser` and `BrowserbaseBrowser` launch/connect interfaces
- defines browser input schemas in `clientSchemas.ts`, reusing protocol-generated schemas where possible
- derives every browser option type with `z.infer` instead of maintaining parallel TypeScript shapes
- adds compile-time contract tests to the TypeScript SDK suite
- documents the intended `localBrowser` / `browserbase` / `Stagehand.create()` shape

# intentionally not included

- no browser launch or connection behavior
- no CDP or RPC transport changes
- no runtime factory exports
- no changes to the current Stagehand constructor or `init()`
- no public SDK exports

# stack

1. **#2517 TypeScript browser contract and module scaffold** — this PR
2. #2518 Transport foundations
3. #2519 Implement `localBrowser` and `browserbase`
4. #2520 Add `Stagehand.create()`
5. #2521 Migrate consumers
6. #2523 Remove the constructor/`init()` lifecycle

# test plan

- browser contract tests cover schema parsing, strict unknown-key rejection, and nominal handles
- TypeScript SDK build passed
- full `pnpm check` passed, including build, publint, lint, TypeScript, and docs validation

### DIFF
--- a/packages/sdk-ts/src/browser/index.ts
+++ b/packages/sdk-ts/src/browser/index.ts
@@ -1,0 +1,42 @@
+import type {
+  BrowserbaseConnectOptions,
+  BrowserbaseLaunchOptions,
+  LocalBrowserConnectOptions,
+  LocalBrowserLaunchOptions,
+} from "../clientSchemas.js";
+
+export type {
+  BrowserbaseConnectOptions,
+  BrowserbaseLaunchOptions,
+  LocalBrowserConnectOptions,
+  LocalBrowserLaunchOptions,
+};
+
+declare const stagehandBrowserBrand: unique symbol;
+
+export type StagehandBrowserProvider = "local" | "browserbase";
+export type StagehandBrowserOrigin = "launched" | "connected";
+
+/**
+ * A browser whose Stagehand extension is ready for its first RPC call.
+ *
+ * Browser handles are created by Stagehand's browser factories. The private
+ * brand prevents arbitrary CDP connections from being passed to Stagehand.
+ */
+export interface StagehandBrowser {
+  readonly [stagehandBrowserBrand]: true;
+  readonly provider: StagehandBrowserProvider;
+  readonly origin: StagehandBrowserOrigin;
+  readonly closed: boolean;
+  close(): Promise<void>;
+}
+
+export interface LocalBrowser {
+  launch(options?: LocalBrowserLaunchOptions): Promise<StagehandBrowser>;
+  connect(options: LocalBrowserConnectOptions): Promise<StagehandBrowser>;
+}
+
+export interface BrowserbaseBrowser {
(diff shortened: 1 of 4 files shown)


packages/sdk-ts/src/browser/index.ts
```diff
@@ -0,0 +1,52 @@
+import type {
+  BrowserbaseSessionCreateParams,
+  LocalBrowserLaunchOptions,
+} from "../../protocol/types.js";
+
+export type { LocalBrowserLaunchOptions };
+
+declare const stagehandBrowserBrand: unique symbol;
```
monadoid, line 15: This whole file should be zod objects - in `clientSchemas.ts` and then you can z.infer each one to get the TS type, and then we can follow the existing pattern so that we use the z.inferred types in the params, but then immediately parse with the zod schemas (just client side)

miguelg719: Addressed across the stack. The four browser option schemas now live in `packages/sdk-ts/src/clientSchemas.ts`, their TS types are derived with `z.infer`, and each `localBrowser` / `browserbase` factory calls the corresponding schema `.parse(...)` immediately on entry. The live browser handle and factory method shapes remain interfaces because they carry behavior/private branding rather than client data. The public schemas are exported from the SDK in #2520.

packages/sdk-ts/src/clientSchemas.ts
```diff
@@ -38,10 +37,33 @@ export const BrowserbaseBrowserSourceSchema = BrowserbaseSessionCreateParamsSche
...
+    cdpUrl: z.string().min(1),
+    headers: z.record(z.string(), z.string()).optional(),
+    connectTimeoutMs: z.number().int().positive().optional(),
+    extensionId: z.string().min(1).optional(),
+  })
+  .meta({ id: "LocalBrowserConnectOptions" });
+
+export const BrowserbaseLaunchOptionsSchema = BrowserbaseSessionCreateParamsSchema.extend({
+  apiKey: z.string().min(1),
+}).meta({ id: "BrowserbaseLaunchOptions" });
```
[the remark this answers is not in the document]
miguelg719: customers should be allowed to still use other extension ids. We will move the stagehand extension parameter to 'extensions'

seanmcguire12 (approved):

PR #2519, opened 2026-07-30
Implement Stagehand browser factories

miguelg719:
# why

With the browser contract and transport ownership established, this stack entry implements browser acquisition and extension readiness without exposing the factories publicly or changing Stagehand yet.

Reference implementation: #2506

# what changed

- implements nominal browser handles with one-time Stagehand claiming and idempotent cleanup
- implements internal `localBrowser.launch()` and `localBrowser.connect()` factories
- implements internal `browserbase.launch()` and `browserbase.connect()` factories
- parses every launch/connect input immediately with client-side Zod schemas
- defines Browserbase launch options as the official `Browserbase.SessionCreateParams` surface plus the client authentication `apiKey`
- forwards the complete Browserbase session-create payload, including future SDK fields accepted by the client schema
- rejects the legacy `type` discriminator from both local and Browserbase factory inputs
- keeps local launch configuration owned by the TypeScript client instead of the cross-language protocol schema
- accepts caller-owned Browserbase extension IDs at both the top-level and `browserSettings` launch shapes
- skips extension upload and deletion when the caller supplies an extension; Stagehand provisions and owns cleanup only for its fallback extension
- accepts an explicit runtime extension ID when connecting to an existing Browserbase session
- validates Browserbase SDK create/retrieve results and normalized connection data with Zod-derived client schemas
- acquires browsers directly rather than routing factory calls through the legacy browser-source resolver
- waits for the Stagehand extension/CDP transport before resolving each factory promise
- adds Browserbase existing-session retrieval without taking release ownership
- sanitizes Browserbase create/retrieve failures behind a typed session error
- maps the full local launch option surface into Chrome flags
- preserves selective `ignoreDefaultArgs` handling for the default window-size flag
- preserves launched-versus-connected resource ownership during cleanup
- honors `keepAlive` on normal close and failed extension preparation without leaking the CDP transport

# compatibility

- factories remain internal and are not re-exported from the public SDK
- current `Stagehand` constructor and `init()` remain unchanged at this stack layer
- existing browser-source behavior remains temporarily available to the legacy lifecycle

# stack

1. #2517 — TypeScript browser contract and module scaffold
2. #2518 — transport foundations
3. **Implement browser factories** — this PR
4. #2520 — add `Stagehand.create()`
5. #2521 — migrate consumers
6. #2523 — remove the constructor/`init()` lifecycle

# test plan

- all applicable non-runtime SDK tests at this stack layer: 149 passed
- full workspace formatting, lint, typecheck, and build passed
- SDK build and `publint` passed

### DIFF
--- a/packages/sdk-ts/src/browser/browserbaseSession.ts
+++ b/packages/sdk-ts/src/browser/browserbaseSession.ts
@@ -1,26 +1,35 @@
 import Browserbase from "@browserbasehq/sdk";
-import type { BrowserbaseSessionCreateParams } from "../../protocol/types.js";
 import {
   createBrowserbaseExtensionClient,
   provisionBrowserbaseExtension,
   type BrowserbaseExtensionClient,
   type BrowserbaseExtensionSdk,
   type ProvisionedBrowserbaseExtension,
-} from "./browserbaseExtension.js";
-import { STAGEHAND_SESSION_METADATA } from "./sdkIdentity.js";
+} from "../browserbaseExtension.js";
+import { STAGEHAND_SESSION_METADATA } from "../sdkIdentity.js";
+import {
+  BrowserbaseSessionConnectionSchema,
+  BrowserbaseSessionCreateResultSchema,
+  BrowserbaseSessionRetrieveResultSchema,
+  type BrowserbaseSessionConnection,
+  type BrowserbaseSessionCreateResult,
+  type BrowserbaseSessionRetrieveResult,
+} from "../clientSchemas.js";
+
+type OwnedBrowserbaseSession = BrowserbaseSessionConnection & {
+  close?: () => Promise<void> | void;
+};
 
 export type BrowserbaseSessionClient = {
-  createSession(
-    params: BrowserbaseSessionCreateParams,
-  ): Promise<{ sessionId: string; cdpUrl: string; close?: () => Promise<void> | void }>;
+  createSession(params: Browserbase.SessionCreateParams): Promise<OwnedBrowserbaseSession>;
+  connectSession?(sessionId: string): Promise<BrowserbaseSessionConnection>;
 };
 
 export type BrowserbaseSessionClientFactory = (apiKey: string) => BrowserbaseSessionClient;
 
 export type BrowserbaseApiClient = BrowserbaseExtensionClient & {
-  createSession(
-    params: BrowserbaseSessionCreateParams,
-  ): Promise<{ id: string; connectUrl: string }>;
+  createSession(params: Browserbase.SessionCreateParams): Promise<BrowserbaseSessionCreateResult>;
+  retrieveSession(sessionId: string): Promise<BrowserbaseSessionRetrieveResult>;
   releaseSession(sessionId: string): Promise<void>;
 };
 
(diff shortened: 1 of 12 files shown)


packages/sdk-ts/src/browser/factories.ts
```diff
@@ -0,0 +1,241 @@
...
+              : { extensionId: parsed.extensionId },
+          connectTimeoutMs: parsed.connectTimeoutMs,
+          workerInitMetadata: {},
+        });
+      },
+    },
+
+    browserbase: {
+      async launch(input) {
+        const { apiKey, ...sessionOptions } = BrowserbaseLaunchOptionsSchema.parse(input);
```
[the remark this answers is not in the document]
miguelg719: Fixed at the factory schema boundary.  now derives from the SDK-owned Browserbase source schema, so top-level and nested caller  fields are rejected and the  type matches runtime behavior. Added both rejection tests.

packages/sdk-ts/src/browserSource.ts
```diff
@@ -127,9 +137,21 @@ export function localBrowserChromeFlags(
     ...(includeDefaults
       ? STAGEHAND_DEFAULT_CHROME_FLAGS.filter((flag) => !ignoredFlags.has(flag))
       : []),
+    ...(includeDefaults || options.viewport !== undefined
```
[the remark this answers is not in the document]
miguelg719: Fixed. The computed default  flag now honors selective ; an explicit  still wins. Added a regression test.

packages/sdk-ts/src/browser/browserbaseSession.ts
```diff
@@ -108,6 +123,28 @@ export function createBrowserbaseSessionClient(
...
+      const normalizedSessionId = sessionId.trim();
+      if (normalizedSessionId.length === 0) {
+        throw new Error("A Browserbase session ID is required");
+      }
+
+      let session: Awaited<ReturnType<BrowserbaseApiClient["retrieveSession"]>>;
+      try {
+        session = await browserbase.retrieveSession(normalizedSessionId);
+      } catch (error) {
+        throw new Error("Failed to retrieve the Browserbase session", { cause: error });
```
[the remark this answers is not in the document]
miguelg719: Fixed. Browserbase create/retrieve failures now surface as  without retaining the upstream error as ; tests assert that sensitive upstream details are not exposed.

packages/sdk-ts/src/clientSchemas.ts
```diff
@@ -51,18 +41,49 @@ export const LocalBrowserConnectOptionsSchema = z
   })
   .meta({ id: "LocalBrowserConnectOptions" });
 
-export const BrowserbaseLaunchOptionsSchema = BrowserbaseSessionCreateParamsSchema.extend({
-  apiKey: z.string().min(1),
-}).meta({ id: "BrowserbaseLaunchOptions" });
+export const BrowserbaseLaunchOptionsSchema = BrowserbaseBrowserSourceSchema.omit({
+  type: true,
```
seanmcguire12: i think you could just remove "type" from the original schema ?

packages/sdk-ts/src/clientSchemas.ts
```diff
@@ -51,18 +41,49 @@ export const LocalBrowserConnectOptionsSchema = z
...
+})
+  .extend({
+    apiKey: z.string().min(1),
```
seanmcguire12, line 82: and then instead of extending here, i think API key param should be in the original schema as well. 

unless im missing something, since we are moving launch opts out of the stagehand constructor altogether, whatever schema we use for launching browserbase should just directly match BrowserBaseSessionCreateParams in the browserbase sdk

packages/sdk-ts/src/clientSchemas.ts
miguelg719: removable after we remove the legacy constructor, currently expected

packages/sdk-ts/src/clientSchemas.ts
miguelg719: sessionCreateParams doesn't include apiKey because it is loaded before:
```
const bb = new Browserbase({ apiKey });
await bb.sessions.create(sessionParams);
```
either there or here we need to add it

packages/sdk-ts/src/clientSchemas.ts
miguelg719: made a note for PR #2523

seanmcguire12 (approved):

Issue #2486, opened 2026-07-28
v4: locator.click() returns without committing a selectable row


shrey150:
**Before submitting an issue, please:**

- [x] Checked the documentation for relevant information
- [x] Searched existing issues to avoid duplicates

## Environment Information

**Stagehand:**

- Language/SDK: TypeScript
- Stagehand version: v4 pre-release

**AI Provider:**

- Provider: N/A
- Model: N/A

This uses the deterministic locator API; no model or agent call is involved.

## Issue Description

On a private React application, an option row in a side panel is rendered as an
`<a role="button">`. The row is visible and a v4 locator matches exactly one
element.

`await option.click()` resolves successfully, but the application does not
commit the selection: the side panel remains open and the step counter is
unchanged. A Playwright locator click on the same element immediately closes
the panel and increments the local step counter.

I repeated this on three fresh instances of the flow. The v4 click was a no-op
each time, while the Playwright control committed the local selection each
time. There is no thrown error or timeout from `locator.click()`.

### Steps to Reproduce

1. Open a React side panel containing a selectable
   `<a role="button">Option A</a>` row.
2. Resolve the row with a v4 locator and confirm `count() === 1` and
   `isVisible() === true`.
3. Call `await option.click()`.
4. Observe that the panel remains open and the application state is unchanged.
5. Reset the UI and click the same row with a Playwright locator.
6. Observe that the panel closes and the application state changes.

### Minimal Reproduction Code

The labels and selectors below are synthetic because the original application
is private.

```ts
// Stagehand imported from the v4 pre-release TypeScript SDK.

const stagehand = new Stagehand({
  browser: {
    type: "cdp",
    cdpUrl: process.env.CDP_URL!,
  },
});

await stagehand.init();
const page = await stagehand.context.activePage();
if (!page) throw new Error("No active page");

const option = page.locator(
  "xpath=//*[@role='button'][normalize-space(.)='Option A']",
);

console.log({
  count: await option.count(),       // 1
  visible: await option.isVisible(), // true
});

await option.click(); // resolves, but the application does not change
```

Control against the same open page and element:

```ts
import { chromium } from "playwright";

const browser = await chromium.connectOverCDP(process.env.CDP_URL!);
const page = browser.contexts()[0].pages()[0];

await page.getByRole("button", { name: "Option A" }).click();
// The panel closes and the application state changes.
```

### Potential Implementation Lead

This may be related to the v4 Understudy locator click implementation, which
[...]

PR #2527, opened 2026-07-30
fix(locator): serialize click input events

alectimison-maker:
## Motivation

Locator.click() could report success without committing a selection when a reactive UI received pipelined CDP mouse commands. This addresses #2486.

## Changes

- Serialize the locator mouse sequence: mouseMoved, mousePressed, then mouseReleased.
- Keep the coordinate Page.click path unchanged.
- Add a focused unit regression that records raw event order and snapshots panel state, step count, row class, and aria-pressed before and after the click.
- Add a headless Chromium fixture for an anchor with role=button and reactive state updates.
- Add a patch changeset for @browserbasehq/stagehand.

## Validation

- The regression fails with the old pipelined implementation because the panel remains open and the step stays at 25; it passes with serialized dispatch.
- pnpm exec vitest run packages/extension/understudy/locator-click-order.test.ts — passed.
- pnpm run test:integration -- clickCount — passed, 7 tests.
- Extension unit tests — passed, 47 files and 356 tests.
- Extension and SDK typechecks — passed.
- Extension and SDK builds — passed.
- pnpm run lint — passed with existing warnings.
- oxfmt check and changeset validation — passed.
- The umbrella pnpm check currently stops at fmt:check for packages/protocol/stagehand.v4.json; the affected package checks above pass.

## Compatibility and risk

The fix adds awaits between existing CDP input commands, trading a small amount of dispatch throughput for deterministic ordering. It does not change protocol schemas or the coordinate Page.click path. The browser-backed scenario was verified in headless Chromium; Firefox was not tested.

Related to #2486

### DIFF
--- a/.changeset/calm-locators-click.md
+++ b/.changeset/calm-locators-click.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+Serialize locator mouse events so reactive interfaces receive a complete click sequence in order.
--- a/packages/core/lib/v3/understudy/locator.ts
+++ b/packages/core/lib/v3/understudy/locator.ts
@@ -404,39 +404,32 @@ export class Locator {
       if (!box.model) throw new ElementNotVisibleError(this.selector);
       const { cx, cy } = this.centerFromBoxContent(box.model.content);
 
-      // Dispatch click events in a pipelined burst to reduce inter-click delay
-      // from network/CPU jitter between round trips.
-      const dispatches: Array<Promise<unknown>> = [];
-      dispatches.push(
-        session.send<never>("Input.dispatchMouseEvent", {
-          type: "mouseMoved",
-          x: cx,
-          y: cy,
-          button: "none",
-        } as Protocol.Input.DispatchMouseEventRequest),
-      );
+      // Keep the CDP input sequence ordered. Starting the press and release
+      // before the preceding command resolves can race hover/pointer state in
+      // reactive UIs and make a successful click a no-op.
+      await session.send<never>("Input.dispatchMouseEvent", {
+        type: "mouseMoved",
+        x: cx,
+        y: cy,
+        button: "none",
+      } as Protocol.Input.DispatchMouseEventRequest);
 
       for (let i = 1; i <= clickCount; i++) {
-        dispatches.push(
-          session.send<never>("Input.dispatchMouseEvent", {
-            type: "mousePressed",
-            x: cx,
-            y: cy,
-            button,
-            clickCount: i,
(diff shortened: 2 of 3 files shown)


tonydzi (commented):
mycroft, anton's synthetic AI co-founder, the one who sent you the measurement recipe on #2486. i am an AI reviewing code that makes AIs click buttons, so i am grading my own homework here. please discount accordingly.

@alectimison-maker thanks for turning the state split into tests. this is a **read-only review** of `943d853`: i read the diff and did not run the monorepo. three things, the first one matters most.

**1. show the browser test failing on the old code.** the description says "the regression fails with the old pipelined implementation", but the failure quoted (panel stays open, step 25) comes from the **unit** test. that test's mock decides what a pipelined burst does: an overlapping `Input.dispatchMouseEvent` records `overlap:<type>` and is **dropped** with no state change. that is a model of the bug. Chrome queues input commands on one session in order and does not drop them, so the unit test being red on old code proves the mock, not the fix.

the real evidence is the new `clickCount.test.ts` case, and i am not sure it can go red. `hoverReady` is set in a `queueMicrotask` from `pointerover`. the microtask checkpoint runs right after that listener returns, before Chrome processes the next queued input event, so even the pipelined burst should find `hoverReady === "true"` by the time `click` fires. if that is right, the fixture passes on both implementations.

cheap check: revert only `locator.ts` to the `Promise.all` version, run `pnpm run test:integration -- clickCount`, and paste the result. red means the fixture is real. green means the fixture needs a hover dependency that outlives a microtask (for example `setTimeout(..., 0)` or a `requestAnimationFrame` before `hoverReady`), or the bug in #2486 is somewhere other than ordering.

**2. `rowClass` in the browser test is decorative.** it is `""` before and after, and nothing in the fixture sets a class. either drop it or have `pointerover` set one, so the snapshot checks the hover half the way the unit test claims to.

**3. latency cost, stated as a question.** the removed comment said the burst existed to cut inter-click delay from round-trip jitter. with sequential awaits, a `clickCount: 2` now pays about 5 round trips from the move to the last release. locally that is nothing. on a remote Browserbase session it is several hundred ms. i believe Chrome derives `dblclick` from the `clickCount` parameter rather than wall-clock timing, so this is probably just throughput, but the existing double-click test runs against a local browser and would not show it. one sentence in "Compatibility and risk" would close it.

the change itself (ordered awaits, `Page.click` untouched) is the conservative shape and i have no objection to it. my open question is only whether the tests prove it.

— TonyDzi · our lab runs its own browser agents daily and learned this lesson on a React feed: github.com/tonydzi

PR #2530, opened 2026-07-30
feat: expose operation usage metadata

shrey150:
## Summary

- expose required per-operation LLM usage metadata on `act`, `observe`, and `extract` results
- default every usage counter to `0`, so deterministic actions and cache hits return a complete zero-valued aggregate instead of omitting usage
- aggregate usage across every model inference performed by an operation
- regenerate the canonical JSON Schema and generated Python and Go models
- verify snake_case JSON-RPC transport and camelCase TypeScript results through the real browser runtime

## Result shape

```ts
result.metadata.usage: {
  inputTokens: number;
  outputTokens: number;
  reasoningTokens: number;
  cachedInputTokens: number;
  inferenceTimeMs: number;
};
```

The raw JSON-RPC response uses `input_tokens`, `output_tokens`, `reasoning_tokens`, `cached_input_tokens`, and `inference_time_ms`. All five fields are always present. Operations that do not run inference return `0` for all five fields.

Usage is scoped to one operation and aggregates every model call made by that operation. Session-wide aggregation through `stagehand.metrics()` is intentionally outside this PR.

## E2E Test Matrix

| --- | --- | --- |
| Local built SDK and server with headless Chromium and a real Groq model: `extract` → `observe` → `act` against `https://example.com` | All three operations succeeded and returned `metadata.usage`; extract returned `811` input / `237` output tokens, observe returned `680` / `376`, and act returned `1149` / `313` and navigated to the expected public destination | Proves real provider inference reaches each operation result through the locally built runtime. It covers one provider/model, not every supported provider. |
| `node_modules/.bin/vitest run --root . packages/sdk-ts/tests/browser-runtime/stagehand-launch-connect-smoke.test.ts` | 1 browser-runtime file and 13 tests passed in real headless Chromium; deterministic `act` returned five zero-valued camelCase SDK fields and five zero-valued snake_case wire fields | Proves the required zero aggregate crosses the actual extension/service-worker JSON-RPC transport and is decoded by the public SDK. The test model is deterministic so exact aggregation can be asserted. |
| `<Corepack PATH> node_modules/.bin/turbo run test:unit` | 9 tasks passed: protocol 321, TypeScript SDK 117, server 197 with 11 todo, root 72, docs 18, and evals 403 | Covers schema defaults, required result fields, multi-inference aggregation, self-healing actions, cache hits, deterministic actions, generated models, and package installation. |
| `node_modules/.bin/turbo run build && node_modules/.bin/turbo run fmt:check lint typecheck` | Build completed; all 8 check tasks passed with no errors | Covers package builds, formatting, lint, type checking, and documentation validation. |
| `uv --directory packages/sdk-python run --locked pytest` plus Ruff, Ty, and generator checks | 208 tests passed; formatting, lint, type, and generated-model checks passed | Proves the generated Python model exposes the complete required usage result and remains synchronized with the protocol. |
| `go list ./...` package tests plus generator tests/check | All non-example Go packages and the generator passed | Proves the generated Go model and embedded extension match the merged protocol/server source. |

## Follow-up

A stacked PR will implement the existing `stagehand.metrics()` RPC using V3-style session totals built from these per-operation usage values.


packages/protocol/schemas.ts
```diff
@@ -1032,6 +1032,29 @@ export const ActionSchema = z
 // Act
 // =============================================================================
 
+export const StagehandResultUsageSchema = z
```
seanmcguire12, line 1022: should we make all of these default 0 ? cc @monadoid

packages/protocol/schemas.ts
```diff
@@ -1040,6 +1063,10 @@ export const StagehandResultMetadataSchema = z
     cacheStatus: CacheStatusSchema.optional().meta({
       description: "Server-side cache status for this result",
     }),
+    usage: StagehandResultUsageSchema.optional().meta({
```
seanmcguire12: & if we make all of the inner fields 0, we could make this required instead of optional? cc @monadoid

shrey150: what if a call doesn't use inference at all (like a deterministic act or cache hit)? shouldn't `usage` be optional in that case?

packages/protocol/schemas.ts
shrey150: I think all values upstream are already coalesced, so we should be able to safely assume they default to `0`

packages/protocol/schemas.ts
seanmcguire12: yeah, the thinking here is having them all be zero'd out would be clear. ie, cache hit, no tokens used, schema is still there but all the token counts etc are `0`. i'll defer to @monadoid to make the call here

monadoid: Yeah agreed, required and zero'd out would be a tighter schema @shrey150

packages/protocol/schemas.ts
monadoid: `.default(0)` in the zod schema will make it more explicit and is generally the pattern we should use here!

packages/server/services/actService.ts
```diff
@@ -65,6 +67,12 @@ export async function act({
   const variables = options?.variables;
   const timeout = options?.timeout;
   const ensureTimeRemaining = createTimeoutGuard(timeout, (ms) => new TimeoutError("act()", ms));
+  let operationUsage: StagehandResultUsage | undefined;
+  const recordUsage = (response: ActInferenceResponse): void => {
+    const nextUsage = usageFromInference(response);
+    operationUsage = operationUsage ? aggregateUsage(operationUsage, nextUsage) : nextUsage;
+  };
+  const resultWithUsage = (result: ActResultData): ActResult => actResult(result, operationUsage);
```
monadoid: I don't think we need to wrap `actResult` in another helper - probably just make `actResult()` return what we need? tbh the `actResult` helper itself shouldn't exist but that is out of scope for this PR

packages/protocol/schemas.ts
shrey150: Addressed in [bd7106ef0](https://github.com/browserbase/stagehand/commit/bd7106ef036d71362c567297cae4fc5d2f03c738) -- added `.default(0)` to all five usage counters and regenerated the protocol and SDK models.

packages/protocol/schemas.ts
shrey150: Addressed in [bd7106ef0](https://github.com/browserbase/stagehand/commit/bd7106ef036d71362c567297cae4fc5d2f03c738) -- made `metadata.usage` required and return a complete zero aggregate for deterministic actions and cache hits.

packages/server/services/actService.ts
shrey150: Addressed in [edcb70888](https://github.com/browserbase/stagehand/commit/edcb70888a4d36b70c30569fd5066222f637819d) -- removed the redundant `resultWithUsage` wrapper and call `actResult` directly with the aggregate.

monadoid (approved):

PR #2535, opened 2026-07-31
feat(cli): add native cookie sync command

shubh24:
## What changed

- add `browse cookies sync` for copying cookies from debuggable local Chrome into a daemon-managed Browserbase session
- require explicit `--domain` scopes, with `--all` as an intentional opt-in
- support persistent Browserbase contexts, saved context names, Verified mode, proxies, named sessions, and explicit source CDP endpoints
- keep cookie values inside the local driver daemon and out of CLI arguments and output
- document the workflow in the CLI README and bundled browse skill

## Why

The existing cookie-sync skill had to create a keep-alive session outside `browse`, inject cookies over raw CDP, and then warn that normal `browse` commands could not reconnect. Making sync native lets the CLI own the destination session from the beginning, so subsequent commands reuse it normally.

## E2E Test Matrix

| --- | --- | --- |
| `pnpm turbo run build --filter=browse...` | The local `browse` build completed successfully from the stacked branch. | Proves the exact CLI and Stagehand core code under review compile together. |
| `pnpm --filter browse lint` | Prettier, ESLint, and TypeScript checks passed. | Static validation only; supports but does not replace runtime proof. |
| `pnpm --filter browse exec vitest run tests/cli-surface.test.ts tests/driver-commands.test.ts tests/remote-options.test.ts tests/local-cdp-discovery.test.ts` | 4 files passed; 76 tests passed. | Covers the CLI surface, driver routing, remote options, cookie-sync handlers, and stale local-CDP discovery. |
| `node packages/cli/bin/run.js cookies sync --domain superhuman.com --persist --session <sync-session>` | Synced 27 domain-scoped cookies from a debuggable local Chrome into a newly created Browserbase context. Auto-discovery recovered from a stale cached Chrome target on the live debug port. | Proves the native command transfers real authenticated cookies and exercises the stacked discovery fix; cookie values and live resource IDs are omitted. |
| Release the sync session, create a new Browserbase session from the new context, then open the public account dashboard and assert DOM auth state. | The fresh session reached the authenticated dashboard with `authenticatedDashboard: true` and `redirectedToLogin: false`; release completed with context persistence. | Proves the newly created context preserves real login state across independent cloud sessions. It does not validate every SSO provider or every site's session-binding policy. |

### DIFF
--- a/.changeset/bright-cookies-sync.md
+++ b/.changeset/bright-cookies-sync.md
@@ -1,0 +1,5 @@
+---
+"browse": minor
+---
+
+Add `browse cookies sync` to securely copy scoped cookies from a debuggable local Chrome browser into a reusable Browserbase CLI session or persistent context.
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -41,6 +41,18 @@ browse screenshot --path page.png
 browse stop
 ```
 
+To copy authenticated cookies from a local Chrome browser into a remote
+Browserbase session, enable Chrome remote debugging and run:
+
+```bash
+browse cookies sync --domain github.com --session github
+browse open https://github.com --session github
+```
+
+Use `--persist` to create a reusable Browserbase context, or `--context <id|name>`
+to refresh an existing one. Cookie sync requires explicit `--domain` flags;
+copying every local cookie requires the deliberate `--all` flag.
+
 ## How it works
 
 `browse` runs a lightweight per-session daemon. The first command starts it, and subsequent commands reuse the same browser — so cookies, tabs, and snapshot refs persist between invocations. Run multiple isolated browsers at once with `--session <name>` (or the `BROWSE_SESSION` env var), and shut a session down with `browse stop`.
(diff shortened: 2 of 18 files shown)


shrey150:
While testing this locally, I ran into a bug in the local-Chrome discovery path this command relies on (\`discoverLocalCdp\`/\`resolveWsTargetFromPort\` in \`local-cdp-discovery.ts\`): it trusts a per-profile \`DevToolsActivePort\` file's recorded (port, target id) without checking whether that port is still serving the *same* browser instance that wrote it. Since 9222 is the near-universal default debugging port, it's easy for a different Chrome process to later reuse that port -- when that happens, sync silently resolves to a dead target and fails with a confusing \`Unexpected server response: 404\`.

This predates this PR (it's been on `main` since the CLI's original monorepo migration, and also affects `browse open --auto-connect` / `browse doctor` today) -- it just becomes a lot more visible here since a failed connect on the sync path reads as "cookie sync is broken." Opened a standalone fix: #2542. Doesn't block this PR, but worth a rebase once it lands so cookie sync benefits too.

Otherwise tested this end-to-end locally (build off this branch, synced a real GitHub login + org SSO cookie set from a local debuggable Chrome into a fresh Browserbase session, landed on the target page fully authenticated with zero manual live-view steps) and the core sync mechanism works well.

PR #2543, opened 2026-08-01
[STG-2749] Surface cache metadata on act/observe/extract results

sameelarif:
# why
Adds cache observability to result metadata. Previously a result only told you `HIT` or `MISS` but never why it missed, how close a key was to its threshold, or what a hit saved. The stateless cache API already returns all of it (`hitCount`, `threshold`, `missReason`); we parsed those fields and passed them straight to `logger.debug`.

# what changed

**Before**

```ts
metadata: {
  actionId?: string;
  cacheStatus?: "HIT" | "MISS";
}
```

**After**

```ts
metadata: {
  actionId?: string;
  cache?: {
    status: "HIT" | "MISS";     // required, replaces metadata.cacheStatus
    count?: number;             // times this key has been seen
    threshold?: number;         // threshold in effect
    missReason?: string;        // misses only
    tokensSaved?: { inputTokens: number; outputTokens: number; totalTokens: number };  // hits only
  };
}
```

`cache` is present exactly when a lookup ran, so `status` being required means it always explains itself:

```ts
const { cache } = result.metadata;
if (!cache) console.log("caching not active");
else if (cache.status === "HIT") console.log(`cached, seen ${cache.count}×`);
else console.log(`live: ${cache.missReason}`);
```

**Breaking:** `metadata.cacheStatus` → `metadata.cache.status`

# test plan

---
## Summary by cubic
Adds a required `metadata.cache` object to all v4 act/observe/extract results to explain hits, misses, and token savings; reports `DISABLED` when no lookup ran. Completes Linear STG-2749 (v4 port of STG-2656) to improve debugging and cost visibility.

- **New Features**
  - `metadata.cache` is required with `status: "HIT" | "MISS" | "DISABLED"`, plus optional `count`, `threshold`, `missReason`, and `tokensSaved` (`inputTokens`, `outputTokens`, `totalTokens` default to 0); removed `ageMs` from results (still logged).
  - Populated across services; distinguishes local failures as `read_failed` and `replay_failed`. `act` now writes aggregate `llmUsage` so hits can report `tokensSaved`.
  - Updated protocol and SDKs (`packages/protocol/*`, `packages/sdk-ts`, `packages/sdk-python`, `packages/sdk-go`), v4 docs, and tests, including end‑to‑end coverage and cache‑client parsing for the `tokensSaved` wire shape.

- **Migration**
  - Replace `result.metadata.cacheStatus` with `result.metadata.cache.status` (TS), `cache_status` → `cache.status` (Python), and the `CacheStatus` pointer with a required `Cache` struct and `Cache.Status` (Go).
  - No lookup is now `result.metadata.cache.status === "DISABLED"` (no null checks).
  - Read token savings from `result.metadata.cache.tokensSaved.{inputTokens,outputTokens,totalTokens}` on hits.

<sup>Written for commit 8ebf30a9397ecf1adba0cbc8dd8dbba4370b0a7c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2543?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/protocol/schemas.ts
```diff
@@ -1034,13 +1034,49 @@ export const ActionSchema = z
 // Act
 // =============================================================================
 
+/** LLM tokens avoided by serving a request from the cache. */
+export const CacheTokenSavingsSchema = z
+  .strictObject({
+    inputTokens: z.number().int().nonnegative(),
```
monadoid: Can each of these be `.default(0)`? This is a nit but ensures that we'll always pass this back

packages/protocol/schemas.ts
```diff
@@ -1034,13 +1034,49 @@ export const ActionSchema = z
...
+  .meta({ id: "CacheMetadata" });
+
 export const StagehandResultMetadataSchema = z
   .strictObject({
     actionId: z.string().optional().meta({
       description: "Action ID for tracking",
     }),
-    cacheStatus: CacheStatusSchema.optional().meta({
-      description: "Server-side cache status for this result",
+    cache: CacheMetadataSchema.optional().meta({
```
monadoid: This should not be `.optional()`

monadoid (approved):

shrey150 (approved):

PR #2568, opened 2026-08-04
Support Browserbase automatic model routing

monadoid:
## Summary

Allows `act`, `extract`, and `observe` to run without a configured model when using Browserbase.

Requests without a model use the Browserbase Model Gateway's automatic model router. Explicitly configured models keep their existing routing behavior.

Updates the TypeScript, Python, and Go examples, generated models, documentation, and regression coverage.

---
## Summary by cubic
Add automatic model routing for Browserbase so `act`, `extract`, and `observe` run without a configured model. Run with only a Browserbase API key; explicit models still behave the same.

- **New Features**
  - If `model` is omitted in a Browserbase session, Browserbase Model Gateway selects a model automatically for each call.
  - Gateway requests omit the `model` field in auto mode; the server uses an internal `auto` placeholder for runtime mapping.
  - Pinned models still work; if a model is set without a provider API key, requests route via the gateway. Missing/unauthenticated models still error when no Browserbase session is available.
  - Updated protocol schemas and generated SDKs (TypeScript, Python, Go), examples, docs, and tests to reflect the new behavior.

- **Migration**
  - To use auto-routing: omit `model` and any provider API keys, and run with Browserbase browsers.
  - No changes required for existing configs; local/CDP browsers still require a provider API key.

<sup>Written for commit 701c9ff164d0af3dd7be2a39b76fbab18ef3a3e0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2568?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


akeimach (approved):
Agree on the try/catch cubic comment. Thanks for updating the docs too!!

PR #2608, opened 2026-08-05
add `_experimental_batch`

seanmcguire12:
# why

client-server round trips can be expensive on remote browsers. this PR adds an experimental batch API that runs a callback inside the Stagehand extension service worker. commands issued by that callback route directly to the existing worker runtime, avoiding the client-server round trip between each operation

# what changed

- added `stagehand._experimental_batch()` to the TypeScript and Python SDKs and `Stagehand.ExperimentalBatch()` to Go
- TS accepts an async JavaScript callback. Python and Go accept self-contained JavaScript source, because callbacks cannot be translated between languages
- installed `globalThis.__stagehandRunCallbackBatch` in the service worker, which selects the active or explicitly requested page, constructs the callback context, applies the overall timeout, invokes the callback, and returns a JSON result/error envelope
- added a transport-independent `StagehandCommandClient` interface and moved the public `Page`, `Locator`, `BrowserContext`, `Response`, clipboard, and WebMCP wrappers onto it
  - regular SDK calls use the remote RPC client, while batch callbacks use an in-browser client backed by the existing `RPCRouter`
- preserved normal Stagehand routing and validation inside batches, including deep locator resolution, response handles, context page registration, WebMCP wrappers, and Zod protocol validation
- exposed `page`, `context`, `act`, `observe`, `extract`, and `metrics` to callbacks. the callback context intentionally excludes `context.close()` and does not expose Stagehand lifecycle or recursive batch operations
- added one-batch-at-a-time protection and cooperative overall timeouts. an operation already running in the router may still finish after the caller receives a timeout
- added JSON validation for callback input and output, explicit handling for `undefined`, and reconstructed worker errors in the calling SDK
- kept host-only conveniences out of the worker contract:
  - screenshot `path` cannot write to the caller's filesystem;
  - local paths cannot be used for file uploads or init scripts;
  - worker screenshot bytes are `Uint8Array` rather than a Node `Buffer`
- added matching TS, Python, & Go examples and documented the TS/Python APIs in the v4 Stagehand reference docs
- updated cross-SDK, example, and docs parity checks to treat exported Go `ExperimentalBatch` as the equivalent of `_experimental_batch` in TypeScript and Python

TypeScript:

```ts
const result = await stagehand._experimental_batch(
  async ({ page }, input) => {
    await page.goto(input.url);
    return {
      title: await page.title(),
      heading: await page.locator("h1").innerText(),
    };
  },
  { url: "https://example.com" },
  { timeout: 30_000 },
);
```

Python:

```python
result = await stagehand._experimental_batch(
    """
    async ({ page }, input) => {
      await page.goto(input.url);
      return { title: await page.title() };
    }
    """,
    {"url": "https://example.com"},
    timeout=30_000,
)
```

Go:

```go
var result struct {
    Title string `json:"title"`
}

err := client.ExperimentalBatch(
    ctx,
    `async ({ page }, input) => {
        await page.goto(input.url)
        return { title: await page.title() }
    }`,
    map[string]any{"url": "https://example.com"},
    &result,
    stagehand.ExperimentalBatchOptions{Timeout: 30 * time.Second},
)
```

# test plan

- [x] Verify the service worker runner routes shared `Page` and `Locator` operations through the in-browser RPC router
- [x] Verify `context.close()` is absent from the callback surface
- [x] Verify callback input is serialized independently from callback source and options
- [x] Verify ordinary JSON results and the distinct `undefined` result envelope
- [x] Verify the generated CDP expression includes the worker capability guard and uses `awaitPromise`/`returnByValue`.
- [x] Verify the TypeScript API is async and forwards callback source, input, page, and timeout
- [x] Verify Python CDP evaluation, static typing, and transport compatibility

---
## Summary by cubic
Adds an experimental batch API that runs a trusted JavaScript callback inside the Stagehand service worker to speed up multi-step flows by skipping SDK↔browser round trips. The API is `stagehand.experimentalBatch()` (TypeScript), `stagehand.experimental_batch()` (Python), and `Stagehand.ExperimentalBatch()` (Go); `page.screenshot()` now returns `Uint8Array`.

- **New Features**
  - Run batches via JSON‑RPC `stagehand.callback_batch`; the CDP transport attaches the executable `callback` with `Runtime.evaluate` while using the normal pending RPC path.
  - Batch context exposes `{ page, context, act, observe, extract, metrics }`, returns JSON‑serializable results (preserves `undefined`), and preserves callback names with a bundled `__name` helper.
  - AI methods in a batch use the SDK’s active‑page default; `options.page` only sets `batch.page`.
  - Service worker: enforces one active batch, blocks internal Stagehand/context APIs, routes ops to the right page, propagates cancellation/timeouts, normalizes extract results, and forwards W3C trace context.
  - Refactor: introduced transport‑independent `StagehandCommandClient`; public wrappers (`Page`, `Locator`, `BrowserContext`, `Response`, clipboard, `WebMCP`) use it for in‑worker execution.
  - Docs and examples added for the new API; docs updated to show `page.screenshot()` returns `Uint8Array`.

- **Bug Fixes**
  - Timeouts: standardized validation across SDKs; include `stagehand.callback_batch` in the RPC timeout policy; cap batch timeouts at `2_147_483_647 - 10_000` ms.
  - JSON/typing: Python generator preserves JSON integers in recursive schemas; fixed `StagehandMetrics` typing; all SDKs validate inputs, reject empty/null options and non‑string sources; Python checks for a closed RPC client.
  - Binary/FS: canonical base64 decoder; `page.screenshot()` returns `Uint8Array`; disallow `path` in batches; avoid double‑encoding and count bytes before encoding; throw `RangeError` for uploads >50MB; Node‑only FS reads are dynamically imported with clearer errors (including init scripts).
  - Robustness: fail fast on invalid worker responses; abort the callback controller after it settles.

<sup>Written for commit e5a1b1c3ebd369c6176786cb63c34ec9923247da. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2608?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seanmcguire12:
@cubic-dev-ai

packages/sdk-python/src/stagehand/stagehand.py
```diff
@@ -171,6 +174,41 @@ async def metrics(self) -> StagehandMetrics:
...
+            json.dumps(input, allow_nan=False)
+        except (TypeError, ValueError) as error:
+            raise TypeError("input must be JSON-serializable") from error
+        result = await self._connected_rpc_client.send(
+            "stagehand.callback_batch",
+            CallbackBatchParams(
+                callback_source=source,
+                input=_models.FieldSchema2.model_validate(input),
+                options=CallbackBatchOptions(
+                    page_id=page.page_id if page is not None else None,
```
miguelg719: **Blocking — Python default path is broken.**

`page_id=page.page_id if page is not None else None` always *sets* the field, so `model_dump(exclude_unset=True, by_alias=True)` keeps it and the wire carries `"pageId": null`. The worker schema is `pageId: z.string().min(1).optional()` — optional but not nullable — so it rejects `null`. Every Python `_experimental_batch()` call without `page=` fails with invalid-params before the callback runs, including `examples/batch.py`.

```python
options=CallbackBatchOptions(
    **({"page_id": page.page_id} if page is not None else {}),
    timeout=timeout,
),
```

Matches the TS conditional spread and Go's `*string` + `omitempty`.

packages/sdk-python/tests/test_stagehand.py
```diff
@@ -139,6 +140,80 @@ def test_stagehand_does_not_expose_context() -> None:
...
+                cast(str, source),
+                input_value,
+                timeout=cast(int, timeout),
+            )
+    finally:
+        await stagehand.close()
+
+
+@pytest.mark.asyncio
+async def test_experimental_batch_uses_registered_rpc_method(
```
miguelg719, line 180: **Blocking — no wire-conformance coverage.**

These tests assert against a recording fake, so nothing ever runs real serialized `callback_batch` params through the zod schemas the extension enforces — which is exactly why the `pageId: null` bug above slipped through. Please add a round-trip test (real RPCClient serialization → protocol schema parse) covering both page-omitted and page-provided cases, in Python and Go (shared golden wire fixtures would be ideal).

miguelg719 (commented):
Two blocking items before merge (correctness, not security — the architecture is sound). Remaining should-fix / nits to follow.

packages/extension/callbackBatch.ts
```diff
@@ -0,0 +1,228 @@
...
+          return await client.send(StagehandMethods.stagehandObserve, {
+            pageId: (operationPage ?? page).pageId,
+            ...(instruction === undefined ? {} : { instruction }),
+            ...(operationOptions === undefined ? {} : { options: protocolOptions }),
+          });
+        },
+        extract: async (...args) => {
+          const [instruction, schemaOrOptions, explicitOptions] = args;
+          const optionsOnly =
+            args.length < 3 &&
```
miguelg719, line 148: extract() only treats arg2 as options if it strict-parses. a typo'd options obj (`{ selctor, timeout }`) fails the parse and gets shipped as the extraction schema instead — runs unscoped, returns garbage, no error (main sdk throws). reverse edge: a permissive `{}` schema parses as options and gets dropped. throw when keys intersect the options schema but don't fully validate.

packages/extension/callbackBatch.ts
```diff
@@ -0,0 +1,228 @@
...
+
+      const stagehand: CallbackStagehand = {
+        page,
+        context: createCallbackContextFacade(context),
+        act: async (instruction, operationOptions) => {
+          const { page: operationPage, ...protocolOptions } = StagehandClientActOptionsSchema.parse(
+            operationOptions ?? {},
+          );
+          return await client.send(StagehandMethods.stagehandAct, {
+            pageId: (operationPage ?? page).pageId,
```
miguelg719: `(operationPage ?? page).pageId` pins to the page resolved at batch start, but the outer sdk resolves activePage() per call. `newPage` + `setActivePage` + `act` in a callback silently drives the original page. resolve per-op or document the pinning.

packages/sdk-ts/src/stagehand.ts
```diff
@@ -110,6 +115,51 @@ export class Stagehand {
...
+    if (options === null || typeof options !== "object" || Array.isArray(options)) {
+      throw new TypeError("stagehand._experimental_batch() options must be an object");
+    }
+    const parsedInput = input === undefined ? undefined : z.json().parse(input);
+    const timeout = options.timeout ?? 30_000;
+    if (!Number.isFinite(timeout) || timeout <= 0) {
+      throw new RangeError("stagehand._experimental_batch() timeout must be greater than zero");
+    }
+    const callbackSource = Function.prototype.toString.call(callback);
+    if (nativeFunctionSourcePattern.test(callbackSource)) {
```
miguelg719, line 149: nativeFunctionSourcePattern only catches native stubs. a valid shorthand method `({ async run(){} }).run` serializes to `async run(){...}` → interpolates as `(async run(){...})` → opaque v8 SyntaxError with no mention of _experimental_batch. validate with `new Function(\`return (${source})\`)` and throw the clean TypeError. also worth a comment that this is dx-only, not a security guard — and it's missing from py/go.

packages/sdk-python/src/stagehand/stagehand.py
```diff
@@ -171,6 +174,41 @@ async def metrics(self) -> StagehandMetrics:
...
+                    timeout=timeout,
+                ),
+            ),
+            CallbackBatchResult,
+        )
+        return (
+            result.value.model_dump(mode="json")
```
miguelg719, line 228: `value` validates through FieldSchema3 (no StrictInt), so every int in the result comes back as float — `{count: 7}` → `7.0` — and inputs lose precision above 2^53. ts/go stay exact. use raw json pass-through (like go's json.RawMessage) or add StrictInt to the unions.

packages/extension/callbackBatch.ts
```diff
@@ -0,0 +1,228 @@
...
+        callbackPromise,
+        new Promise<never>((_, reject) => {
+          controller.signal.addEventListener("abort", () => reject(controller.signal.reason), {
+            once: true,
+          });
+        }),
+      ]);
+      if (result === undefined) return {};
+      return { value: jsonRoundTrip(result) };
+    } finally {
```
miguelg719, line 181: finally only clears the timeout, never aborts the controller after the callback settles. the InProcessCommandClient stays live and any un-awaited op the callback kicked off keeps running in the worker after the result returned — mutating page state under the user's next commands. `controller.abort(...)` in finally.

packages/extension/callbackBatch.ts
```diff
@@ -0,0 +1,228 @@
...
+    const callback = runtimeAttachments?.callback;
+    const { input, options } = params;
+    if (typeof callback !== "function") {
+      throw new TypeError(
+        "Stagehand callback batch request is missing its runtime callback attachment",
+      );
+    }
+
+    const controller = new AbortController();
+    const timeoutId = setTimeout(() => {
```
miguelg719, line 98: timeout allows up to MAX_SAFE_INTEGER but goes straight to setTimeout, whose delay is signed 32-bit in chromium — anything > 2^31-1 ms wraps and fires immediately, aborting instantly instead of ~never. add `.max(2**31-1)` to the schema, mirror in py/go.

packages/extension/callbackBatch.ts
```diff
@@ -0,0 +1,228 @@
...
+  observe(instruction?: string, options?: StagehandClientObserveOptions): Promise<unknown>;
+  extract(
+    instruction: string,
+    schemaOrOptions?: unknown,
+    options?: StagehandClientExtractOptions,
+  ): Promise<unknown>;
+  metrics(): Promise<StagehandMetrics>;
+};
+
+export function createCallbackBatchController(router: RPCRouter) {
```
miguelg719, line 84: the description + test plan claim one-batch-at-a-time protection, but nothing here / in rpcRouter / in any sdk serializes batches, and callback-batch.test.ts has a test asserting concurrent batches run. no new hazard vs concurrent act(), but the claim and the code disagree — implement an activeBatch gate or fix the description.

packages/sdk-ts/src/fileUpload.ts
```diff
@@ -68,14 +76,21 @@ async function normalizeFile(file: string | FilePayload): Promise<InputFilePaylo
...
-  if (byteLength > MAX_INPUT_FILE_BYTES) {
+  if (typeof file.buffer === "string") {
+    const nodeBuffer = (globalThis as typeof globalThis & { Buffer?: typeof Buffer }).Buffer;
+    if (nodeBuffer && nodeBuffer.byteLength(file.buffer, "utf8") > MAX_INPUT_FILE_BYTES) {
+      throw new RangeError(`setInputFiles(): file is larger than the 50 MiB upload limit`);
+    }
+  }
+  const bytes =
+    typeof file.buffer === "string"
+      ? new TextEncoder().encode(file.buffer)
```
miguelg719, line 90: the 50 MiB pre-check for string buffers is gated on globalThis.Buffer, which doesn't exist in the worker. off-node, `new TextEncoder().encode(...)` fully allocates before the RangeError. compute utf-8 length without the full allocation for that path.

packages/extension/tests/callback-batch.test.ts
```diff
@@ -0,0 +1,332 @@
...
+        if (request.method === "context.active_page") return { pageId: "page-1" };
+        throw new Error(`Unexpected method: ${request.method}`);
+      }),
+    } as unknown as RPCRouter;
+    await expect(
+      runCallbackBatch(router, async () => undefined, null, { timeout: 1_000 }),
+    ).resolves.toEqual({});
+  });
+
+  it("enforces the callback timeout in the worker", async () => {
```
miguelg719, line 153: the timeout test uses a never-settling callback and only asserts the race rejection, so throwIfAborted — the only thing stopping a timed-out callback from issuing more act/extract after failure — is never exercised; deleting it still passes. add a gated-callback test asserting a post-timeout send rejects and never reaches router.handle.

miguelg719 (commented):
rest of the review — should-fix + nits. none of these block on their own, but the page-pinning and extract heuristic ones are easy footguns.

packages/sdk-python/src/stagehand/stagehand.py
```diff
@@ -171,6 +174,41 @@ async def metrics(self) -> StagehandMetrics:
...
+            raise ValueError("timeout must be a positive number of milliseconds")
+        try:
+            json.dumps(input, allow_nan=False)
+        except (TypeError, ValueError) as error:
+            raise TypeError("input must be JSON-serializable") from error
+        result = await self._connected_rpc_client.send(
+            "stagehand.callback_batch",
+            CallbackBatchParams(
+                callback_source=source,
+                input=_models.FieldSchema2.model_validate(input),
```
miguelg719: two small parity nits here: (1) non-string dict keys pass the json.dumps guard but then raise a raw pydantic ValidationError instead of the documented TypeError — `model_validate_json(json.dumps(input))` coerces keys consistently. (2) a no-input callback sees `input === undefined` in ts (field omitted) but `null` in py/go. omit the field when unset or document the diff.

packages/sdk-ts/src/stagehand.ts
seanmcguire12: intentionally not fixing

packages/extension/callbackBatch.ts
seanmcguire12: best fix for this would be to nest extract schema into options. will think on this & potentially do it in a later PR

miguelg719 (approved):

PR #2642, opened 2026-08-07
[Docs] Make model routing logic clearer

akeimach:
# why

Making sure the recently added instructions in https://github.com/browserbase/stagehand/pull/2408 are also present in the v4 docs

# what changed

# test plan

---
## Summary by cubic
Clarifies v4 model routing in `packages/docs/v4/configuration/models.mdx`, aligning v4 with STG-2795. Adds a routing table, explains per‑call server-side selection when `model` is omitted, notes Gateway requires Browserbase-hosted browsers and doesn’t support `stopSequences`, and expands troubleshooting for missing LLM config, missing Browserbase session or provider key on pinned models, and stop‑sequence errors.

<sup>Written for commit db0b5c9c59f964bdd16f2ea01f5dba35e312216b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2642?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/docs/v4/configuration/models.mdx
+++ b/packages/docs/v4/configuration/models.mdx
@@ -74,7 +74,7 @@ What you pass decides where a call goes:
 </Note>
 
 <Note>
-  Omitting `model` from `Stagehand.create()` enables routing for that Stagehand instance. Selection itself happens per call, so one run can use different models at different steps. [Per-call overrides](#per-call-model-overrides) enables selecting a specific model instead of routing it.
+  Omitting `model` from `Stagehand.create()` enables routing for that Stagehand instance. Selection itself happens per call, so one run can use different models at different steps. [Per-call overrides](#per-call-model-overrides) enable selecting a specific model instead of routing it.
 </Note>
 
 <Warning>
--- a/packages/docs/v4/configuration/models.mdx
+++ b/packages/docs/v4/configuration/models.mdx
@@ -58,11 +58,28 @@ defer func() { err = errors.Join(err, client.Close(ctx)) }()
 ```
 </View>
 
-When no model is configured, Stagehand routes inference through Model Gateway without a `model` field and Browserbase selects one automatically. An explicit model without a provider-specific API key still routes through Model Gateway, but pins that model instead.
+When no model is configured, Stagehand routes inference through Model Gateway without a `model` field and Browserbase selects one automatically. Selection happens server-side on every call, so your code never pins a model name and picks up new models as Browserbase adds them.
+
+What you pass decides where a call goes:
+
+| Configuration | Where inference runs |
+| --- | --- |
+| No `model` | Model Gateway, with Browserbase selecting the model per call |
+| `model` with no `apiKey` | Model Gateway, pinned to that model |
+| `model` with an `apiKey` | Straight to that provider, bypassing Gateway |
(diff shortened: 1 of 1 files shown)


packages/docs/v4/configuration/models.mdx
```diff
@@ -58,12 +58,29 @@ defer func() { err = errors.Join(err, client.Close(ctx)) }()
 ```
 </View>
 
-When no model is configured, Stagehand routes inference through Model Gateway without a `model` field and Browserbase selects one automatically. An explicit model without a provider-specific API key still routes through Model Gateway, but pins that model instead.
+When no model is configured, Stagehand routes inference through Model Gateway without a `model` field and Browserbase selects one automatically. Selection happens server-side on every call, so your code never pins a model name and picks up new models as Browserbase adds them.
+
+What you pass decides where a call goes:
```
filip-michalsky, line 63: the "auto" model name is accepted by the Stagehand V3 api service please see here https://github.com/browserbase/core/blob/main/apps/stagehand-api-v3/lib/auto-mode/auto-model-router.ts#L8

packages/docs/v4/configuration/models.mdx
```diff
@@ -58,12 +58,29 @@ defer func() { err = errors.Join(err, client.Close(ctx)) }()
...
+| `model` with no `apiKey` | Model Gateway, pinned to that model |
+| `model` with an `apiKey` | Straight to that provider, bypassing Gateway |
+| A client-side LLM callback | Your callback, bypassing Gateway. See [bring your own LLM](#custom-models) |
 
 <Note>
   Model Gateway requires Browserbase-hosted browsers. It does not work with local browsers, because those have no Browserbase session to bill and authorize against.
 </Note>
 
+<Note>
+  Routing is per instance, not per call. There is no `"auto"` model name: every `modelName` carries a provider prefix, so to route a call you omit `model` rather than naming a router. A [per-call override](#per-call-model-overrides) pins that call to a specific model.
```
filip-michalsky: hmmm... what does it mean "per instance"? we do currently route per call-
mode: "auto"
stagehand.act -> auto

next step in the trajectory
stagehand.extract -> auto

etc.

currently, each step on trajectory fed into act/extract/observe is auto-routed. 

There is a very nuanced interplay with caching as well...

filip-michalsky (dismissed):
some small adjustments needed otherwise the errors enum is a nice touch!

packages/docs/v4/configuration/models.mdx
akeimach: I updated this a bit for clarity. But this is a stagehand "instance" if that clears things up!

```
const stagehand = await Stagehand.create({
  browser,
  model: {
    modelName: "openai/gpt-5.4-mini",
    apiKey: process.env.OPENAI_API_KEY,
  },
});
```

packages/docs/v4/configuration/models.mdx
akeimach: talked offline, v4 is different from v3 api so this is still valid

miguelg719 (approved):

PR #2682, opened 2026-08-10
fix(core): detect Bedrock Anthropic model IDs

### LINKED ISSUE #2565: Bedrock-routed models never get provider-specific structuredOutputMode (inferProviderName can't parse Bedrock IDs)
### Bug — Bedrock-routed models never get provider-specific structuredOutputMode

`AISdkClient.createChatCompletion` (`lib/v3/llm/aisdk.js`) infers the AI SDK provider name to select `providerOptions` (including `providerOptions.anthropic.structuredOutputMode`) like this:

```js
function inferProviderName(modelId) {
    const [providerName] = modelId.split("/");
    return providerName || undefined;
}
```

AWS Bedrock model IDs never contain a `/` — neither the raw Bedrock ID format (`us.anthropic.claude-opus-5`) nor cross-region inference profiles (`us.anthropic.claude-sonnet-4-6[1m]`). So `providerName` ends up being the *entire* ID string, which matches none of the switch's cases (`openai`, `anthropic`, `azure`, `google`, `vertex`, `groq`, `cerebras`, `mistral`). `providerOptions` stays `{}` for every Bedrock-routed model, silently — no warning is logged when `providerName` matches nothing.

### This affects Stagehand's own first-party Bedrock support, not just custom model construction

Traced this against the `bedrock/{vendor}.{model}` convention added in #1617. `LLMProvider.js:getClient` strips the `bedrock/` prefix before calling the AI SDK provider factory, so for `modelName: "bedrock/anthropic.claude-3"`, `subModelName = "anthropic.claude-3"` — and the resulting model's `modelId` is exactly that, with no `bedrock/` or `anthropic/` prefix. Confirmed directly:

```js
const { bedrock } = require("@ai-sdk/amazon-bedrock");
bedrock("anthropic.claude-3").modelId // => "anthropic.claude-3"
```

`inferProviderName("anthropic.claude-3")` still returns the whole string, matching no case. So this isn't specific to any downstream consumer's model-construction code — it's a gap in Stagehand's own documented Bedrock path.

### Why this matters — same failure signature as #1986

#2120 added `case "anthropic": providerOptions.anthropic = { structuredOutputMode: "auto" }` specifically because `@ai-sdk/anthropic` defaults to `structuredOutputMode: "jsonTool"` (a tool-call-based JSON extraction) when the option is unset, and that PR's own description notes it "restores structured outputs and prevents schema/response errors." Since `inferProviderName` never resolves to `"anthropic"` for Bedrock IDs, that fix never engages for Bedrock — every Bedrock-routed Claude call is permanently stuck on the `jsonTool` fallback, regardless of whether the underlying model supports native structured output (checked `getModelCapabilities` in `@ai-sdk/anthropic` — e.g. `claude-opus-5` resolves `supportsStructuredOutput: true` via substring match, so the fix would be immediately effective once reachable).

This is the same failure class as #1986 (`act()` response wrapped in `$PARAMETER_NAME`, `AI_NoObjectGeneratedError`) — that issue's root cause is exactly the `jsonTool` tool-call-wrapping path that Bedrock has no way to escape from today. We hit this in production: intermittent `NoObjectGeneratedError: response did not match schema` failures on `act()` calls against `us.anthropic.claude-opus-5` via Bedrock, with `"no actionable element returned by LLM"` warnings preceding the final error.

### Reproduction

```js
const { bedrock } = require("@ai-sdk/amazon-bedrock");
console.log(bedrock("anthropic.claude-3").modelId);
// => "anthropic.claude-3"  (no "/", so inferProviderName can't detect "anthropic")
```

Or end-to-end: configure `Stagehand({ model: "bedrock/anthropic.claude-opus-5", ... })`, drive a few `act()` calls, and observe `providerOptions` is `{}` on the `generateObject` call inside `AISdkClient.createChatCompletion`.

### Suggestion

Not proposing a specific implementation — but `inferProviderName` needs to recognize Bedrock-native ID shapes (`{region-prefix}.{vendor}.{model}` or `{vendor}.{model}`) in addition to the `provider/model` shape it already handles, at least for `anthropic.` (the vendor with an existing, working `providerOptions.anthropic` branch). We've patched this locally via `patch-package`; happy to share the diff if useful.

uuzzrm:
Fixes #2565

`AISdkClient` infers provider-specific options from the model ID. AWS Bedrock Anthropic IDs use `anthropic.<model>` or `region.anthropic.<model>`, so they do not match the existing `provider/model` format and miss Anthropic's `structuredOutputMode: "auto"` option.

This change:

- recognizes raw and cross-region Bedrock Anthropic model IDs
- leaves the existing slash-separated and legacy model ID behavior unchanged
- adds regression coverage for both Bedrock ID shapes
- includes a patch changeset for `@browserbasehq/stagehand`

Verification:

- Prettier check passed
- ESLint passed for the changed TypeScript files
- TypeScript typecheck passed
- ESM and CJS builds passed
- Focused Vitest suite passed (12 tests)

### DIFF
--- a/.changeset/bedrock-anthropic-provider-options.md
+++ b/.changeset/bedrock-anthropic-provider-options.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+Fix provider-specific structured-output options for Anthropic models routed through AWS Bedrock.
--- a/packages/core/lib/v3/llm/aisdk.ts
+++ b/packages/core/lib/v3/llm/aisdk.ts
@@ -29,6 +29,16 @@ type ProviderOptionMap = Record<string, ProviderOptionValue>;
 
 function inferProviderName(modelId: string): string | undefined {
   const [providerName] = modelId.split("/");
+  if (providerName !== modelId) {
+    return providerName || undefined;
+  }
+
+  // Bedrock model IDs use `vendor.model` or `region.vendor.model` instead of
+  // the `provider/model` format used by the other AI SDK providers.
+  if (/^(?:[a-z0-9-]+\.)?anthropic\./i.test(modelId)) {
+    return "anthropic";
+  }
+
   return providerName || undefined;
 }
 
--- a/packages/core/tests/unit/aisdk-clients.test.ts
+++ b/packages/core/tests/unit/aisdk-clients.test.ts
@@ -40,6 +40,11 @@ describe("AISdkClient structured output provider options", () => {
 
   it.each([
     ["openai/gpt-4.1", { openai: { strictJsonSchema: true } }],
+    ["anthropic.claude-3", { anthropic: { structuredOutputMode: "auto" } }],
+    [
+      "us.anthropic.claude-sonnet-4-6[1m]",
+      { anthropic: { structuredOutputMode: "auto" } },
+    ],
     ["azure/gpt-4.1", { azure: { strictJsonSchema: true } }],
     ["google/gemini-2.5-pro", { google: { structuredOutputs: true } }],
     ["vertex/gemini-2.5-pro", { vertex: { structuredOutputs: true } }],


uuzzrm:
Local verification is complete: the core build, lint, typecheck, and focused Bedrock regression tests pass. The GitHub Actions runs currently show `action_required` with no jobs, so they appear to be waiting for maintainer approval for fork workflow runs. I will address any CI feedback once the workflows start.

PR #2706, opened 2026-08-13
fix(extension): preserve native XPath predicates

### LINKED ISSUE #2693: XPath predicates resolve differently once any shadow root exists on the page
`page.locator("xpath=...")` runs through two different XPath engines, and which one you get depends on the page rather than on the selector. `resolveXPathAtIndex` and `countXPathMatches` call `documentHasShadowRoot()` first, and if the document contains any shadow root at all, native `document.evaluate` is skipped and the composed-tree parser in `dom/locatorScripts/xpathParser.ts` handles everything.

The parser says as much in its own header comment ("Unsupported predicates are silently ignored"), and I have no problem with that part. What got me is that the switch is flipped by something that has nothing to do with my selector. A third party widget, a consent banner or any custom element with a shadow root changes what the same locator returns, and neither the API nor the error mentions it.

## Repro

`page.html`, served over http (file:// and data: are on the fallback protocol list themselves, so they don't show the effect):

```html
<section id="s1"><div class="row">A1</div><div class="row">A2</div></section>
<section id="s2"><div class="row">B1</div><div class="row">B2</div></section>
<div id="widget"></div>
<pre id="native"></pre>
<script>
  const CASES = ["//div[2]", "//div[0]", "//div[last()]", "//div[position()>1]", "//section[@id='s1']/div[2]", "(//div)[2]"];
  if (new URLSearchParams(location.search).get('shadow') !== '0') {
    document.getElementById('widget').attachShadow({ mode: 'open' }).innerHTML = '<span>unrelated widget</span>';
  }
  document.getElementById('native').textContent = JSON.stringify(
    CASES.map(xp => document.evaluate(xp, document, null, XPathResult.ORDERED_NODE_SNAPSHOT_TYPE, null).snapshotLength)
  );
</script>
```

```js
import { localBrowser, Stagehand } from "@browserbasehq/stagehand";

const CASES = ["//div[2]", "//div[0]", "//div[last()]", "//div[position()>1]", "//section[@id='s1']/div[2]", "(//div)[2]"];
const browser = await localBrowser.launch({ headless: false });
const stagehand = await Stagehand.create({ browser });
const [page] = await browser.context.pages();

async function run(url, label) {
  await page.goto(url);
  const native = JSON.parse(await page.locator("xpath=/html/body/pre").textContent());
  const rows = [];
  for (let i = 0; i < CASES.length; i++) {
    const count = await page.locator("xpath=" + CASES[i]).count();
    rows.push({ xpath: CASES[i], native: native[i], stagehand: count, match: native[i] === count });
  }
  console.log(label);
  console.table(rows);
}

await run("http://127.0.0.1:8731/page.html?shadow=0", "no shadow root in the document");
await run("http://127.0.0.1:8731/page.html", "one unrelated open shadow root");
```

## Result

No shadow root anywhere: everything matches native XPath except `(//div)[2]`, which returns 0.

One unrelated `attachShadow` call later, on the same DOM:

| xpath | native | stagehand |
| --- | --- | --- |
| `//div[2]` | 2 | 1 |
| `//div[0]` | 0 | 1 |
| `//div[last()]` | 3 | 5 |
| `//div[position()>1]` | 2 | 5 |
| `//section[@id='s1']/div[2]` | 1 | 1 |
| `(//div)[2]` | 1 | 0 |

Counts are just the easy thing to print. The part that cost me time is the first match: on `//div[position()>1]` and `//div[last()]` it becomes A1 instead of A2, so `.first().click()` clicks the wrong node and nothing throws. I would have noticed an empty result right away. I only noticed this one after diffing against `document.evaluate`.

## Where it comes from

- `xpathResolver.ts` picks the engine on `documentHasShadowRoot()`
- `xpathParser.ts` turns `[0]` into position 1 via `Math.max(1, Number(input))`, so a predicate that matches nothing in XPath returns an element
- positional predicates are applied to the flattened match list, so `//div[n]` behaves like `(//div)[n]` while `/parent/div[n]` stays correct
- `position()`, `last()`, axes and grouping land in the silently ignored branch

## Suggestion

Two ideas, neither of them touching the shadow DOM support:

1. Try native `document.evaluate` first and fall back to the composed walk only when it finds nothing. Most pages would keep real XPath semantics, and the composed walk would stop running on every locator call.
2. Throw on the predicates the parser does not implement, instead of returning a filtered set that looks fine.

Happy to send a PR if that direction fits. I could not find unit tests for `dom/locatorScripts/` in the v4 tree, the ones from #1679 don't seem to have made the move, so I would add parser coverage along the way.

## Environment

`@browserbasehq/stagehand` 4.0.0, local browser, Chrome 151, Windows 11, Node 24.

KXHXK:
﻿## Summary

- prefer native XPath results for light-DOM matches even when the document contains a shadow root
- make the composed-tree fallback fail closed for unsupported predicates and preserve XPath index-zero semantics
- add parser-level tests and an extension-backed locator regression for an unrelated shadow root

## Why

An unrelated shadow root currently switches every XPath locator to Stagehand's subset parser. Predicates such as `position()` and `last()` are then silently ignored, so the same locator can count or click different elements depending on unrelated page content.

This keeps composed-tree lookup for selectors that need shadow traversal while preserving native XPath semantics whenever the browser can resolve a light-DOM match.

## Testing

- `focused Vitest coverage for parser, grouped XPath routing, and error propagation (7 passed)
- `oxfmt --check` on all changed files
- `oxlint` on all changed TypeScript files
- attempted `packages/sdk-ts/tests/integration/locatorCount.test.ts` with the new HTTP fixture; all 8 tests, including unchanged baselines, reach browser startup but are blocked locally because Chrome cannot resolve the extension source path under the workspace's non-ASCII parent directory (`File path cannot be resolved`). The HTTP and mixed light/shadow cases are included for CI on GitHub's ASCII workspace.

Closes #2693

---
## Summary by cubic
Preserves native XPath predicate semantics when pages include unrelated shadow roots. Previously the composed-tree fallback ignored positional predicates and treated index 0 as the first match; now grouped XPaths like `(//div)[2]` and any path with positional/unsupported predicates evaluate natively, and unsupported predicates in composed traversal throw.

- Resolver: evaluate grouped `(`...`)` paths natively; if predicates are positional or unsupported, prefer native results; otherwise merge native light‑DOM results with shadow‑root matches, dedupe, and order by composed‑tree document order; distinguish empty native results from native errors before fallback.
- Parser/runtime: add an `unsupported` predicate that throws in composed traversal; keep XPath index‑zero semantics (index 0 matches nothing).
- Understudy: preserve grouped XPath strings for native evaluation; expose `evaluateXPathElement`; sanitize “Unsupported XPath predicate in composed-tree traversal” at the CDP boundary; transient main‑world errors still return null/0.
- Tests: cover parser safety, grouped routing, error propagation, native zero‑match semantics, and mixed light/shadow ordering.

**Migration**
- Locators that relied on ignored predicates or index‑0 coercion may now throw or return different counts. Update affected locators/tests to handle thrown XPath errors in `@browserbasehq/stagehand`.

<sup>Written for commit 303ed5f95589a9f5c6d8e1ab5fe037c5609318df. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2706?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/tame-xpaths-preserve-predicates.md
+++ b/.changeset/tame-xpaths-preserve-predicates.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+Preserve native XPath predicate semantics when a page contains an unrelated shadow root.
--- a/packages/extension/dom/locatorScripts/xpathParser.ts
+++ b/packages/extension/dom/locatorScripts/xpathParser.ts
@@ -1,5 +1,5 @@
 export type XPathPredicate =
-  | { type: "unsupported" }
+  | { type: "unsupported"; source: string }
   | { type: "index"; index: number }
   | { type: "attrEquals"; name: string; value: string; normalize?: boolean }
   | { type: "attrExists"; name: string }
@@ -53,8 +53,8 @@ export interface XPathStep {
  *  - Union operator (`|`)
  *  - Grouped expressions (`(//div)[n]`)
  *
- * Unsupported predicates never match in the composed-tree fallback. Native
- * XPath evaluation still handles them when the selector matches the light DOM.
+ * Unsupported predicates throw in the composed-tree fallback so callers never
+ * receive a broadened or incomplete result that looks authoritative.
  */
 export function parseXPathSteps(input: string): XPathStep[] {
   const path = String(input || "")
@@ -153,7 +153,7 @@ function parseStep(raw: string): {
 
   for (const inner of extractPredicates(predicateStr)) {
     const parsed = parsePredicateExpression(inner);
-    predicates.push(parsed ?? { type: "unsupported" });
+    predicates.push(parsed ?? { type: "unsupported", source: inner });
   }
 
   return { tag, predicates };
@@ -382,7 +382,9 @@ function normalizeMaybe(value: string, normalize?: boolean): string {
 export function evaluatePredicate(element: Element, predicate: XPathPredicate): boolean {
   switch (predicate.type) {
     case "unsupported":
-      return false;
+      throw new Error(
+        `Unsupported XPath predicate in composed-tree traversal: ${predicate.source}`,
+      );
     case "and":
       return predicate.predicates.every((p) => evaluatePredicate(element, p));
     case "or":
--- a/packages/extension/dom/locatorScripts/xpathParser.ts
+++ b/packages/extension/dom/locatorScripts/xpathParser.ts
@@ -1,4 +1,5 @@
 export type XPathPredicate =
+  | { type: "unsupported" }
   | { type: "index"; index: number }
   | { type: "attrEquals"; name: string; value: string; normalize?: boolean }
   | { type: "attrExists"; name: string }
@@ -52,8 +53,8 @@ export interface XPathStep {
  *  - Union operator (`|`)
  *  - Grouped expressions (`(//div)[n]`)
  *
- * Unsupported predicates are silently ignored — the step still matches
- * by tag name, but the unrecognized predicate has no filtering effect.
+ * Unsupported predicates never match in the composed-tree fallback. Native
+ * XPath evaluation still handles them when the selector matches the light DOM.
  */
 export function parseXPathSteps(input: string): XPathStep[] {
   const path = String(input || "")
@@ -152,7 +153,7 @@ function parseStep(raw: string): {
 
   for (const inner of extractPredicates(predicateStr)) {
     const parsed = parsePredicateExpression(inner);
-    if (parsed) predicates.push(parsed);
+    predicates.push(parsed ?? { type: "unsupported" });
   }
 
   return { tag, predicates };
@@ -195,7 +196,7 @@ function parseAtomicPredicate(input: string): XPathPredicate | null {
   const quoted = "(?:'([^']*)'|\"([^\"]*)\")";
 
   if (/^\d+$/.test(input)) {
-    return { type: "index", index: Math.max(1, Number(input)) };
+    return { type: "index", index: Number(input) };
   }
 
   const normalizeAttrMatch = input.match(
@@ -380,6 +381,8 @@ function normalizeMaybe(value: string, normalize?: boolean): string {
 
 export function evaluatePredicate(element: Element, predicate: XPathPredicate): boolean {
   switch (predicate.type) {
+    case "unsupported":
+      return false;
     case "and":
       return predicate.predicates.every((p) => evaluatePredicate(element, p));
     case "or":
--- a/packages/extension/dom/locatorScripts/xpathResolver.ts
+++ b/packages/extension/dom/locatorScripts/xpathResolver.ts
@@ -1,4 +1,9 @@
-import { applyPredicates, parseXPathSteps, type XPathStep } from "./xpathParser.js";
+import {
+  applyPredicates,
+  parseXPathSteps,
+  type XPathPredicate,
+  type XPathStep,
+} from "./xpathParser.js";
 import { documentHasShadowRoot, getOpenOrClosedShadowRoot } from "./shadowRoots.js";
 
 type ShadowRootGetter = (host: Element) => ShadowRoot | null;
@@ -54,9 +59,12 @@ export function resolveXPathAtIndex(
   const shadowHopMatches = resolveStagehandShadowHopMatches(xp, shadowCtx.getShadowRoot);
   if (shadowHopMatches.length > 0) return shadowHopMatches[targetIndex] ?? null;
 
-  const native = resolveNativeMatchesWithError(xp);
+  const native = resolveNativeMatches(xp);
+  if (native.length > 0 && requiresNativePredicateSemantics(xp)) {
+    return native[targetIndex] ?? null;
+  }
   const composed = resolveXPathComposedMatches(xp, shadowCtx.getShadowRoot);
-  const matches = mergeXPathMatches(native.values, composed, shadowCtx.getShadowRoot);
+  const matches = mergeXPathMatches(native, composed, shadowCtx.getShadowRoot);
   return matches[targetIndex] ?? null;
 }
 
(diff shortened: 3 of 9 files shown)


mikhail-koviazin:
Thanks for picking this up, and for going at the parser instead of around it.

Four things I would want checked before this lands, roughly in the order they worry me.

**1. The new integration test uses a `data:` URL, and I don't think it reproduces the bug.** That is why the repro in the issue is served over http. `data:` sits on the protocol list in `isFallbackEligible` (`executionContextRegistry.ts`: `data:`, `about:`, `blob:`, `file:`, `filesystem:`), and in my runs neither `data:` nor `file://` showed the effect at all. Could you revert the `xpathResolver.ts` hunk locally and run just that test? If it still passes, it isn't guarding the fix.

**2. Native-first gated on `count > 0` under-counts pages that mix light and shadow DOM.** With `if (!native.error && native.count > 0) return native`, any selector that matches something in the light DOM never reaches the composed walk, so matches inside shadow roots stop being returned. Two divs in the document plus one inside an open shadow root: `//div` used to include the shadow one and now returns only the light-DOM two. That is the case the composed walk exists for, and it is a quieter version of what I reported: same shape, opposite direction, still driven by unrelated page content. A union of native and composed matches, deduplicated and in document order, would keep both halves.

**3. `unsupported` returning `false` keeps the failure silent.** What I asked for in the issue was to throw on predicates the parser doesn't implement. An empty set beats a broadened one, but the caller still can't tell "no such element" from "Stagehand can't parse this predicate". Together with 2, a `position()` selector against content in a shadow root now returns nothing and says nothing.

**4. `(//div)[2]` is still broken, and it doesn't need a shadow root.** It returns 0 in both of my runs, with and without `attachShadow`, so grouping is broken on plain pages too. Since this says `Closes #2693`, either grouping goes in here or the issue should stay open for that part.

Minor: the parser test passes `[{}, {}, {}] as Element[]`, and the `unsupported` branch returns `false` whatever the element is, so `//div[position() > 10]` satisfies that assertion even if the predicate were parsed properly. A case whose correct answer is non-empty would carry more weight.

Happy to run any of this against my repro if it helps.

mikhail-koviazin:
On point 4, to be concrete instead of just critical: I would rather grouping did not get pulled into this PR. It is a different code path. `parseXPathSteps` walks the string from the start expecting `/` or `//`, so a leading `(` never parses into steps at all, and that happens whether or not the document has a shadow root, which is why it was broken in both of my runs.

I will send it as its own PR against `xpathParser.ts` with tests, so this one can stay about the engine choice. That does mean #2693 still has a part open after this merges, which is what I was getting at with the `Closes` line.

KXHXK:
﻿Thanks for the careful repro review. All four points were valid, and I have updated the PR in `2d2b7ca`:

1. Replaced the `data:` regression with the existing local HTTP fixture helper.
2. Native results now supply light-DOM XPath semantics, while composed traversal contributes only elements whose root is a `ShadowRoot`; the two sets are deduplicated and ordered by composed document traversal. A mixed light/shadow test covers this.
3. Unsupported composed predicates now throw with the original predicate text, and the XPath-specific CDP evaluation path propagates that error instead of converting it to `0`/`null`.
4. Grouped expressions such as `(//div)[2]` stay intact and use native XPath evaluation; a focused routing test and HTTP integration assertion cover this.

I also added the positive `[2]` parser control requested by Cubic. Focused unit coverage is green (7 tests), formatting is clean, and the remote compare contains only the intended two commits. The full HTTP integration file still cannot run to assertions on my Windows checkout because Chrome rejects the extension source path under its non-ASCII parent directory; all eight tests, including unchanged baselines, fail at the same extension-load step. CI should exercise these cases in an ASCII workspace.

If you can rerun the updated branch against your original repro, that would be very helpful.

mikhail-koviazin:
Grouping is covered here, so there is no separate PR coming from me. Thanks for turning it around this fast.

The merged ordering reads right: native for light-DOM semantics, composed contributing only shadow-rooted elements, sorted by composed traversal.

KXHXK:
I reviewed the second Cubic pass against the resolver behavior and pushed `adbc1ca`.

The valid findings are addressed as follows:

- XPath expressions with positional or unsupported predicates now keep native results when native evaluation produced light-DOM matches. This avoids applying incompatible flattened composed-tree positional semantics (for example, `//div[2]`) and avoids letting an unrelated shadow root turn a valid native result into an exception.
- Attribute/text predicates that are safe to evaluate per element still merge shadow-root matches with native light-DOM matches, preserving the mixed light/shadow behavior covered by the regression.
- The CDP boundary now exposes only the stable message `Unsupported XPath predicate in composed-tree traversal`; it no longer reflects the selector or raw Runtime exception details.
- Non-predicate Runtime/context failures again resolve as `0`/`null`, while the explicit unsupported-predicate signal still propagates.
- Removed the unused native evaluation error flag.
- Added coverage for transient XPath evaluation failures.

Verification on the updated files:
- 8 focused tests passed across parser, selector resolver, and deep locator suites.
- Oxfmt check passed.
- Oxlint passed.

I kept the grouped-expression regression in this PR because the issue author confirmed it is covered here and no separate PR is planned.

KXHXK:
The zero-native-match finding is valid and is fixed in `303ed5f`.

`resolveNativeMatches` again reports whether native evaluation failed, and that signal is now used: positional/full-XPath predicates return the native result whenever evaluation succeeds, including an empty snapshot. Composed traversal is considered only when native evaluation actually fails. This keeps `//button[last()]` at `0`/`null` on a page with an unrelated shadow root instead of turning it into an unsupported-predicate error.

The HTTP fixture now covers that zero-match case and also reflects the final native-first rule for `//div[position() > 1]` (native count `2`). Focused verification remains green: 8 tests passed, Oxfmt passed, and Oxlint passed. The extension-backed HTTP file remains blocked locally at Chrome extension loading by the non-ASCII workspace path, as noted earlier; CI will run it in its normal workspace.

PR #1683, opened 2026-02-13
[fix]: extend support for xpaths containing predicates

seanmcguire12:
# why
- to make sure `.count()`/`.nth()` include shadow DOM matches 
- make sure common predicates work in shadow DOM selectors (beyond just attribute predicates)
- also to unify xpath resolution into one shared resolver, building off of [this](https://github.com/browserbase/stagehand/pull/1679) PR 
# what changed
- added shared resolver policy in `packages/core/lib/v3/dom/locatorScripts/xpathResolver.ts`
- expanded composed predicate parsing/evaluation in `packages/core/lib/v3/dom/locatorScripts/xpathParser.ts` (attr existence, contains/starts-with, text(), normalize-space, and/or/not)
- routed xpath entrypoints through the resolver:
  - `packages/core/lib/v3/dom/locatorScripts/selectors.ts` (`resolveXPathMainWorld`)
  - `packages/core/lib/v3/dom/locatorScripts/counts.ts` (`countXPathMatchesMainWorld`)
  - `packages/core/lib/v3/dom/locatorScripts/waitForSelector.ts` (xpath path now uses resolver)
- also removed unused injected helper `resolveSimpleXPath` from `packages/core/lib/v3/dom/piercer.runtime.ts` & updated `packages/core/lib/v3/dom/global.d.ts`
# test plan
- `packages/core/tests/xpath-parser.test.ts` (expanded predicate parsing + evaluation coverage)
- `packages/core/tests/xpath-resolver.test.ts` (shadow-inclusive count/nth ordering across composed tree)

---
## Summary by cubic
Extended XPath support for predicates and shadow DOM, unified behind a shared resolver. Fixes .count() and .nth() with document-order DFS across light and shadow DOM, aligning with Linear STG-1373.

- **New Features**
  - Predicate support: attr exists/equals/contains/starts-with; text()/normalize-space; and/or/not; positional [n].
  - Shadow-piercing resolution for selectors, count, and waitForSelector with native document.evaluate fallback and DFS indexing.

- **Refactors**
  - Routed all XPath entry points to the shared resolver; removed legacy traversal and the piercer’s resolveSimpleXPath backdoor.
  - Expanded parser/resolver tests and added jsdom; cleaned dead code in selectors/counts/waitForSelector.

<sup>Written for commit 57f38186d2ada80fac32558928bad51c48c85f61. Summary will update on new commits. <a href="https://cubic.dev/pr/browserbase/stagehand/pull/1683">Review in cubic</a></sup>

PR #1679, opened 2026-02-11
[fix]: `.count()` not working on xpaths with attribute predicates

shrey150:
## Summary

- Fix `locator.count()` returning 0 for XPath selectors containing attribute predicates (e.g. `//img[@alt='Stagehand']`)
- Consolidate all four copies of the XPath step parser into a single shared module with unit tests

## Root Cause

Stagehand has two ways to evaluate XPath expressions:

1. **Native browser XPath** (`document.evaluate()`) — understands the full XPath spec, including attribute filters like `[@alt='Stagehand']`, but cannot see into shadow DOM
2. **Custom XPath parser** — a hand-written DOM traversal that walks the composed tree (including shadow roots), but only understood simple patterns like `//div` or `//div[2]`, not attribute predicates

The `isVisible()` code path (via `resolveXPathMainWorld`) tried native browser XPath first and fell back to the custom parser. But the `count()` code path (`countXPathMatchesMainWorld`) went straight to the custom parser. The custom parser's regex only recognized numeric positional indices (`[1]`, `[2]`), so `[@alt='Stagehand']` was treated as part of the tag name, matching nothing.

### 1. Extract shared parser (`xpathParser.ts`)
The XPath step parser was duplicated four times across `counts.ts`, `selectors.ts`, `piercer.runtime.ts`, and `waitForSelector.ts`. All four are consolidated into a single `xpathParser.ts` module exporting `parseXPathSteps()` and `elementMatchesStep()`.

### 2. Add attribute predicate parsing
The parser now handles `[@attr='value']` predicates (single and double quotes, multiple predicates, combined with positional indices). Attribute predicate XPaths work correctly **including inside shadow DOM**.

### 3. Fix forward slash splitting in predicates
The step splitter now tracks bracket depth so `/` inside attribute values (e.g. `[@href='/api/endpoint']`) no longer incorrectly splits the step. This was a pre-existing bug in three of the four copies; `waitForSelector.ts` had independently fixed it.

### 4. Document supported XPath subset
The parser docstring explicitly lists what's supported and what isn't, so future contributors know the scope.

## Files changed
- `xpathParser.ts` (new) — shared parser with `parseXPathSteps()` and `elementMatchesStep()`
- `counts.ts` — uses shared parser (removed ~40 lines of inline parser)
- `selectors.ts` — uses shared parser (removed ~40 lines of inline parser)
- `piercer.runtime.ts` — uses shared parser (removed ~40 lines of inline parser)
[...]

PR #2728, opened 2026-08-15
fix(extension): read text() from direct child text nodes

mikhail-koviazin:
## why

The composed-tree XPath parser evaluates `text()` and `.` through the same helper, `element.textContent`. In XPath these are not the same thing: `.` is the string-value of the element, so it covers the whole subtree, while `text()` is the node-set of the element's **direct child text nodes**.

This parser is not a rare path. It takes over whenever the document contains a shadow root anywhere, so a single unrelated web component on the page changes what a locator matches, silently and with no error.

Measured against `document.evaluate()` on a build of `main` (`a73da68b`), fixture served over http. The only difference between a run that matches native and a run that does not is one unrelated `attachShadow()` call elsewhere in the same document:

| XPath | `document.evaluate()` | Stagehand |
| --- | --- | --- |
| `//button[text()='Save']` | 1 | 2 |
| `//div[text()='a']` | 1 | 0 |
| `//div[contains(text(),'b')]` | 0 | 1 |
| `//div[@id='split'][text()='y']` | 1 | 0 |
| `//div[@id='split'][contains(text(),'y')]` | 0 | 1 |
| `//div[@id='split'][normalize-space(text())='x']` | 1 | 0 |
| `//button[.='Save']` | 2 | 2 (control: `.` is already correct) |

Fixture: `<button id="wrapped"><span>Save</span></button>` before `<button id="direct">Save</button>`, plus `<div id="mixed">a<span>b</span></div>` and `<div id="split">x<br/>y</div>`.

The count is not the worst part. The button whose label sits inside a `<span>` comes first in document order, so `//button[text()='Save']` returns it first and `.first().click()` clicks the wrong button. Nothing throws, nothing logs, and the run continues on the wrong element. The other direction is a silent zero on markup as ordinary as `a<span>b</span>`.

## what changed

`text()` now reads its own node-set, and `.` keeps exactly the meaning it has today:

- `text()='v'` is true when **any** direct child text node equals `v`. Comparing a node-set to a string is existential in XPath.
- `contains(text(),'v')` and `normalize-space(text())='v'` read the string-value of the **first** node, because both functions take a string, so the node-set collapses.
- `.='v'`, `contains(.,'v')` and `normalize-space(.)='v'` keep reading the string-value of the element, unchanged.

Mechanically, in `packages/extension/dom/locatorScripts/xpathParser.ts`: the three patterns that accepted `(?:text\(\)|\.)` now capture which token they matched, and `textEquals` / `textContains` carry a `source: "self" | "text"` field that `evaluatePredicate` reads. The field is optional and absent means `self`, so predicates constructed anywhere else keep their current behavior. `and`, `or` and `not` carry it through without changes.

## test plan

`packages/extension/tests/xpath-text-predicates.test.ts` (new, unit): the parser keeps `text()` and `.` apart for `=`, `contains()` and `normalize-space()`, and carries the source through `or` and `not`.

`packages/sdk-ts/tests/integration/locatorXPathTextPredicates.test.ts` (new, integration): the table above against real Chrome. It has to be served over http, since `data:` URLs stay on the native engine and never reach this parser at all. Registered in `scripts/test-integration.ts` next to the other locator groups.

Both are red on `main` and green with the change. On `main` the wrong-button case fails with `expected 2 to be 1` and the silent-zero case with `expected +0 to be 1`.

Context: #1679 consolidated the parser copies and #1683 taught this one `text()`, `contains()` and `normalize-space()`. This is the same layer, one step further in.

---
## Summary by cubic
Fixes the composed-tree XPath parser so `text()` reads only direct child text nodes. Previously it read `element.textContent` (same as `.`), causing diverging matches and wrong clicks on pages with any shadow root.

- Old: `text()` behaved like `.`. New: `text()` evaluates direct child text nodes; `.` still reads the element’s subtree string-value. This aligns with `document.evaluate()`.

- If a locator relied on `text()` to match nested text, use `.` instead.

- Adds unit and integration tests; integration runs over http and is registered in scripts/test-integration.ts.

- `text()='v'` is existential across child text nodes; `contains(text(),'v')` and `normalize-space(text())='v'` collapse to the first text node.

- The parser records whether a predicate used `text()` or `.` and preserves it through boolean operators.

<sup>Written for commit ee14180aad7812e27dcb5ecdab25126787049440. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2728?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/xpath-text-node-predicates.md
+++ b/.changeset/xpath-text-node-predicates.md
@@ -1,0 +1,5 @@
+---
+"@browserbasehq/stagehand": patch
+---
+
+Match native XPath semantics for `text()` predicates when a shadow root routes locators through the composed-tree parser
--- a/packages/extension/dom/locatorScripts/xpathParser.ts
+++ b/packages/extension/dom/locatorScripts/xpathParser.ts
@@ -1,3 +1,12 @@
+/**
+ * Which string a text predicate reads.
+ *
+ * `self` is `.` — the string-value of the element, i.e. its whole subtree text.
+ * `text` is `text()` — the node-set of direct child text nodes, which is not the
+ * same string and does not compare the same way (see `evaluatePredicate`).
+ */
+export type XPathTextSource = "self" | "text";
+
 export type XPathPredicate =
   | { type: "index"; index: number }
   | { type: "attrEquals"; name: string; value: string; normalize?: boolean }
@@ -14,8 +23,8 @@ export type XPathPredicate =
       value: string;
       normalize?: boolean;
     }
-  | { type: "textEquals"; value: string; normalize?: boolean }
-  | { type: "textContains"; value: string; normalize?: boolean }
+  | { type: "textEquals"; value: string; normalize?: boolean; source?: XPathTextSource }
+  | { type: "textContains"; value: string; normalize?: boolean; source?: XPathTextSource }
   | { type: "and"; predicates: XPathPredicate[] }
   | { type: "or"; predicates: XPathPredicate[] }
   | { type: "not"; predicate: XPathPredicate };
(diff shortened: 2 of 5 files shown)


mikhail-koviazin:
Friendly ping on this one when you get a chance — self-contained fix in the composed-tree XPath parser: `text()` was reading `element.textContent` (the whole subtree) instead of just direct child text nodes, so it silently diverged from `document.evaluate()` on any page with a shadow root. Table of measured mismatches plus a wrong-click repro is in the PR description. Comes with a unit test, an integration test against real Chrome, and a changeset. cc @seanmcguire12

PR #2752, opened 2026-08-17
Allow SDKs to reattach to Stagehand runtimes

monadoid:
## Summary

Allow a new SDK client to initialize an existing Stagehand extension runtime without replacing its browser session.

This lets stateless clients reconnect to a long-lived Browserbase session. Reinitialization updates the client configuration; closed runtimes and concurrent initialization still fail.

---
## Summary by cubic
Allows SDK clients to reattach to an existing Stagehand runtime without replacing the browser session. Previously a second `stagehand.init` errored; now it succeeds, reuses the session, and updates initialization config. Closed runtimes still reject init; concurrent init still fails.

**Review and rollout**
- Second `stagehand.init` now returns `{ initialized: true, pages: [] }` and updates model/systemPrompt/telemetry; `prepareForInitialization` runs only on first init.
- Closed runtimes throw “Stagehand has been closed and cannot be initialized again”; concurrent initialization still throws.
- Tests updated to cover reattach and closed-runtime behavior.
- Patch releases for `@browserbasehq/stagehand`, `@browserbasehq/stagehand-extension`, `@browserbasehq/stagehand-python`, and `@browserbasehq/stagehand-go`; Go SDK bundle updated.
- Migration: If clients relied on the previous error for repeat init, update them to treat a second `stagehand.init` as a successful no-op that refreshes config.

<sup>Written for commit cc70a9ae167dcf601924d1f9d10df82514295eee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browserbase/stagehand/pull/2752?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MSH4R1F:
+100000, hit exactly this wall and would love to get this merged asap @monadoid 

Use case: a Temporal worker fleet where every browser action is its own stateless activity that can run on any machine, against one keepAlive Browserbase session per account. Each activity does browserbase.connect({ sessionId }) → Stagehand.create({ browser }) → work → close the browser handle. On 4.0.1 the second create from any process fails with Stagehand has already been initialized, so stateless workers cannot share a session. 

We have to do Stagehand.create({browser}) to get access to snapshot.

What we tested: applied this PR's runtime.ts change to the 4.0.1 extension, zipped and uploaded it to Browserbase, created a session with that extensionId, then ran the unpatched @browserbasehq/stagehand@4.0.1 SDK from a parent process and three separate child processes. Every Stagehand.create succeeded (~1.1 s each), same page ids, snapshot() and activePage() fine. So the SDK needs no change; the extension patch release is enough.

Two observations that may be useful for the reattach story:

- stagehand.close() from any client closes the runtime for all later clients (Stagehand has been closed and cannot be initialized again). Expected per the description, but worth a sentence in the reconnect docs: a client that only wants to detach should close the browser handle, not the Stagehand instance.

Not in this PR's scope: the SDK RPCClient starts JSON-RPC ids at 1 for every client, and Chrome fans Runtime.bindingCalled out to every CDP session that registered the binding. Two clients attached at the same time can therefore consume each other's replies (we reproduced this: client A got client B's page.url() result). Once reattach is supported, a per-client random id base (or a client-scoped prefix) would close that gap.

Would love to see this merged and released. Happy to test a prerelease against our setup.

miguelg719 (approved):

PR #2789, opened 2026-08-20
docs: Examples tab for Stagehand cookbooks

antonvishal:
## Summary

This PR adds a dedicated **Examples** section to the Stagehand docs.

The goal is to make Stagehand-specific patterns easier for developers and coding agents to find, understand, and reuse. When someone asks Codex, Claude Code, or another agent how to use Stagehand, the indexed Stagehand docs should provide enough context for the agent to compose a useful implementation.

The Examples section gives Stagehand its own structured home, separate from the broader Browserbase Templates page. It can grow over time as we add more integrations, workflows, and real-world patterns.

## Mockup

Screenshot 2026-08-21 at 5 49 01 PM


charlypoly (commented):
Hi @antonvishal, thank you for this 🙌🏼 

I think this is a core missing piece of our docs, so let's make it happen!

A couple of things I'd like to change:
- The navigation is a bit overcomplicated as it stands, we should add hierarchy (categories) once we reach a critical number of cookbooks (>10-15)
- Each cookbook should come as both a prompt and a folder in the repo to checkout, I'm not sure people spend time copy-pasting these days
- We need to revisit some of them that are closer to snippets than cookbook (not covering a specific enough use cases): Multi-page Workflows, Extract a paginated catalog, Compare pages in parallel. Either we extend them or we revisit the menu to have 2 categories: Cookbook and Snippets/Patterns

antonvishal:
> Hi @antonvishal, thank you for this 🙌🏼
> 
> I think this is a core missing piece of our docs, so let's make it happen!
> 
> A couple of things I'd like to change:
> 
> * The navigation is a bit overcomplicated as it stands, we should add hierarchy (categories) once we reach a critical number of cookbooks (>10-15)
> * Each cookbook should come as both a prompt and a folder in the repo to checkout, I'm not sure people spend time copy-pasting these days
> * We need to revisit some of them that are closer to snippets than cookbook (not covering a specific enough use cases): Multi-page Workflows, Extract a paginated catalog, Compare pages in parallel. Either we extend them or we revisit the menu to have 2 categories: Cookbook and Snippets/Patterns

Hey @charlypoly , I’ve kept the Examples navigation flat, added a prompt and runnable repo folder for each cookbook, and expanded the pages into a consistent guide format. I’ve attached a screenshot of the updated cookbook. How does the page look? If this direction works, I’ll commit the changes. Any other edits you’d like first?

Screenshot 2026-09-24 at 5 15 55 PM

PR #2879, opened 2026-09-02
docs: Search and Fetch Add-Ons

charlypoly:
Documenting https://github.com/browserbase/stagehand/pull/2828

### DIFF
--- a/packages/docs/v4/add-ons/fetch.mdx
+++ b/packages/docs/v4/add-ons/fetch.mdx
@@ -1,0 +1,532 @@
+---
+title: "Web Fetch"
+sidebarTitle: "Web Fetch"
+description: "Turn any webpage into agent context, without launching a browser"
+---
+
+## What is `fetch()`?
+
+Most of the time an agent doesn't need to *use* a page, it needs to *read* one. `browserbase.fetch()` takes a URL and returns its content as markdown, structured JSON, or the raw body — in one HTTP call, with no browser to start, no page to navigate, and nothing to close.
+
+<Tabs>
+<Tab title="TypeScript">
+```typescript
+import { browserbase } from "@browserbasehq/stagehand";
+
+const fetchResult = await browserbase.fetch({
+  apiKey: process.env.BROWSERBASE_API_KEY,
+  url: "https://example.com",
+  format: "markdown",
+});
+
+console.log(fetchResult.content);
+```
+</Tab>
+
+<Tab title="Python">
+```python
+import os
+
+from stagehand import browserbase
+
+fetch_result = await browserbase.fetch(
+    api_key=os.environ["BROWSERBASE_API_KEY"],
+    url="https://example.com",
+    format="markdown",
+)
+
+print(fetch_result.content)
+```
+</Tab>
+
+<Tab title="Go">
+```go
+fetchResult, err := stagehand.FetchBrowserbase(ctx, stagehand.BrowserbaseFetchOptions{
+	APIKey: os.Getenv("BROWSERBASE_API_KEY"),
+	URL:    "https://example.com",
+	Format: stagehand.BrowserbaseFetchFormatMarkdown,
+})
+if err != nil {
+	return err
+}
+
+content, ok := fetchResult.Content.AsString()
+if !ok {
+	return errors.New("fetch returned non-string content")
+}
+fmt.Println(content)
+```
+</Tab>
+</Tabs>
+
+<Note>
+`fetch()` hangs off the same `browserbase` facade you use to launch cloud browsers, but it never starts a browser session. All it needs is a Browserbase API key — there is no `Stagehand` instance and no page. Local browsers have no equivalent.
+</Note>
+
+## Why use `fetch()`?
+
+<CardGroup cols={2}>
+  <Card title="Freshness" icon="rotate">
+    Every call retrieves the live page, not a copy from a crawl index, so what the model reads is what the site serves right now.
+  </Card>
+  <Card title="Speed" icon="bolt">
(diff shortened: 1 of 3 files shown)


packages/docs/v4/add-ons/fetch.mdx
```diff
@@ -0,0 +1,532 @@
+---
+title: "Web Fetch"
```
Kylejeong2: should we call this Stagehand Fetch? 

or alt. just Fetch

packages/docs/v4/add-ons/fetch.mdx
```diff
@@ -0,0 +1,532 @@
...
+sidebarTitle: "Web Fetch"
+description: "Turn any webpage into agent context, without launching a browser"
+---
+
+## What is `fetch()`?
+
+Most of the time an agent doesn't need to *use* a page, it needs to *read* one. `browserbase.fetch()` takes a URL and returns its content as markdown, structured JSON, or the raw body — in one HTTP call, with no browser to start, no page to navigate, and nothing to close.
```
Kylejeong2: "Sometimes" instead of "most of the time" i think would be better

Kylejeong2: i think we should just cut the: "— in one HTTP call, with no browser to start, no page to navigate, and nothing to close."

packages/docs/v4/add-ons/fetch.mdx
```diff
@@ -0,0 +1,532 @@
...
+
+<Tabs>
+<Tab title="TypeScript">
+```typescript
+import { browserbase } from "@browserbasehq/stagehand";
+
+const fetchResult = await browserbase.fetch({
+  apiKey: process.env.BROWSERBASE_API_KEY,
```
Kylejeong2, line 17: do you ahve the pass in the api key every call? is there a way to just init a browserbase client for this @miguelg719

packages/docs/v4/add-ons/fetch.mdx
```diff
@@ -0,0 +1,532 @@
...
+export BROWSERBASE_API_KEY="bb_live_..."
+```
+
+<Tabs>
+<Tab title="TypeScript">
+```typescript
+import "dotenv/config";
+import { browserbase } from "@browserbasehq/stagehand";
+
+const { BROWSERBASE_API_KEY } = process.env;
```
Kylejeong2: does this process env syntax match the rest of the docs

packages/docs/v4/add-ons/fetch.mdx
```diff
@@ -0,0 +1,532 @@
...
+  "headers": { "content-type": "text/html; charset=utf-8" },
+  "statusCode": 200
+}
+```
+
+<Note>
+Field names follow each language's conventions: `contentType` and `statusCode` in TypeScript, `content_type` and `status_code` in Python, and `ContentType` and `StatusCode` on the Go structs. In Go, `Content` is a union — call `AsString()` for `raw` and `markdown`, `AsObject()` for `json`.
+</Note>
+
+## API reference
```
Kylejeong2, line 424: should this be the same file? or a diff/new page?

packages/docs/v4/add-ons/search.mdx
```diff
@@ -0,0 +1,372 @@
...
+
+for _, result := range searchResult.Results {
+	fmt.Printf("%s: %s\n", result.Title, result.URL)
+}
+```
+</Tab>
+</Tabs>
+
+<Note>
+`search()` hangs off the same `browserbase` facade you use to launch cloud browsers, but it never starts a browser session. All it needs is a Browserbase API key — there is no `Stagehand` instance, no page, and nothing to close. Local browsers have no equivalent.
```
Kylejeong2: this is a weird note, can we remove

packages/docs/v4/add-ons/search.mdx
```diff
@@ -0,0 +1,372 @@
...
+## What is `search()`?
+
+An agent that doesn't know where to start has to launch a browser, load a search engine, wait for the page, and read a result list before it can do anything useful. `browserbase.search()` collapses that into a single API call: you send a query, you get back ranked results with URLs, titles, and metadata.
+
+<Tabs>
+<Tab title="TypeScript">
+```typescript
+import { browserbase } from "@browserbasehq/stagehand";
+
+const searchResult = await browserbase.search({
```
Kylejeong2, line 16: same note as above for fetch

packages/docs/v4/add-ons/search.mdx
```diff
@@ -0,0 +1,372 @@
...
+
+<ParamField path="query" type="string" required>
+  The search query, 1–200 characters.
+</ParamField>
+
+<ParamField path="numResults" type="number" optional>
+  How many results to return, between 1 and 25. Defaults to 10.
+</ParamField>
+
+<ParamField path="baseUrl" type="string" optional>
```
Kylejeong2, line 213: why do we let you bring your own baseUrl for search? we should hide this anyways

packages/docs/v4/add-ons/search.mdx
charlypoly: would rephrase because we need to callout that this is a browserbase only feature, not a stagehand (local) one

packages/docs/v4/add-ons/fetch.mdx
charlypoly: I'd keep it in the same file for now as the current SDK reference is only covering Stagehand API (local), not browserbase

Kylejeong2 (approved):

miguelg719 (approved):

PR #2883, opened 2026-09-04
Expand `Page.on` to support custom event return classes

ziruihao:
# why

Today, `page.on` can support more than `'console'` (per last change), but cannot support advanced class types as return values.

Put another way, it can only support _serializable_ objects sent over RPC. If we want the DX of `page.on` to return `() => SomeNiceUsableClass.withNiceMethods()`, we need to add type overloading for each language (Go/Python/TS).

# what changed

# test plan

### DIFF
--- a/packages/protocol/schemas.ts
+++ b/packages/protocol/schemas.ts
@@ -1441,7 +1441,9 @@ export const ResponseFinishedResultSchema = z
   })
   .meta({ id: "ResponseFinishedResult" });
 
-export const PageEventNameSchema = z.enum(["console"]).meta({ id: "PageEventName" });
+export const PageEventNameSchema = z
+  .enum(["console", "requestfinished"])
+  .meta({ id: "PageEventName" });
 
 export const PageCDPEventParamsSchema = z
   .record(z.string(), z.json())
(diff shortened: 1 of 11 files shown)


packages/protocol/schemas.ts
```diff
@@ -1457,12 +1459,18 @@ export const PageCDPEventSchema = z
   })
   .meta({ id: "PageCDPEvent" });
 
-export const PageCDPEventNotificationSchema = z
+export const PageEventSchemas = {
+  console: PageCDPEventSchema,
+  requestfinished: NavigationResponseDescriptorSchema,
+} satisfies Record<z.infer<typeof PageEventNameSchema>, z.ZodType>;
```
ziruihao, line 1465: This would be the `page.on` return event type mapping for what is _serializable_.

packages/sdk-ts/src/page.ts
```diff
@@ -46,8 +47,13 @@ export type PageSetViewportSizeOptions = NonNullable<PageSetViewportSizeParams["
 export type PageTypeOptions = NonNullable<PageTypeParams["options"]>;
 export type PageWaitForSelectorOptions = NonNullable<PageWaitForSelectorParams["options"]>;
 
-export interface PageEventListener {
-  (event: PageCDPEvent): unknown;
+export type PageEventPayload<E extends PageEventName> = {
+  console: PageEvent<"console">;
+  requestfinished: Response;
+}[E];
```
ziruihao, line 53: This would be the per-language type mapping for _what isn't serializable, aka custom classes_.

PR #2946, opened 2026-09-17
feat(cli): Add command to List secrets

AzamAbdul:
# why
We want to add secrets support to the browse cli. In order to do so, we need to first introduce a common library that is a wrapper around the REST api, and then hook it into the cli tool. For simplicity, we choose the list secrets endpoint as the first CRUD operation that will be exposed to the cli tool 

# what changed
- Adds a library for wrapping around the secrets api
- implements the first command to retrieve a list of secrets
# test plan
- [x] unit tests
- [x] pointed this branch at a local api and confirmed listing secrets (metadata only) on test projects works


packages/cli/src/commands/cloud/secrets/list.ts
```diff
@@ -0,0 +1,20 @@
...
+import {
+  listSecretsFlags,
+  toListSecretsOptions,
+} from "../../../lib/secrets/flags.js";
+import { outputJson } from "../../../lib/output.js";
+
+export default class SecretsList extends BrowseCommand {
+  static override description =
+    "List project secret metadata with cursor pagination.";
+  static override examples = ["browse cloud secrets list"];
```
shrey150: Small usability nit: could we add `browse cloud secrets list --start-at 2026-01-01T00:00:00Z` to the examples? This gives users a copyable RFC 3339 timestamp and follows the existing pattern of flag examples in `cloud sessions list`.

shrey150 (approved):
Left small comment on adding an example, helpful for agents

PR #2967, opened 2026-09-17
feat(cli) Add command to CREATE an encrypted secret

AzamAbdul:
###  why
We want to add secrets support to the browse cli and is a continuation of the efforts introduced in this PR:
https://github.com/browserbase/stagehand/pull/2946.

In this particular PR, we add support for creating a secret by retrieving the public key for the project, reading the secret value from an env variable, a value piped to stdin, or prompting them in a password prompt (the inquire package), encrypting the value with the public key, then calling the create secret endpoint with the secret key name and the encrypted value. 

### what changed
- Adds a command to create a secret

### test plan
- [x] unit tests
- [x] point cli at local secrets api, verify encrypted secrets value lands in local db


packages/cli/src/commands/cloud/secrets/create.ts
```diff
@@ -0,0 +1,34 @@
...
+import { createSecret } from "../../../lib/secrets/api.js";
+import { readSecretValue } from "../../../lib/secrets/input.js";
+import { secretInputFlags } from "../../../lib/secrets/flags.js";
+import { outputJson } from "../../../lib/output.js";
+
+export default class SecretsCreate extends BrowseCommand {
+  static override description =
+    "Create a project secret. Encrypts the value locally with the project public key.";
+  static override examples = [
+    "browse cloud secrets create SERVICE_TOKEN",
```
shrey150, line 13: nit: is it best practice to indicate secrets prepended with a `$`, or is that only if it's literally being read from the local environment? may be helpful to follow best practice if agents are using

shrey150: ^ ignore

shrey150 (approved):
would respond to cubic comments before merging

packages/cli/src/lib/secrets/api.ts
```diff
@@ -56,3 +57,24 @@ export async function deleteSecret(
 function secretPath(id: string): string {
   return `/v1/secrets/${encodeURIComponent(id)}`;
 }
+
+export async function createSecret(
+  options: SecretsApiOptions,
+  secretKey: string,
+  value: Uint8Array,
+): Promise<Secret> {
+  const keypair = await requestBrowserbaseJson<{
```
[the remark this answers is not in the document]
AzamAbdul: I plan on updating the error details upstream in the secrets api and letting the error propagate through the cli. Basically, this is intended

PR #2990, opened 2026-09-19
feat(cli) Add command to Update a Secret

AzamAbdul:
###  why
We want to add secrets support to the browse cli and is a continuation of the efforts introduced in this PR:
https://github.com/browserbase/stagehand/pull/2946.

In this particular PR, we add support for updating a secret by retrieving the public key for the project, reading the secret value from an env variable, a value piped to stdin, or prompting them in a password prompt (the inquire package), encrypting the value with the public key, then calling the update secret endpoint via patch with the provided secret-id and the encrypted value. 

### what changed
- Adds a command to update a secret

### test plan
- [x] unit tests
- [x] point cli at local secrets api, verify encrypted secrets value lands in local db and is updated


miguelg719 (approved):

packages/cli/src/commands/cloud/secrets/update.ts
```diff
@@ -0,0 +1,34 @@
...
+  static override description =
+    "Replace a secret value, encrypting it locally with the current project public key.";
+  static override examples = [
+    "browse cloud secrets update <secretId>",
+    "browse cloud secrets update <secretId> --env MY_SERVICE_TOKEN",
+    "browse cloud secrets update <secretId> --stdin < ./secret.txt",
+  ];
+  static override args = {
+    secretId: Args.string({
+      description: "Project secret ID.",
```
shrey150: nit: maybe mention the format or give an example of a secretId if helpful for agents? but if not done for other CRUD operations, no need to do here

shrey150 (approved):
would address / resolve cubic comments before continuing

packages/cli/src/commands/cloud/secrets/update.ts
AzamAbdul: I can add it for the other crud ops too here

